PR #6, opened 2026-04-02
issue-5-a1-projection-schema-g

tomohisa:
## Execution Report

Piece `default` completed successfully.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+**/bin/
+**/obj/
+.takt/runs/
--- a/IntentSystem.sln
+++ b/IntentSystem.sln
@@ -1,0 +1,56 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.0.31903.59
+MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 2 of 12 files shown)


tomohisa:
Repair in place is enough here; no reset needed. Please keep this PR and fix the projection contract gaps below.

Required fixes:

1. `ImplementationIssuePacket` must carry the fields that now exist in the parent issue packet shape:
- `technical_baseline`
- `project_local_guide`
- `intent_baseline`

2. The projection input side must be able to represent more than a single success signal. Right now `SubSliceRow.SuccessSignal` and `ProjectionMapper` collapse acceptance to one string. The current source of truth for issue #5 includes multiple acceptance criteria, so the model and mapper need to support that shape.

3. The projection input side must be able to carry the embedded intent summary that the issue body now requires. We no longer rely on links alone; the issue packet must embed the core intent context needed for implementation. That means the projection model and mapper need a place to receive and emit `intent_baseline`.

4. Update tests so they verify the current parent contract, not the older reduced packet shape. In particular, add coverage that the mapped implementation packet contains the baseline fields above, and that multiple acceptance criteria can round-trip into the projected packet.

What is already good:
- .NET baseline is respected
- `node_modules/` is not present
- `.takt/runs/` is ignored

Please treat the parent issue body for #5 as the current contract source and bring the mapper/model layer up to that shape.

PR #14, opened 2026-04-02
issue-13-d1-clarify-inbox-goal

tomohisa:
## Execution Report

Piece `default` completed successfully.


tomohisa:
Repair in place is enough here; please keep this PR open and fix the clarification artifact path contract.

Required direction:

1. `ClarificationPathResolver` must align with the current parent clarification artifact layout under `.intent-cli/clarifications/<execution-unit>/...`.
2. Do not hardcode a single `.json` artifact as the canonical output path. The current parent baseline treats clarification artifacts as managed Markdown/YAML artifacts for the review flow.
3. If you need internal JSON for implementation convenience, keep that as an internal detail only after the external artifact contract matches the parent baseline. Do not change the outward artifact path contract in this PR.
4. Update tests so they verify the current parent path/layout contract rather than a JSON-only artifact assumption.

Keep scope limited to `D1 Clarify Inbox`:
- clarify artifact path/boundary
- deterministic per-execution-unit artifact layout
- no queue resume / answer application / workflow expansion

What is already good:
- .NET baseline is respected
- tests are passing
- no `node_modules/` or `.takt/runs/` are committed

Please treat the parent `D1` packet and `Interview And Clarification Artifact Contract` as the contract source and bring the path resolver/tests into alignment.

tomohisa:
One more concrete repair direction for the remaining model mismatch.

The current parent D1 contract expects the clarification artifact model to be able to carry at least these concepts:
- `clarification_source`
- `question_id`
- `question_text`
- `reason`
- `affected_intents`
- `affected_execution_units`
- `blocking_or_nonblocking`
- `clarification_return_path`

A good repair path is:

1. Expand `ClarificationItem` so it can represent the canonical clarification artifact shape from the parent contract.
2. Keep the existing `Id` / `Question` / `Context` only if they map cleanly to the canonical fields. If they are aliases, make that mapping explicit in naming or serialization.
3. Update serializer tests so they verify those required fields are present in the artifact contract, not just the reduced internal shape.
4. Keep scope limited to artifact boundary/modeling. Do not add answer-application or queue-resume behavior here; that remains for `D2`.

Suggested interpretation if you want a minimal but contract-complete model:
- `Id` -> `question_id`
- `Question` -> `question_text`
- `Context` should likely be split into `reason` plus the explicit affected/return-path fields instead of staying as one opaque string

Please use the parent `Interview And Clarification Artifact Contract` as the source of truth and make the model/serializer round-trip that shape directly.

tomohisa:
One more concrete repair note: the remaining mismatch is now about the outward clarification artifact field contract, not the path layout.

The current parent D1 contract still expects the clarification artifact to round-trip these canonical fields:
- `execution_unit`
- `status`

Your current model now has:
- `affected_execution_units`
- `state`

Those are useful, but they do not replace the required outward contract fields.

Recommended repair direction:

1. Add the canonical outward fields `execution_unit` and `status` back into the model/serializer contract.
2. Keep `affected_execution_units` only if you want extra context, but do not use it as a replacement for `execution_unit`.
3. Either rename `state` to the outward `status` field or make serialization map it explicitly to `status`.
4. Update serializer tests so they assert the serialized artifact contains `execution_unit` and `status` in addition to any extra fields you keep.

A safe interpretation is:
- `execution_unit` = the primary linked execution unit for this clarification artifact
- `affected_execution_units` = optional wider context if more than one unit is impacted
- `status` = the canonical artifact field name for `open / answered / applied / cancelled`

Please keep scope limited to the clarification artifact contract. Do not expand into answer application or queue resume; that still belongs to `D2`.

tomohisa:
Latest review note against the current PR head: the remaining blocker is now only the outward clarification artifact contract guarantee.

The current parent D1 contract still requires `affected_execution_units` as part of the clarification artifact shape. Your model has that property, but the serializer and tests no longer enforce or prove it as a required outward contract field.

Recommended repair direction:

1. Add `affected_execution_units` back to the serializer's required contract field validation.
2. Update serializer tests so the serialized outward artifact explicitly contains `affected_execution_units`.
3. Add one negative test proving deserialization fails when `affected_execution_units` is missing.
4. Keep scope limited to the clarification artifact contract only. Do not expand into answer application or queue resume.

At this point, path layout and the primary canonical fields are already aligned. The remaining work is specifically to make the serializer/test layer guarantee the full outward clarification artifact contract from the parent spec.

PR #95, opened 2026-04-06
[codex] Implement intake apply command

tomohisa:
## Summary

- add `intent-cli intake apply <domain>` to read the current patch draft and deterministically update only the named parent source-of-truth files
- parse and validate the current `.intent-cli/intake/<domain>.patch.md` contract before applying idempotent managed update blocks per target file
- return a human-facing summary with changed file paths, applied edit count, and source concept refs, with command/parser/renderer/router coverage

## Verification

- `dotnet test IntentSystem.sln`

### DIFF
--- a/src/IntentSystem.Cli/Commands/CommandRouter.cs
+++ b/src/IntentSystem.Cli/Commands/CommandRouter.cs
@@ -78,6 +78,7 @@ internal static class CommandRouter
                 ["compile"] = IntakeCompileCommand.Execute,
                 ["foldin"] = IntakeFoldinCommand.Execute,
                 ["patch"] = IntakePatchCommand.Execute,
+                ["apply"] = IntakeApplyCommand.Execute,
                 ["autostart"] = IntakeAutostartCommand.Execute
             }
         };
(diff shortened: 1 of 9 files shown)


tomohisa:
Remaining blocker is the `G34` mutation contract.

The current parent source of truth fixes `intake apply <domain>` as the actual source-of-truth update boundary from `.intent-cli/intake/<domain>.patch.md` into the current parent files. This PR currently succeeds by appending/replacing an `intake-apply` managed block that contains metadata and proposed edits, but that managed-block format is not part of the current parent intent/spec baseline. In other words, the command can report success without actually applying the patch draft to the intended file content.

To repair in place:
- do not treat the managed-block format as the successful `G34` outcome unless that format is first added to the parent source of truth
- make `IntakeApplyCommand` apply the patch draft to the intended file content itself
- keep changed files limited to `target_file_paths`
- keep the command scope inside deterministic parent file update plus human-facing summary
- update `IntakeApplyCommandTests` so they prove real source-content updates instead of asserting the synthetic marker block

Scope stays `G34`; do not expand into execution sub-slice creation, issue packet generation, queue mutation, autostart, review, merge, or closeout.

PR #101, opened 2026-04-06
[codex] Issue 100 intake issue command

tomohisa:
## Summary
- add `intent-cli intake issue <domain>` as a thin issue artifact generation boundary
- resolve selected intake-origin issue-ready execution units from current `execution/` source-of-truth and generate issue artifacts deterministically
- keep the command read/write limited to issue artifact generation only, without queue mutation or downstream execution paths

## Testing

### DIFF
--- a/src/IntentSystem.Cli/Commands/CommandRouter.cs
+++ b/src/IntentSystem.Cli/Commands/CommandRouter.cs
@@ -80,6 +80,7 @@ internal static class CommandRouter
                 ["patch"] = IntakePatchCommand.Execute,
                 ["apply"] = IntakeApplyCommand.Execute,
                 ["execution"] = IntakeExecutionCommand.Execute,
+                ["issue"] = IntakeIssueCommand.Execute,
                 ["autostart"] = IntakeAutostartCommand.Execute
             }
         };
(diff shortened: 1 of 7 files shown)


tomohisa:
The remaining issue is the runtime input contract for `intake issue <domain>`.

Right now the implementation assumes the current `execution/` source-of-truth already contains a structured outward shape like:

- `execution_unit:`
- `source_file_path:`
- `target_part:`
- `dependencies:`
- `readiness_notes:`
- `verification_hints:`

under a section containing `` `intake execution apply <domain>` ``.

That is the problem: the current parent `G37` source of truth does **not** define that structured `execution/` representation yet, and the real parent repo does not currently contain those markers / bullet fields. So the command is only proven against a synthetic fixture contract, not against the current canonical parent input.

What needs to be true for this PR to be acceptable:

- derive intake-origin issue-ready units from a **parent-defined** execution boundary
- do **not** require invented `execution_unit:` / `source_file_path:` bullets unless that shape is first established in the parent source of truth
- update the tests so they prove the **canonical parent input shape**, not a child-invented fixture shape
- keep the scope at `G37`
- do not expand into queue mutation, child issue creation, autostart, review, or closeout

So the repair direction is:

1. either change the implementation to read from an already-defined parent execution/apply boundary
2. or, if this structured execution-unit representation is really required, first make that representation explicit in the parent source of truth and then align the command/tests to that canonical contract

Everything else in the PR looks appropriately thin. The blocker is specifically the hidden runtime discovery contract.

PR #103, opened 2026-04-06
[codex] Issue 102 queue enqueue command

tomohisa:
## Summary
- add `intent-cli queue enqueue <execution-unit>` as a thin queue insertion command from an existing packet artifact
- resolve queue item metadata from `.intent-cli/issues/<execution-unit>/packet.yaml` and append a deterministic `queued` run event
- keep duplicate handling and queue mutation isolated from child issue creation, autostart, and workflow or review execution

## Testing

### DIFF
--- a/src/IntentSystem.Cli/Commands/CommandRouter.cs
+++ b/src/IntentSystem.Cli/Commands/CommandRouter.cs
@@ -69,6 +69,7 @@ internal static class CommandRouter
                 ["list"] = QueueListCommand.Execute,
                 ["show"] = QueueShowCommand.Execute,
                 ["next"] = QueueNextCommand.Execute,
+                ["enqueue"] = QueueEnqueueCommand.Execute,
                 ["dispatch"] = QueueDispatchCommand.Execute,
                 ["transition"] = QueueTransitionCommand.Execute
             },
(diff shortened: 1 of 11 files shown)


tomohisa:
The remaining issue is the `packet.yaml` section-name contract that `queue enqueue <execution-unit>` reads.

Right now the implementation routes the packet through `ProjectionPacketSerializer.Deserialize(...)`, and the tests generate fixtures with:

- `implementation_issue_packet`
- `review_context_packet`

That is the problem: the current parent `G38` source of truth does **not** use those section names. The canonical parent packet now uses:

- `implementation_issue`
- `review`

So this PR is still only proven against the older child-side packet shape, not against the real parent packet that `queue enqueue` is supposed to consume.

What needs to be true for this PR to be acceptable:

- runtime packet loading must accept the canonical parent `G38` packet shape
- tests must prove the same canonical parent packet shape
- do not keep locking in `implementation_issue_packet` / `review_context_packet` as the success path
- keep the scope at `G38`
- do not expand into child issue creation, autostart, review, or closeout

So the repair direction is:

1. align `QueueEnqueueCommand` packet loading with the current parent packet contract
2. update `QueueEnqueueCommandTests` to use the canonical parent-style packet fixture
3. verify the command still stays thin: queue insertion plus `queued` append only

Everything else in the PR looks appropriately scoped. The blocker is specifically the packet contract mismatch against the current parent source of truth.

PR #119, opened 2026-04-07
[codex] Add generate-from-current bridge

tomohisa:
## Summary
- add `generate-from-current bridge <domain>` as a thin bridge from reconstructed artifacts into the standard intake/interview flow
- regenerate `.intent-cli/intake/<domain>.concept.yaml` and `.intent-cli/interviews/<domain>/` deterministically for the selected domain only
- add parser, renderer, and router/runtime coverage for the bridge path

## Testing

### DIFF
--- a/src/IntentSystem.Cli/Commands/GenerateFromCurrentBridgeCommand.cs
+++ b/src/IntentSystem.Cli/Commands/GenerateFromCurrentBridgeCommand.cs
@@ -137,6 +137,7 @@ internal static class GenerateFromCurrentBridgeCommand
                 .OrderBy(value => value, StringComparer.Ordinal)
                 .ToArray(),
             KnownUnknowns = reconstructedInterview.RecommendedFollowUpQuestions
+                .Concat(reconstructedInterview.BridgeQuestions.Select(question => question.QuestionText))
                 .Concat(reconstructedInterview.Gaps)
                 .Distinct(StringComparer.Ordinal)
                 .OrderBy(value => value, StringComparer.Ordinal)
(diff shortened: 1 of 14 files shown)


tomohisa:
One remaining blocker here is the bridge semantic contract.

Right now `GenerateFromCurrentBridgeCommand` derives `reason` and `blocking_or_nonblocking` by matching hard-coded English prefixes in the reconstructed follow-up question text. That means the runtime contract is effectively tied to the exact `G45` wording, even though `G46` is supposed to bridge reconstructed artifacts into the standard intake baseline, not infer semantics from literal phrasing.

What is wrong:
- bridge behavior changes if the reconstructed question wording changes, even when the meaning stays the same
- the parent source of truth does not currently define `question text prefix -> reason/blocking` as the canonical contract
- the tests lock in the same hidden contract by asserting behavior from those exact strings

Resolution conditions:
- stop deriving bridge semantics from `StartsWith(...)` checks on question text
- carry explicit bridgeable metadata in the reconstructed artifact, or establish a parent-defined derivation contract that is independent of literal wording
- update tests so they prove that canonical artifact contract, not exact English phrasing
- keep scope inside `G46`; no need to broaden into source collection, reconstruction, queue/launch, or later workflow steps

Once the bridge semantics are driven by explicit artifact data instead of question wording, this should be in good shape.

tomohisa:
One remaining blocker is the reconstructed interview consistency contract.

The previous wording-coupled bridge issue is fixed, but the runtime still accepts a reconstructed interview artifact where `recommended_follow_up_questions` and `bridge_questions` disagree. Right now the bridge uses those two sections differently:
- `recommended_follow_up_questions` feeds the regenerated `.intent-cli/intake/<domain>.concept.yaml`
- `bridge_questions` feeds the regenerated `.intent-cli/interviews/<domain>/` tree

So a malformed `.reconstructed-interview.md` can silently produce two different question sets in the standard intake baseline while still reporting success.

What is wrong:
- the deserializer accepts any combination of those two sections without validating that they represent the same downstream question set
- the current tests only prove aligned fixtures, so this split contract is still unguarded
- `G46` is supposed to bridge reconstructed artifacts deterministically into the standard intake baseline, not allow contradictory downstream outputs from one input artifact

Resolution conditions:
- either validate a one-to-one match between `recommended_follow_up_questions` and `bridge_questions`
- or make one canonical question list drive both the concept packet and interview artifact generation
- add tests that prove malformed / conflicting reconstructed interview input is rejected, or that one canonical source is used consistently
- keep scope inside `G46`; no need to broaden into source collection, reconstruction inference, queue/launch, or later workflow steps

Once the bridge cannot emit contradictory standard intake artifacts from one reconstructed interview input, this should be in good shape.

PR #199, opened 2026-04-09
Add auto-id and text input to bug report

### LINKED ISSUE #196: G84 Bug Report Auto-ID And Text Input
## Summary

Implement `intent-cli bug report <domain> [<bug-id>] --title <text> [--text <text> | --from-file <path>]` as a working command.

## Why

- bug report should default to auto-generating `bug-id` when omitted
- bug report should accept a single long-form inline text input without requiring a prepared file

## In Scope

- optional bug-id support
- deterministic auto-id generation
- inline `--text <text>` input
- existing `--from-file <path>` fallback retention
- report artifact / router / command tests

## Out Of Scope

- bug triage
- bug plan / repair creation
- queue mutation
- stdin / `$EDITOR` fallback
- workflow execution

## Acceptance

- omitted bug id path deterministically resolves a generated bug id and report artifact path
- `--text <text>` path produces the same bounded report artifact shape as `--from-file <path>`
- explicit bug id path remains supported
- no scope drift into triage, repair launch, or workflow execution

tomohisa:
## Summary
- allow `bug report` to omit `bug-id` and deterministically generate one from the input
- add inline `--text <text>` support alongside existing `--from-file <path>` input
- keep the existing explicit bug-id and file-based path intact while tightening router and command coverage

Closes #196

### DIFF
--- a/src/IntentSystem.Cli/Commands/BugReportCommand.cs
+++ b/src/IntentSystem.Cli/Commands/BugReportCommand.cs
@@ -1,3 +1,6 @@
+using System.Security.Cryptography;
+using System.Text;
+
 namespace IntentSystem.Cli.Commands;
 
 internal static class BugReportCommand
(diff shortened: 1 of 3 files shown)


tomohisa:
Remaining blocker is the `G84` auto-id contract.

Parent source-of-truth fixes omitted `bug-id` to a deterministic repo-local auto-id from the current date plus normalized title. The current child implementation instead hashes `domain + title + problemStatement`, which introduces hidden pre-triage dedupe semantics: identical later reports collapse into the same lineage, while minor wording changes fork a new one.

Please keep the fix inside `G84` and change the default id generation back to the parent-defined date/title baseline. The main things to prove in tests are:
- omitted `bug-id` uses the current date plus normalized title baseline
- same title on the same date stays deterministic
- repeated observations are not implicitly folded by content hash
- `--text <text>` and `--from-file <path>` still converge to the same bounded report artifact shape apart from `report_source`

No need to widen scope into triage, dedupe policy changes, stdin/editor fallback, or any downstream repair flow.

Issue #237, opened 2026-04-10
Implementation repair: Review normalized result does not match launched review request (BUG-20260410-review-result-request-mismatch)


tomohisa:
# Implementation repair: Review normalized result does not match launched review request (BUG-20260410-review-result-request-mismatch)

Repair child implementation targets for 'Review normalized result does not match launched review request' (BUG-20260410-review-result-request-mismatch) using .intent-cli/bugs/BUG-20260410-review-result-request-mismatch.plan.yaml: .intent-cli/issues/G96/packet.yaml

## Implementation Repair Targets
- .intent-cli/issues/G96/packet.yaml

tomohisa:
Accepted via PR #238.

PR #238, opened 2026-04-10
Repair review result/request mismatch

tomohisa:
## Summary
- include `upstream_request_ref` in normalized direct-run result artifacts
- persist the same launched upstream request ref when synthesizing implement, fix, and review results
- add regression coverage so review normalized results stay aligned with the launched review request

## Verification
- `dotnet test tests/IntentSystem.Cli.Tests/IntentSystem.Cli.Tests.csproj --filter "FullyQualifiedName~ReviewRunCommandTests|FullyQualifiedName~RunImplementCommandTests|FullyQualifiedName~RunFixCommandTests|FullyQualifiedName~DirectRunResultArtifactJsonTests|FullyQualifiedName~RunCommandTests"`
- `dotnet test IntentSystem.sln`

Refs #237

### DIFF
--- a/src/IntentSystem.Cli/Commands/DirectRunCommandSupport.cs
+++ b/src/IntentSystem.Cli/Commands/DirectRunCommandSupport.cs
@@ -92,6 +92,7 @@ internal static class DirectRunCommandSupport
             context,
             executionUnit,
             entryKindValue,
+            upstreamRequestRef,
             launchedAt,
             launchResult);
 
@@ -224,6 +225,7 @@ internal static class DirectRunCommandSupport
         CliContext context,
         string executionUnit,
         string entryKind,
+        string upstreamRequestRef,
         DateTimeOffset launchedAt,
         DirectRunLaunchResult launchResult)
     {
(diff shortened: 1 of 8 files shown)


tomohisa:
One blocker remains. This patch carries `upstream_request_ref` into `.intent-cli/runs/<execution-unit>.result.json`, but nothing in the runtime uses that field to reject or repair a stale review result. `RunCommand.TryReadDirectRunResultArtifact` still accepts any result whose `entry_kind` matches (`review`), and the new tests only assert that the ref is serialized. The dogfooding failure behind #237 was stronger: the review request and provider raw log were for the current launched review session, while the normalized result still held stale provider/session context. Please complete the fix by making normalized review result synthesis and/or root-run consumption validate against the current launched request boundary, so a stale review result cannot satisfy the current review wait path merely by sharing `entry_kind=review`.

tomohisa:
One blocker remains on the updated head. The new request-boundary check is directionally correct, but `DirectRunResultArtifact` now makes `upstream_request_ref` required during deserialization. In real dogfooding state, existing `.intent-cli/runs/<execution-unit>.result.json` files were written before this field existed, so `intent-cli run` now aborts with `JsonException` before it can even classify the result as stale or waiting. I reproduced this on toy-calc-sample with the current branch. Please make result-artifact reading backward-compatible for pre-field artifacts, or add a deterministic migration/fallback path, so old runtime state does not crash the root loop.

Issue #346, opened 2026-04-19
[G128] Auto-Continue Root Run Into Existing Intake Launch After Closeout


tomohisa:
# [G128] Auto-Continue Root Run Into Existing Intake Launch After Closeout

## Goal

Repair the next orchestration gap after `G127`: when the current child slice is already accepted/closed out and the current repo intent state already contains a clarification-free next intake-origin issue-ready slice, root `intent-cli run` must continue through existing `intake issue` / `queue dispatch` / `intake launch` surface instead of stopping and requiring manual GitHub or queue operations.

## Why This Slice Exists Now

- the canonical workflow is `review -> merge -> closeout -> next issue -> implement` without a manual GitHub step in between
- current accepted surface already has command-level issue creation and launch plumbing:
  - `intake issue`
  - `queue dispatch`
  - `intake launch`
- however, current dogfooding still required manual packet/queue/GitHub intervention before `run` could continue to the next slice
- therefore the remaining gap is orchestration over existing accepted command surface, not missing GitHub mutation plumbing itself

## Current Accepted Baseline

- root `run` is the orchestration entrypoint
- child issue creation should happen through `intent-cli` command surface, not manual `gh issue create`
- existing intake-origin issue/launch commands are already accepted and should be reused
- planner-specific multi-provider expansion is a later continuation, not this slice

## Target

- repo: `submodules/intent-system`
- path: `.`
- part: `root run post-closeout bridge into existing intake issue/launch surface`

## In Scope

- root-run post-closeout continuation when a next intake-origin issue-ready slice already exists and no clarification is required
- detecting that the current execution unit is complete and the next issue-ready intake slice is actionable
- invoking existing `intake issue`, `queue dispatch`, `intake launch`, or equivalent already-accepted command boundaries instead of stopping
- leaving canonical runtime artifacts and queue/runs history for the newly launched next slice
- focused tests for `closeout -> next issue -> launch` continuation
- live dogfooding proof on the toy-calc shape

## Out Of Scope

- introducing a brand-new GitHub mutation path
- replacing `queue dispatch` / `intake launch`
- full AI planner/provider abstraction for generating new issue-ready slices from ambiguous intent state
- redesigning all `run` stop semantics
- review/fix runtime bugs unrelated to the post-closeout bridge

## Acceptance Criteria

[...]

PR #347, opened 2026-04-19
[G128] Auto-continue root run into intake launch

tomohisa:
## Summary
- teach root `run` to auto-continue into intake when the current queue is fully closed out and an intake execution draft still has launchable units
- reuse the existing `intake issue` and `intake launch` command surfaces instead of introducing a new dispatch path
- add root-run regression coverage for launchable and already-completed intake slices

## Verification
- `dotnet test tests/IntentSystem.Cli.Tests/IntentSystem.Cli.Tests.csproj --filter RunCommandTests --nologo`
- `dotnet test IntentSystem.sln --nologo`

Relates to #346

### DIFF
--- a/src/IntentSystem.Cli/Commands/IntakeIssueCommand.cs
+++ b/src/IntentSystem.Cli/Commands/IntakeIssueCommand.cs
@@ -71,12 +71,26 @@ internal static class IntakeIssueCommand
 
     internal static IntakeIssueResult ExecuteCore(string repoRoot, string domain)
     {
+        return ExecuteCore(repoRoot, domain, executionUnit: null);
+    }
+
+    internal static IntakeIssueResult ExecuteCore(string repoRoot, string domain, string? executionUnit)
+    {
         var baseline = LoadIssueBaseline(repoRoot);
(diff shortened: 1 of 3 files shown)


tomohisa:
P1: `G128` broadens post-closeout auto-continuation from "the next issue -> implement" into "launch every launchable execution unit in the first intake domain". `RunCommand` only resolves a domain, then unconditionally calls `intake issue` and `intake launch` on that whole domain ([RunCommand.cs:502-585](/private/tmp/intent-pr347-review/src/IntentSystem.Cli/Commands/RunCommand.cs:502)), and the new regression test explicitly locks in launching both `AUTH-01` and `AUTH-02` from one continuation ([RunCommandTests.cs:225-289](/private/tmp/intent-pr347-review/tests/IntentSystem.Cli.Tests/RunCommandTests.cs:225)). The accepted `G128` contract is narrower: after closeout, root should auto-continue into the next actionable clarification-free slice, singular, not bulk-create/bulk-launch every issue-ready unit in a domain. Please narrow the selection to exactly one next execution unit and reuse per-unit accepted boundaries for that one unit only; do not let one closeout step fan out across multiple sibling execution units.

Issue #398, opened 2026-04-22
G146 Document Request-Update Repair Loop Handoff


tomohisa:
# [G146] Document Request-Update Repair Loop Handoff

## Goal

Add one repo-local operator note in the child repository that explains the
request-update repair loop from actionable review comment through child-side fix
push and back into host-side rereview.

## Why This Slice Exists Now

`G144` and `G145` already proved the parent-backed green path. The remaining
automation path to verify is the repair loop after an actionable review comment.
We need a parent-backed, docs-only slice that is safe to repair in place when a
seeded request-update comment is applied to its PR.

## Current Observed State or Reproduction

- `intent-pr-request-update` marks a PR that needs repair work
- the child-side request-update runner should pick that PR and perform the fix
- after a successful fix push, the PR should transition to
  `intent-pr-rereview-ready`
- host-side review-closeout should then pick up the PR again for rereview
- there is not yet one repo-local child-side note that explains the full
  request-update repair loop in one place

## Accepted Baseline You May Assume

- child-facing GitHub issues must be standalone contracts for issue-only workers
- `G144` verified the target-scoped green path
- `G145` documented the meaning of `intent-pr-rereview-ready`
- current accepted automation split is:
  - issue runner owns issue -> PR
  - request-update runner owns repair after actionable review feedback
  - review-closeout owns host-side review, rereview pickup, and accepted
    closeout
- this slice remains documentation-only

## Target Repo / Path / Part

- Repo: `J-Tech-Japan/intent-system`
- Path: `ops/intent-request-update-repair-loop.md`
- Part: `operator-facing documentation for request-update repair loop handoff`

## In Scope

- add one markdown document at `ops/intent-request-update-repair-loop.md`
- explain the repair loop in order:
  - actionable review comment
  - `intent-pr-request-update`
  - child-side fix work
  - `intent-pr-rereview-ready`
  - host-side rereview pickup
- distinguish child-side repair ownership from host-side rereview ownership
- keep the slice documentation-only and safe for repair-in-place

## Out Of Scope

- code changes
- automation prompt rewrites
- label renaming
- workflow redesign
- scheduling or ownership changes for automations

## Acceptance Criteria

- `ops/intent-request-update-repair-loop.md` exists in the child repo
- the document explains the ordered request-update repair loop
- the document explicitly distinguishes child-side repair from host-side
  rereview
- the document remains documentation-only

## Verification

- confirm `ops/intent-request-update-repair-loop.md` exists
[...]

PR #399, opened 2026-04-22
G146 Document Request-Update Repair Loop Handoff

tomohisa:
## Summary
- add an operator-facing note for the request-update repair loop handoff
- explain the ordered flow from actionable review comment through rereview pickup
- distinguish child-side repair ownership from host-side rereview ownership

## Validation
- git diff --check
- verified required repair-loop and ownership wording in ops/intent-request-update-repair-loop.md

Refs #398

### DIFF
--- a/ops/intent-request-update-repair-loop.md
+++ b/ops/intent-request-update-repair-loop.md
@@ -1,0 +1,53 @@
+# Request-Update Repair Loop Handoff
+
+This note explains the request-update repair loop for target-scoped PR automation.
+
+## Ordered Repair Loop
+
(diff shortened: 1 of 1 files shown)


tomohisa (commented):
This repair-loop smoke note explains the overall flow, but it still does not state one operator-critical transition: after a successful child-side repair push, the request-update runner must remove `intent-pr-request-update`, add `intent-pr-rereview-ready`, and only then hand the PR back to host-side review-closeout for rereview. Please narrow the fix to that missing transition contract and keep the slice docs-only.

PR #1165, opened 2026-07-19
G530: Prioritize semantic-facet nodes in AI context supply

### LINKED ISSUE #1164: G530: Prioritize semantic-facet nodes in AI context supply (context collect and packet review context)
# G530: Prioritize semantic-facet nodes in AI context supply (context collect and packet review context)

## Goal

Make the four semantic facets (G529: `vocabulary`, `invariant`, `decider`, `acceptance-property`) the preferred content of AI-facing context supply: `context collect` returns facet-classified nodes in a distinct prioritized section (with a `--facets` filter), and packet scaffolding lists the facet nodes overlapping a slice in its generated `review-context.md` — so implement/review agents receive the semantic core a change must respect, localized instead of buried.

## Why This Slice Exists Now

Operator input (issue #1159): supplying the four human-retained facets preferentially is the second pillar of the initiative — it localizes the semantic context a change needs, which is exactly what ES/CQRS-style additive change protocols require. A field lesson reinforces this: the sekiban-as-a-service G812 delegation (2026-07-14) had to be stopped and decomposed because the implement agent could not hold the full semantic surface of a ~2000-line hand-rolled engine; facet-scoped context is the structural mitigation direction the intent side can contribute. G529 provides the classification; without this slice it stays invisible to agents.

## Current Observed State

- `context collect` (the read-only tasking-thread context surface) returns context without any facet notion — after G529 lands, facet-annotated nodes are classifiable but not prioritized or filterable.
- `packet draft` scaffolds `review-context.md` with intent references but nothing surfaces which vocabulary/invariant/decider/acceptance-property nodes a slice must respect.
- Reviewers reconstruct that surface by hand (see G812 inventory comment) or miss it.

- G529 is merged: `facets` frontmatter, lint validation, search filter, analyze-tree counts exist.
- All surfaces here are read-only; output must stay structurally stable for agent consumption.
- Trees without facet annotations are the norm initially — graceful degradation (current behavior + explicit no-facet-context note) is required, never an error.
- The packet four-file layout and the standalone issue contract are unchanged.

## Target Repo / Path / Part

Target paths: `src/IntentSystem.Cli/**, tests/**, docs/ja/**, docs/en/**`

Target part: `context collect` facet prioritization + filter; `packet draft` review-context Facet context section; developer-reference docs.

## In Scope

- `context collect`: a distinct facet section ordered `vocabulary → invariant → decider → acceptance-property`, placed ahead of unclassified context; `--facets <list>` filter to return only selected facets; documented JSON shape.
- Relevance scoping: facet nodes are selected for the requested domain and, when the caller passes scope hints (paths / intent references), narrowed by overlap with those hints; without hints, all facet nodes of the domain in facet order.
- `packet draft`: generated `review-context.md` gains a "Facet context" section listing facet nodes overlapping the packet `intent_references` (id, facet values, summary, path); regenerating an existing packet preserves hand-edits per existing packet-draft semantics.
- Graceful no-facets degradation with an explicit note.
- Focused tests: ordering, filter, overlap selection, degradation.
- ja/en developer-reference documentation.

## Out Of Scope

- Change-proposal contradiction checking (G531).
- Facet schema changes (G529 owns the schema; the value set stays closed).
- Annotating domain trees; orchestrator guide changes referencing facet context (follow-up after G531).
- Any ranking heuristics beyond facet order + scope overlap (no semantic similarity engines).

The child PR makes `context collect` facet-aware — a prioritized facet section ordered vocabulary/invariant/decider/acceptance-property ahead of unclassified context, a `--facets` filter, scope-hint narrowing, documented stable JSON shape, and an explicit graceful note when a domain has no facet-annotated nodes — and extends `packet draft` so generated `review-context.md` contains a Facet context section listing the facet nodes overlapping the packet intent references (id, facets, summary, path), preserving existing regeneration semantics. Read-only output surfaces only; focused tests cover ordering, filtering, overlap selection, and degradation; ja/en developer reference documents both. No cross-checking, no schema changes, no tree annotation.

## Acceptance Criteria

- `context collect --domain <d>` output contains the prioritized facet section in the specified order ahead of unclassified context; `--facets invariant,decider` returns only those facets.
- With scope hints, only overlapping facet nodes appear; without hints, all domain facet nodes appear in facet order.
- `packet draft` regeneration adds the Facet context section listing overlapping facet nodes with id, facet values, summary, and path.
- A domain with zero facet-annotated nodes yields current-behavior output plus an explicit no-facet-context note on both surfaces.
- JSON shape documented; focused tests green for all of the above.

## Verification

- Run the focused context/packet test suites and show them green.
- Manual check on a fixture tree with one node per facet: collect ordering, filter, packet regeneration.
- `git diff --check` clean.

## Related Links

- Operator input: https://github.com/J-Tech-Japan/intent-system/issues/1159
- Concept node: intents/intent-cli/intent-tree/means/10-semantic-facets.md (ICL.M.SEMANTIC_FACETS, host repo)
- G529 — facet schema (dependency); G531 — change cross-check (sibling).
- Field lesson: SekibanAsAService issue #1744 inventory comment (G812 oversized delegation).

## Knowledge Maintenance

- Intent placement: `intents/intent-cli/intent-tree/means/10-semantic-facets.md`.
- ADR candidate: none — additive read-only behavior.
- Diagram candidate: none.
- Closeout writeback expected: no — orchestrator-guide integration is follow-up after G531.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
## Summary

- Adds a shared `FacetContextSelector` — scans a domain's intent tree, classifies nodes by their G529 `facets:` frontmatter, groups them in the canonical `vocabulary → invariant → decider → acceptance-property` order, and narrows by path/intent-reference scope overlap — used identically by both surfaces below so they can never disagree on what "overlaps".
- `context collect --domain <d>` gains a `## Facet context` section rendered **ahead of** the unclassified queue-state/clarification/automation-bindings/recent-events context, a `--facets <list>` filter (validated against the closed G529 set, matching `intent search --facet`'s existing validation pattern), and a `--scope <paths>` narrowing flag (exact path, ancestor-directory prefix, or the shorter domain-relative id form all count as overlap). A domain with zero facet-annotated nodes at all gets an explicit graceful-degradation note — never an error.
- `packet draft` gains a generated `## Facet context` section inside the scaffolded `review-context.md`, listing the facet nodes overlapping the packet's own `implementation_issue_packet.intent_references` — read from an **existing** `packet.yaml` on disk (never the freshly-templated empty list this same invocation might otherwise write), so a packet whose `packet.yaml` was already hand-edited with real references produces a genuinely useful section the first time `review-context.md` is generated, while the "never overwrite an existing file" scaffold semantics stay exactly as before.

Closes #1164

## Test plan

- [x] `dotnet build src/IntentSystem.Cli/IntentSystem.Cli.csproj -c Debug` — succeeds
- [x] New `FacetContextSelectorTests` (13 tests: canonical ordering, facet filter, scope-overlap in all three hint forms, multi-facet nodes appearing once per matching group, malformed/unknown-value exclusion, missing-domain and no-facet-nodes graceful degradation, summary/path/id derivation)
- [x] New `ContextCollectCommandTests` facet fixtures (6 tests: section ordering ahead of Queue state, `--facets` filter, `--facets` validation error, `--scope` narrowing, graceful degradation note, stable snake_case JSON shape)
- [x] New `PacketDraftCommandTests` facet fixtures (3 tests: fresh scaffold degradation note, regeneration from an existing hand-edited `packet.yaml`'s `intent_references`, existing `review-context.md` never overwritten regardless of facet nodes)
- [x] Full solution test suite — 3386 passed / 1 pre-existing unrelated skip / 0 failed
- [x] `git diff --check` — clean
- [x] Manual end-to-end check on a fixture tree with one node per facet: `context collect` ordering/filter/scope all verified against the real CLI binary; `packet draft` regeneration verified to correctly include only the two intent_references-overlapping nodes and exclude the other two
- [x] ja/en `docs/09-developer-reference.md` (new section) and `docs/04-packets-issues.md` (cross-reference) updated

### DIFF
--- a/docs/en/04-packets-issues.md
+++ b/docs/en/04-packets-issues.md
@@ -6,7 +6,7 @@ This is **host/design** work. When your intent is clear enough to act on, the de
 
 ## What a packet is
 
-A **packet** is a focused, reviewable slice of intent that becomes an executable task. The design thread scaffolds a canonical set of files (`packet.yaml`, `implementation.md`, `review-context.md`, `github-body.md`) that describe exactly what needs to be built.
+A **packet** is a focused, reviewable slice of intent that becomes an executable task. The design thread scaffolds a canonical set of files (`packet.yaml`, `implementation.md`, `review-context.md`, `github-body.md`) that describe exactly what needs to be built. `review-context.md` includes a generated **Facet context** section listing the G529 semantic-facet nodes (vocabulary/invariant/decider/acceptance-property) overlapping the packet's `intent_references` — see [Facet-aware context supply (G530)](09-developer-reference.md#facet-aware-context-supply-g530).
 
 **Issue publish** turns a reviewed packet into a GitHub Issue with a **Standalone Child Issue Contract** — the only source of truth a child implementation agent needs to do the work. The child agent reads the issue body and the repository code; it does not access host metadata.
 
--- a/docs/en/09-developer-reference.md
+++ b/docs/en/09-developer-reference.md
@@ -484,6 +484,14 @@ intent-cli context collect --domain <d> --scope intents/<d>/means,identity/missi
   entirely returns every domain facet node; passing `--scope` with hints
   that all turn out invalid or outside the domain matches NOTHING — it
   never silently falls back to "no scoping requested".
+- A rejected `--scope` hint is never silent either: it produces a
+  `facet_context_scope_warnings` entry (`hint`, `reason`) naming the exact
+  hint and why it was rejected (outside the domain root, a `..` traversal,
+  etc.) — so "matched nothing because every hint was invalid" is never
+  indistinguishable from "matched nothing because a genuinely valid hint
+  just didn't overlap any node". A mixed list still applies the valid
+  hints while reporting every rejected one; `facet_context_all_scope_hints_rejected`
+  is `true` only when every requested hint was rejected.
 - A domain with ZERO facet-annotated nodes at all (not merely a
   `--scope`/`--facets` query that matched nothing) sets `facet_context_note`
   and renders an explicit note instead of an empty section — graceful
@@ -498,16 +506,19 @@ intent-cli context collect --domain <d> --scope intents/<d>/means,identity/missi
 - JSON shape: `facet_context: [{facet, nodes: [{id, facets, summary,
   path}]}]` (always 4 entries, or fewer when `--facets` is passed),
   `facet_context_note: string | null`, `facet_context_warnings: [{path,
-  reason}]`.
+  reason}]`, `facet_context_scope_warnings: [{hint, reason}]`,
+  `facet_context_all_scope_hints_rejected: bool`.
 
 **`intent-cli packet draft`** generates a `## Facet context` section inside
 the scaffolded `review-context.md`, listing the facet nodes overlapping the
 packet's own `implementation_issue_packet.intent_references` — the exact
-same overlap logic `context collect`'s `--scope` uses, so the two surfaces
-can never disagree about what "overlaps". The generated content lives
-between two HTML-comment markers
-(`<!-- BEGIN/END GENERATED FACET CONTEXT (G530) -->`); the rest of
-`review-context.md` is hand-owned and untouched:
+same overlap logic `context collect`'s `--scope` uses (including the same
+rejected-reference visibility above), so the two surfaces can never
+disagree about what "overlaps". The generated content lives between two
+HTML-comment markers (`<!-- BEGIN/END GENERATED FACET CONTEXT (G530) -->`);
+the rest of `review-context.md` is hand-owned and untouched. Marker
+handling is fail-closed — mutation is attempted ONLY when the file carries
+EXACTLY one begin marker and one end marker in that order:
 
 - **File does not exist yet**: written fresh, in full, with the current
   `intent_references` (read from an EXISTING `packet.yaml` on disk if one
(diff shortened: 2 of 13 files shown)


tomohisa:
Request update — the shared selector and packet regeneration still miss required G530 cases.

1. Scope overlap is not the documented/packet-required symmetric normalized overlap. `HintMatchesPath` only accepts `candidate == hint` or `candidate.StartsWith(hint + "/")`. Consequently the documented domain-relative file form `identity/mission.md` does not match node id `identity/mission`; absolute paths are not reduced to repo/domain-relative form; the reverse ancestor case (node path is an ancestor of a deeper scope/reference) never matches; `.`/`..` segments are not normalized or rejected. This makes `context collect` and packet draft agree only by sharing the same false negatives. Normalize separator/path forms safely, reject traversal outside the domain, support absolute/repo-relative/domain-relative/intent-id forms, and implement segment-boundary overlap in both directions with case behavior explicitly pinned. Add `.md` short-form, absolute, reverse-ancestor, multiple hints, separator, traversal, case, and prefix-collision fixtures.

2. Empty/malformed comma lists silently broaden the query. `--facets ","` becomes an empty filter and therefore returns all four groups; `--scope ","` becomes no scope and returns every node. Empty elements such as `vocabulary,,decider` are also silently discarded. These must fail as malformed rather than weaken operator narrowing. Preserve whitespace trimming and first-seen dedup, but require every comma-delimited element to be non-empty and at least one value.

3. Malformed and unknown facet declarations disappear with no visibility. The selector skips malformed nodes and strips unknown values, and `DomainHasAnyFacetNodes` can then produce the same "No facet-annotated nodes found" note as a genuinely unannotated domain. G530 consumption must remain fail-closed/read-only, but JSON and Markdown need structured warnings/exclusions with path and reason so agents know context was omitted; packet draft's generated section needs the same visibility. Unknown values must never be bucketed, but cannot be silent.

4. Normal packet regeneration silently leaves Facet context stale. A fresh `packet draft` creates both template `packet.yaml` (empty references) and `review-context.md`. After the operator adds real `intent_references` to packet.yaml—the ordinary workflow—rerunning `packet draft` skips the existing review-context wholesale, so the newly relevant facet nodes are never generated. The current positive test only covers an unusual packet.yaml-existing/review-context-missing state. Preserve hand edits, but update a clearly delimited generated Facet context block (or provide an explicit safe regeneration mechanism) so newly authored references are reflected without overwriting surrounding hand content; test first scaffold → edit references → rerun, hand edits around the generated block, removed/changed references, and no-overwrite behavior outside that block.

Accepted portions: canonical facet group order and multi-facet representation are deterministic; valid filtering and unknown option errors work; summary extraction uses the first body line after frontmatter; id/facets/summary/path are emitted; JSON/Markdown shapes and ja/en docs are present; no G531 scope or mutation was found. Focused tests pass 56/56, full tests pass 3386 with 1 skip, CI is green, exact head is `2f6b200ab2cad856e274d86694e4c36543efdf47`, and `git diff --check` is clean.

tomohisa:
Request update — two review-repair boundaries remain silent/unsafe.

1. Invalid scope hints match nothing, but are not visibly diagnosed. `NormalizeHintSegments` returns `null` for outside-root absolute paths, traversal, malformed paths, etc.; `Select` immediately filters those nulls out and exposes only facet-declaration warnings. Thus `--scope /outside/root` or an all-traversal hint list produces empty groups indistinguishable from a valid scope that happens to match no nodes. Packet `intent_references` with invalid paths are likewise silently omitted. Preserve the safe match-nothing behavior, but add structured scope warnings containing the original hint and rejection reason/path classification to JSON, Markdown, and generated packet output. Mixed valid+invalid lists must use valid hints while still reporting every rejected hint; all-invalid lists must explicitly say all hints were rejected. Pin absolute-outside, relative traversal, mixed separators, malformed/encoded traversal where supported, and mixed/all-invalid cases.

2. Generated-block marker handling is not fail-closed and does not preserve line-ending style. `WriteOrUpdateReviewContext` finds the first begin marker and the first end marker after it, without requiring exactly one correctly ordered pair. Duplicate pairs/markers can therefore partially update one block and leave another stale; an end-before-begin plus later end can update across unintended hand content; one missing marker silently returns `skipped` with no diagnostic. The replacement also hardcodes `\n`, producing mixed line endings in an existing CRLF file rather than byte-preserving its style. Before mutation, require exactly one begin and one end marker in correct order; duplicate/reversed/one-sided shapes must return a non-success/fail-closed result with the file unchanged. Detect and reuse the existing file's newline convention for the generated replacement, preserving all bytes outside the pair. Add exact marker-count, duplicate, reversed, begin-only, end-only, CRLF, dry-run, and no-mutation fixtures. Legacy files with zero markers may remain untouched/skipped as documented, but malformed marker presence must not look like a healthy legacy file.

Everything else from the prior request is resolved: path forms normalize to domain-root segments, outside/traversal never broaden, case is explicitly ordinal, overlap is symmetric and segment-boundary safe, empty packet references match nothing, comma lists reject empty elements and deduplicate, malformed/unknown facet declarations produce path+reason warnings, regeneration updates changed/removed references and preserves surrounding hand edits, canonical order/fields/filter/summary remain correct, and no G531 scope was added. Focused tests pass 77/77; full tests pass 3407 with 1 skip; CI is green; exact head is `4e7c38d5a15c28ade17a4e5aa404893b1fd17770`; `git diff --check` is clean.

tomohisa:
Request update — one traversal boundary remains.

`NormalizeHintSegments` rejects `..` only after the absolute-path branch. For an absolute hint such as `<domainRoot>/identity/../identity/mission.md` (including the mixed-separator equivalent after `Replace('\\', '/')`), `Path.GetFullPath` canonicalizes away the traversal before the segment loop sees it. The hint is therefore accepted and can match a node, while the G530 contract requires every traversal-bearing hint to be rejected with its original text and a stable reason. This also makes absolute and repo/domain-relative forms disagree on the same path shape.

Reject `..` path segments on the normalized original hint before filesystem canonicalization (without weakening the existing outside-root check), surface the normal scope warning/all-invalid behavior, and pin absolute-under-root traversal plus mixed-separator absolute traversal in selector, context-collect JSON/Markdown, and packet generated-block tests. The rejected hint must match nothing.

All other requested repairs are resolved: warning order/original text/reasons and mixed/all-invalid narrowing are correct; marker classification permits only one ordered pair and leaves malformed/legacy files untouched; CRLF/LF regeneration and idempotence are pinned; accepted G530 ordering, fields, facets, summary, docs/schema, and no-G531 boundary remain intact. Focused tests pass 93/93; the full suite passes 3423 with 1 skip; CI is green; `git diff --check` is clean; exact head is `029580f5e47a6b009a945dcca7704923d91bb1ad`.

PR #1173, opened 2026-07-19
G535: request-update supersedes a stale intent-pr-rereview-ready

### LINKED ISSUE #1172: G535: request-update transition atomically supersedes a stale intent-pr-rereview-ready label
# G535: request-update transition atomically supersedes a stale intent-pr-rereview-ready label

## Goal

Make `automation pr-transition --transition request-update` remove a stale `intent-pr-rereview-ready` label in the same write, so a design amendment arriving while a PR is rereview-ready leaves the PR in a state `worker claim` accepts — closing the canonical-state deadlock window where no installed command can proceed.

## Why This Slice Exists Now

Field finding #5 (design@sekiban-as-a-service-orch, 2026-07-18, SKS-G824 / PR #1760): request-update does not clear `intent-pr-rereview-ready`, while `worker claim` refuses any PR still carrying rereview-ready. A design amendment arriving in the rereview-ready window therefore produced a PR that request-update had marked for repair but claim refused to touch — a deadlock between two canonical rules. The field orchestrator recovered with a non-obvious sequence (review-start, then request-update), but the window is a standing footgun that will recur every time design amends during a rereview wait.

## Current Observed State

- `request-update` adds its repair labels but leaves `intent-pr-rereview-ready` in place.
- `worker claim` (correctly) refuses PRs carrying rereview-ready.
- The combined state `rereview-ready + request-update` is reachable through canonical transitions and is then un-claimable; only the review-start→request-update detour clears it.

- The claim refusal on rereview-ready-without-request-update is CORRECT and stays (a rereview-ready PR is the reviewer's to pick up, not the worker's).
- Label mutations flow only through canonical transitions; raw `gh` edits remain forbidden.
- Supersession principle: a transition that puts a PR into a repair state supersedes pending rereview-readiness; two mutually-refusing labels must never coexist after a canonical transition.
- No other transition changes its label set in this slice.

## Target Repo / Path / Part

Target part: `automation pr-transition` request-update label writes + transition docs.

## In Scope

- request-update clears `intent-pr-rereview-ready` atomically (same label write) when present; transition output names the superseded label for the audit trail.
- Regression fixture reproducing the SKS-G824 / PR #1760 window: rereview-ready PR → request-update → claim succeeds.
- Fixture pinning that claim still refuses rereview-ready-without-request-update.
- All existing transition fixtures stay green (no other label-set changes).
- ja/en pr-transition docs: the supersession rule.

## Out Of Scope

- Any other transition or label vocabulary change; claim-side acceptance of combined states (the fix is supersession at the transition, not relaxation at claim).
- stalled-work classification of repair states (G533).
- Queue surfaces (G534).

The child PR makes `automation pr-transition --transition request-update` remove a stale `intent-pr-rereview-ready` label in the same write when present, with the superseded label named in the transition output; a regression fixture reproduces the SKS-G824 PR #1760 deadlock window and proves `worker claim` accepts the PR afterwards; another fixture pins that claim still refuses rereview-ready PRs that have NOT been transitioned to request-update; all existing transition fixtures remain green; ja/en developer reference documents the supersession rule. No other transition, claim, queue, or stalled-work changes.

## Acceptance Criteria

- Fixture: PR with `intent-pr-rereview-ready` → request-update → rereview-ready gone in the same write, repair labels present, `worker claim` succeeds.
- Transition output names the superseded label.
- Fixture: claim refuses a rereview-ready PR without request-update (unchanged behavior).
- Existing transition fixtures all green.
- ja/en docs updated; focused tests green.

## Verification

- Run the focused pr-transition and claim test suites and show them green.
- Manual run of the deadlock-window scenario on a fixture PR.
- `git diff --check` clean.

## Related Links

- Field finding #5: agmsg message from design@sekiban-as-a-service-orch, 2026-07-18 11:45Z (SKS-G824 / PR #1760; recovery via review-start then request-update).
- G525 — transition-vocabulary completeness direction; G533 — sibling classification slice (repair states in stalled-work).
- intents/intent-cli/intent-tree/means/04-worker-interface-strategy.md

## Knowledge Maintenance

- Intent placement: `intents/intent-cli/intent-tree/means/04-worker-interface-strategy.md` (existing node).
- ADR candidate: none — single-transition label correctness under an existing principle.
- Diagram candidate: none.
- Closeout writeback expected: no — the field recovery sequence becomes unnecessary once landed.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
## Summary

- `automation pr-transition --transition request-update` now clears a stale `intent-pr-rereview-ready` (and its legacy `rereview-ready` string form) in the **same** label write that adds `intent-pr-request-update` and removes `intent-pr-reviewing`.
- Fixes field finding #5 (SKS-G824 / PR #1760): a design amendment arriving while a PR was rereview-ready previously left the PR in a deadlock — `request-update` marked it for repair but `worker claim` (correctly) refuses any PR still carrying `rereview-ready`, and no installed command could break the cycle except a non-obvious `review-start` → `request-update` detour.
- `--write` mode only removes labels actually present (matching the existing convention for `review-start`/`approved`/`review-release`), so reruns stay idempotent; `--dry-run` still reports the full planned removal set. The add+remove edit remains a single `gh` call, so a failure leaves labels completely untouched (no half-transition).
- `worker claim` itself is unchanged — it still refuses a rereview-ready PR that has not been through `request-update`.
- No other transition's label set changed.

## Test plan

- [x] New fixture reproducing the exact SKS-G824/PR#1760 sequence: rereview-ready → `request-update` → `worker claim` succeeds (verified via `WorkerClaimAnalyzer` on the post-transition label set)
- [x] Fixture pinning `worker claim` still refuses an untransitioned rereview-ready PR
- [x] Idempotent rerun (rereview-ready already absent) fixture
- [x] API-failure fixture proving no half-applied transition
- [x] Regression fixture pinning `review-start`/`approved` label sets are byte-identical to before this change
- [x] All existing `AutomationPrTransitionCommandTests` (17) still green
- [x] Full solution test suite: 3579 CLI tests + all other projects passed, 1 pre-existing skip, 0 failures
- [x] Manual end-to-end verification against the built CLI binary with a stubbed `gh` reproducing the real label-mutation shell-out: claim refuses before the transition, `request-update --write` clears rereview-ready and names it in the output, claim succeeds afterward
- [x] ja/en developer-reference docs updated with a new G535 section

Closes #1172

### DIFF
--- a/docs/en/09-developer-reference.md
+++ b/docs/en/09-developer-reference.md
@@ -612,17 +612,38 @@ with no installed command able to proceed. The only escape was a non-obvious
 `review-start` → `request-update` detour.
 
 `request-update` now clears `intent-pr-rereview-ready` (and its legacy
-`rereview-ready` string form) in the **same** label write that adds
+`rereview-ready` string form) in the **same** write that adds
 `intent-pr-request-update` and removes `intent-pr-reviewing` — a repair
-request always supersedes pending rereview-readiness. As with the other
-transitions that clear multiple labels (`review-start`, `approved`,
-`review-release`), `--write` mode only removes labels that are actually
-present, so a rerun (or a PR that was never rereview-ready) stays
-idempotent; `--dry-run` still reports the full planned removal set
-regardless of presence, matching the existing convention. Because the
-add/remove edit is issued as a single `gh` call, the transition stays atomic
-— a failure leaves the PR's labels completely untouched, never a
-half-applied state. No other transition's label set changed.
+request always supersedes pending rereview-readiness.
+
+**Truthful audit output in both modes.** Unlike `review-start`/`approved`/
+`review-release` (whose `--dry-run` intentionally reports the *full*
+planned removal set regardless of presence), `request-update`'s reported
+`remove_labels` — in **both** `--dry-run` and `--write` — is always derived
+from the already-fetched current labels: a PR carrying only
+`intent-pr-rereview-ready` is reported as superseding exactly that label,
+never an absent `intent-pr-reviewing` or absent legacy `rereview-ready`
+alongside it. A rerun (or a PR that was never rereview-ready) reports and
+applies an empty removal set — genuinely idempotent, not merely
+non-erroring.
+
+**One atomic GitHub request, not sequential add/remove.** `gh <kind> edit
+--add-label --remove-label` is a CLI convenience wrapper — its atomicity
+from GitHub's perspective is not guaranteed. `request-update`'s `--write`
+path instead computes the full desired label set (every current label
+minus the ones being superseded, plus `intent-pr-request-update`) and
+replaces it in **one** GitHub REST call — `PUT
+/repos/{repo}/issues/{number}/labels` — via the internal
+`IGitHubLabelSetReplacer.ReplaceLabelSet` seam. A failure in that single
+call leaves the PR's labels completely untouched; there is no window where
+one label lands and another doesn't. GitHub's "Set labels" endpoint has no
+conditional/If-Match support for optimistic concurrency, so a concurrent
+label change racing the read-then-write cannot be prevented outright —
+instead, the adapter re-reads the labels immediately after the PUT and
+throws if the result doesn't match the desired set exactly, rather than
+silently claiming success on a lost update. This atomic-replace path is
+used only by `request-update`; every other transition keeps using the
+pre-existing `ApplyLabelTransitions` add/remove path, unchanged.
 
 With this landed, the SKS-G824 recovery sequence (`review-start` then
 `request-update` to clear a stuck rereview-ready) is no longer necessary —
--- a/docs/en/09-developer-reference.md
+++ b/docs/en/09-developer-reference.md
@@ -634,16 +634,49 @@ path instead computes the full desired label set (every current label
 minus the ones being superseded, plus `intent-pr-request-update`) and
 replaces it in **one** GitHub REST call — `PUT
 /repos/{repo}/issues/{number}/labels` — via the internal
-`IGitHubLabelSetReplacer.ReplaceLabelSet` seam. A failure in that single
-call leaves the PR's labels completely untouched; there is no window where
-one label lands and another doesn't. GitHub's "Set labels" endpoint has no
-conditional/If-Match support for optimistic concurrency, so a concurrent
-label change racing the read-then-write cannot be prevented outright —
-instead, the adapter re-reads the labels immediately after the PUT and
-throws if the result doesn't match the desired set exactly, rather than
-silently claiming success on a lost update. This atomic-replace path is
+`IGitHubLabelSetReplacer.ReplaceLabelSet` seam. If the desired set already
+equals the current set (order-insensitive), this is a genuine no-op — zero
+GitHub calls, not merely an empty removal list. This atomic-replace path is
 used only by `request-update`; every other transition keeps using the
 pre-existing `ApplyLabelTransitions` add/remove path, unchanged.
+
+**Phase-aware failure reporting — stated honestly, never overclaiming
+safety.** A single HTTP call means there is no window where *this call's
+own actions* land half-applied — but that is not the same as every failure
+being known-harmless, and the command's error reporting reflects that
+distinction precisely:
+
+- if the `gh` process for the PUT never starts (e.g. the executable can't
+  be launched), nothing was transmitted — reported as a plain failure,
+  `applied: false`, `may_have_applied: false`;
+- once that process has started, ANY failure (non-zero exit, a
+  write/read error, a timeout) is ambiguous — `gh` may already have
+  transmitted the request and GitHub may already have applied it before
+  the failure surfaced. Reported as `applied: false`, **`may_have_applied:
+  true`**, with the `intended_labels` the mutation was attempting to
+  establish and an exact `recovery_command` (`gh <kind> view <n> --repo
+  <repo> --json labels`) to resolve the ambiguity — never a false "nothing
+  changed" claim;
+- if the PUT itself reports success but the post-write verification read
+  fails, or succeeds but reads back a mismatched set, both are *also*
+  reported as `may_have_applied: true` with the same recovery info — never
+  as a rollback or "no mutation" signal, since the PUT very likely applied
+  in both cases.
+
+**Bounded concurrency model — the honest limit of the post-write
+verification.** GitHub's "Set labels" endpoint has no conditional/If-Match
+support for optimistic concurrency, so a label change racing between the
+caller's initial read (used to compute the desired set) and the PUT cannot
+be prevented outright. The post-write verification read only detects a
+mismatch that is *still present at the moment of that read*. A label added
+by another process **after** the initial read but **before** the PUT — and
+therefore never reflected in the desired set — can be silently overwritten
+by the PUT; if nothing else changes labels between the PUT and the
+verification read, that read will equal the intended set exactly and the
+command will report success even though a concurrent addition was just
+lost. This race is fundamentally undetectable by a read-after-write check
+alone, and the docs and code say so explicitly rather than implying full
+protection.
 
 With this landed, the SKS-G824 recovery sequence (`review-start` then
 `request-update` to clear a stuck rereview-ready) is no longer necessary —
--- a/docs/en/09-developer-reference.md
+++ b/docs/en/09-developer-reference.md
@@ -598,6 +598,40 @@ edits.
 
 ---
 
+### `request-update` supersedes a stale `intent-pr-rereview-ready` (G535)
+
+Field finding #5 (SKS-G824 / PR #1760): `intent-cli automation pr-transition
+--transition request-update` added its repair labels (`intent-pr-request-update`,
+and cleared `intent-pr-reviewing`) but left a pre-existing
+`intent-pr-rereview-ready` in place. `worker claim` correctly refuses any PR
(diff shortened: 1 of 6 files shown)


tomohisa:
G535 implementation finding — the transition does not yet satisfy the exact audit/atomicity contract.

1. `ResolveRemoveLabelsForMode` filters present labels only in `--write` mode. In dry-run it returns the entire planned set, so a PR carrying only `intent-pr-rereview-ready` is reported as also removing absent `intent-pr-reviewing` and legacy `rereview-ready`. The required dry-run/write output must name every superseded label that is present without claiming absent removals. Please derive the effective removal set from the already-fetched current labels in both modes and pin JSON plus text/Markdown output for present-current, present-legacy, combined, and absent/rerun cases. Update ja/en docs, which currently explicitly document the incorrect full-plan dry-run behavior.

2. The claimed no-half-transition evidence is not sufficient: `ThrowingMutator` throws before changing any state, while production still delegates add/remove to `gh pr edit`; one CLI invocation is not proof that the underlying GitHub operations are transactional. The packet requires one atomic GitHub label edit and an API failure with no half-transition. Please make the adapter issue one replace/update operation for the complete final label set (preserving unrelated labels), or otherwise provide a production-boundary mechanism whose single-request atomicity is demonstrable, and add adapter/failure fixtures that would catch remove-succeeded/add-failed (and add-succeeded/remove-failed) behavior.

Packet/AC basis: G535 In Scope and AC require request-update to clear stale rereview-ready in the same write with truthful audit output; the review dispatch additionally requires no preliminary/remedial call and no half-transition on API failure. Focused transition/claim tests currently pass (45/45) and `git diff --check` is clean, but those are necessary rather than sufficient because the fixtures encode the two gaps above.

tomohisa:
Repaired both blockers.

1. **Truthful audit output**: `ResolveRemoveLabelsForMode` now derives `request-update`'s reported/applied removal set from the already-fetched current labels in **both** `--dry-run` and `--write` (previously only `--write` filtered; `--dry-run` reported the full planned set regardless of presence — unlike the other transitions, this is a G535-specific correctness requirement, not just idempotency). A PR carrying only `intent-pr-rereview-ready` is now reported as superseding exactly that label, never an absent `intent-pr-reviewing` or absent legacy `rereview-ready` alongside it. Pinned with JSON output for present-current-only, present-legacy, combined, and absent/rerun cases (text output shares the same `Summary`/`RemoveLabels` plumbing, so it's covered transitively — this command has no Markdown output format).

2. **Atomic single-request replacement**: `request-update --write` no longer delegates to `gh <kind> edit --add-label/--remove-label` (a CLI convenience wrapper whose underlying HTTP atomicity was never actually proven) for its own mutation. It now computes the full desired label set (current labels minus what's being superseded, plus `intent-pr-request-update`) and replaces it via **one** GitHub REST call — `PUT /repos/{repo}/issues/{number}/labels` — through a new `IGitHubLabelSetReplacer.ReplaceLabelSet` seam, implemented only by `GhCliGitHubLabelMutator` and used only by `request-update` (no other transition or any of the ~10 existing `IGitHubLabelMutator` fakes across the test suite needed to change). Since GitHub's "Set labels" endpoint has no conditional/If-Match support for optimistic concurrency, the adapter re-reads labels immediately after the PUT and throws if the result doesn't match the desired set — documented in code and docs as the bounded concurrency model, rather than silently claiming atomic success on a lost update.

Test plan:
- New adapter-level tests (`GhCliGitHubLabelMutatorReplaceLabelSetTests`, 7): exactly-one-PUT-call shape assertion (no `edit` calls anywhere in the sequence), mutation HTTP failure (never attempts the verify re-read), verify-re-read failure (current-label fetch failure), verify-re-read mismatch (concurrent-change failure), policy guard, plus static `Build*Arguments`/`Build*RequestBody` shape tests — all using an injected `GhInvokerOverride` seam against the *real* adapter class, not a fake `IGitHubLabelMutator`.
- Updated/new `AutomationPrTransitionCommandTests` (command-layer, using a `FakeMutator` implementing the new interface): exact desired-set assertions replacing the old add/remove assertions, unrelated-label preservation, combined stale labels (current + legacy), idempotent rerun (empty removal, single replace call with nothing extra), API failure leaves the original label set completely untouched, dry-run calls neither mutation path, review-start/approved label-set-unchanged regression.
- Full solution suite: 3588 CLI tests + all other projects passed, 1 pre-existing skip, 0 failures.
- `git diff --check` clean.
- Manually verified against the built CLI binary with a stubbed `gh` (real `pr view`/`api --method PUT --input -` shape): dry-run now reports only the present `intent-pr-rereview-ready`; write issues exactly one PUT call whose body is `{"labels":["intent-pr-request-update","intent-target","priority-high"]}` — the unrelated `priority-high` label survives untouched.

Head SHA: a0925c0e206818718b8d1bf148f9218cec65ca36. CI green.

tomohisa:
G535 remaining implementation finding — failure/concurrency semantics still overclaim safety.

The single Set-labels PUT and truthful present-label delta resolve the earlier core findings, but the production/docs/tests currently state that any PUT failure leaves labels untouched (`ReplaceLabelSet` XML/docs and `MutationFailure_ThrowsAndNeverAttemptsVerifyRead`). That is not knowable when the server applies the PUT but the response is lost or `gh` exits after transmission. Likewise, a label added concurrently after the command initial read but before its PUT can be overwritten; the subsequent read will equal the intended set and therefore cannot detect that lost update. The current docs incorrectly imply the post-verify catches this case.

Please make failure reporting phase-aware: a pre-request construction/start failure may be known-unapplied, but once the PUT may have been sent, return nonzero without `applied: true` while explicitly saying the atomic set may already have landed and operator re-read/retry is required. A post-PUT read failure or mismatch must give that same unambiguous may-have-applied status; order-insensitive comparison is correct. Correct ja/en docs and fixtures to describe the bounded race honestly: post-verify detects only state differing at verification time, not an unrelated label overwritten before PUT. Also make an already-converged request-update (`desired set == current set` order-insensitively) a genuine no-op that avoids the PUT, while retaining truthful empty removal output.

Contract basis: final G535 rereview dispatch requires accurate may-have-applied reporting, honest read→PUT concurrency limitation, post-verify semantics, and bounded idempotent no-op. Focused adapter/transition/claim tests pass 54/54 and `git diff --check` is clean, but the current assertions encode the false zero-mutation guarantee.

tomohisa:
Repaired the remaining truthfulness gap.

**Phase-aware failure semantics.** `IGitHubLabelSetReplacer.ReplaceLabelSet` no longer treats every PUT failure as "definitely nothing changed." It now returns `LabelSetReplacementCertainty` (`NoOpAlreadyConverged` / `AppliedAndVerified`) on a known-good outcome, and throws `LabelSetReplacementException` carrying a `LabelSetReplacementFailureCertainty` for anything else:

- `KnownUnapplied` — only when the `gh` process for the PUT never actually started (nothing was ever transmitted).
- `MayHaveApplied` — once that process has started, *any* subsequent failure (non-zero exit, mid-transmission I/O error, timeout) is ambiguous, since `gh` may already have transmitted the request and GitHub may already have applied it.
- `AppliedButVerificationReadFailed` — the PUT itself reported success, but the post-write verification read failed.
- `AppliedButMismatchedOrConcurrentlyChanged` — the PUT reported success and the verify read succeeded, but the labels didn't match — the write itself very likely still applied.

`AutomationPrTransitionCommand` surfaces this as structured output in both JSON and text: `applied: false` together with `may_have_applied: true`, the `intended_labels` the mutation was attempting to establish, and an exact `recovery_command` (`gh <kind> view <n> --repo <repo> --json labels`) — never a bare "failed, nothing changed" claim once the outcome is genuinely unknown.

**Idempotent no-op.** Before issuing any PUT, `ReplaceLabelSet` now compares the (order-insensitive) desired set against the already-fetched current set; if they already match, it returns `NoOpAlreadyConverged` and makes **zero** GitHub calls — not merely an empty `remove_labels` list with a redundant PUT underneath.

**Honest concurrency documentation.** Both code comments and the ja/en docs now state the bounded limitation explicitly: the post-write verification only detects a mismatch still present at the moment of that read. A label added by another process *after* the caller's initial read but *before* the PUT is invisible to the desired-set computation and can be silently overwritten; if nothing else changes labels before the verification read, that read will equal the intended set exactly and the command will report success even though a concurrent addition was just lost. This is stated as a fundamental limitation of read-after-write verification, not something the design claims to fully solve.

Test plan:
- New adapter-level tests (8 total, up from 7): already-converged no-op (zero gh calls), gh-process-never-starts → `KnownUnapplied`, post-send failure → `MayHaveApplied` (never attempts the verify read), verify-read failure → `AppliedButVerificationReadFailed`, verify mismatch → `AppliedButMismatchedOrConcurrentlyChanged`, policy-guard refusal is a plain (non-ambiguous) exception, plus the existing shape/atomicity tests.
- New/updated command-layer tests: theory covering all three ambiguous certainties reporting `may_have_applied: true` with the exact recovery command, text-format parity, known-unapplied failure explicitly does NOT carry the ambiguity markers, idempotent rerun now asserts zero `ReplaceLabelSet` GitHub calls (previously asserted a redundant no-op call).
- Full solution suite: 3595 CLI tests + all other projects passed, 1 pre-existing skip, 0 failures.
- `git diff --check` clean.
- Manually verified against the built CLI binary with a stubbed `gh`: an already-converged PR issues zero `gh api` calls; a simulated post-transmission non-zero exit correctly reports `may_have_applied: true` with the exact intended labels and recovery command, while the call log confirms the PUT really was transmitted — proving the report is honest rather than a guess.

Head SHA: d6cc4d053c28f396c524793fd68458fe0c01f13f. CI green.

Issue #1166, opened 2026-07-19
G531: Add read-only intent facet-check scaffold matching change proposals against vocabulary and invariant facets


tomohisa:
# G531: Add read-only intent facet-check scaffold matching change proposals against vocabulary and invariant facets

## Goal

Add a read-only `intent facet-check --domain <d> (--packet <execution-unit> | --terms <list>) --format json` surface that extracts candidate command/event terms from a change proposal (a packet directory, or an explicit term list), and reports: related facet nodes (vocabulary and invariant first), collision candidates against existing vocabulary, and acceptance-property coverage gaps for the packet scope. It is review scaffolding that guides humans — explicitly NOT a semantic verifier and never a blocking gate (exit 0 regardless of findings).

## Why This Slice Exists Now

Operator input (issue #1159) pillar 3: proposals that add commands/events should be checkable against the existing vocabulary and invariants from the intent-tree side, so that reviewers and design threads see contradictions and coverage gaps early. G529 (classification) and G530 (context supply) make facets exist and reach agents; this slice closes the initiative loop by pointing a proposal AT the facets. Scope discipline learned from the field: a lexical/structural scaffold with honest limits beats a pretend semantic verifier — false negatives are acceptable, silent claims of verification are not.

## Current Observed State

- No surface relates a change proposal to intent-tree content; reviewers reconstruct vocabulary/invariant impact by hand (e.g. the SekibanAsAService G812 inventory comment enumerated ~59 affected surfaces manually).
- After G529/G530, facet nodes are classifiable and deliverable but nothing consumes them for proposal checking.

## Accepted Baseline You May Assume

- G529 is merged (`facets` schema + lint/search/analyze). G530 may land before or after; no dependency either way beyond G529.
- Read-only; exit code 0 regardless of findings (a scaffold, not a gate); CI/reviewers cite its report, they do not gate on it.
- Term matching is lexical (exact + near-identical, e.g. case/`-`/`_` normalized); no embeddings or semantic engines.
- Packet layout and validation are unchanged.

## Target Repo / Path / Part

Repository: `J-Tech-Japan/intent-system`

Target paths: `src/IntentSystem.Cli/**, tests/**, docs/ja/**, docs/en/**`

Target part: New `intent facet-check` read-only surface + developer-reference docs.

## In Scope

[...]

PR #1175, opened 2026-07-19
G536: Publish-flow idempotent rerun independently verifies and restores durable state

### LINKED ISSUE #1174: G536: publish-flow idempotent rerun must fully restore durable state or fail loud — never report synced on partial state
# G536: publish-flow idempotent rerun must fully restore durable state or fail loud — never report synced on partial state

## Goal

Fix the `issue publish-flow` idempotent rerun path so it verifies each durable artifact independently against the already-created GitHub issue — queue item `linked_issue`, runs `issue-created` event, `publish.yaml` entry — restores whatever is missing, and reports `durable_state_synced=true` ONLY when everything verifies. When restoration is impossible, fail loud naming the missing artifacts and the recovery command. A synced claim must never be a false positive.

## Why This Slice Exists Now

Field incident (intent-cli-dev orchestrator, 2026-07-19, publishing G530 as issue #1164): host main advanced concurrently after issue creation, forcing a stash + ff-sync mid-publish. The idempotent rerun then reported `durable_state_synced=true` while the queue `linked_issue` and the runs `issue-created` event were in fact missing from the synced state. The orchestrator had to hand-restore the CLI-generated linkage from the preserved stash — a manual durable-state repair the tooling exists to prevent — and explicitly requested this packet. A false-positive synced claim is the worst class of publish defect: stalled-work, WIP gating, and closeout linkage recovery all trust it.

## Current Observed State

- Rerun after issue creation checks something short of the full artifact set; in the field sequence it declared sync while `linked_issue` and the runs event were absent.
- The exact recovery required knowing that the pre-stash state contained the CLI-generated linkage and merging it by hand into the advanced main (host commits 94a44bd6, 1173cc4d).
- `publish-recovery` exists for create-halfway failures but did not flag this partial state.

- The publish lifecycle stages and boundaries (draft → create → publish-flow → automation issue-publish) are unchanged; this slice hardens the rerun/verification path only.
- Concurrent host-main advance during a publish is a legitimate scenario (multi-domain hosts see cross-domain pushes), not operator error.
- Fail-loud with exact missing-artifact names beats silent partial success.

## Target Repo / Path / Part

Target part: `issue publish-flow` idempotent rerun verification/restoration + `publish-recovery` consistency + docs.

## In Scope

- Independent verification of queue `linked_issue`, runs `issue-created`, and `publish.yaml` against the created issue on every idempotent rerun; restoration of missing artifacts within the rerun.
- `durable_state_synced=true` only on full verification; otherwise fail loud with the missing-artifact list and recovery command.
- Regression fixture reproducing the field sequence: create issue → reset durable state to pre-publish snapshot (stash+ff-sync simulation) → rerun restores linkage + runs event → only then synced.
- `publish-recovery` reports the same partial states consistently.
- ja/en docs: idempotent rerun artifact checklist.

## Out Of Scope

- Publish-order/priority selection (G537).
- New publish stages or label-boundary changes.
- Generic git conflict handling (stash/ff-sync mechanics stay the callers concern; this slice fixes the truthfulness of the rerun that follows).

The child PR hardens the `issue publish-flow` idempotent rerun: it independently verifies queue `linked_issue`, runs `issue-created`, and `publish.yaml` against the already-created GitHub issue, restores missing artifacts in the rerun, reports `durable_state_synced=true` only when all verify, and fails loud naming missing artifacts and the recovery command otherwise; `publish-recovery` reports the same partial states consistently; a regression fixture reproduces the issue #1164 field sequence (create → durable-state reset → rerun restores → synced); ja/en developer reference documents the artifact checklist. No publish-stage, label, or ordering changes.

## Acceptance Criteria

- Field-sequence regression fixture: after durable-state reset, the rerun restores `linked_issue` and the runs event, then reports synced; before restoration completes it never reports synced.
- With restoration impossible (e.g. read-only fixture), the command exits non-zero naming exactly the missing artifacts and the recovery command.
- `publish-recovery` on the same fixture reports the identical gap set.
- Full-artifact happy path unchanged (existing publish fixtures green).
- ja/en checklist docs present; focused tests green.

## Verification

- Run the focused publish-flow/publish-recovery suites including the regression fixture and show them green.
- Manual replay of the field sequence on a fixture host.
- `git diff --check` clean.

## Related Links

- Field incident: orchestrator agmsg report 2026-07-19 04:04Z (G530 / issue #1164; host commits 94a44bd6, 1173cc4d).
- G537 — sibling slice (publish-order override, same wake report).

## Knowledge Maintenance

- Intent placement: `intents/intent-cli/intent-tree/means/03-state-and-audit-strategy.md` (existing node).
- ADR candidate: none — truthfulness hardening of an existing contract.
- Diagram candidate: none.
- Closeout writeback expected: no — drop the stash-restore manual recovery from operational notes once landed.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
## Summary

- `issue publish-flow`'s idempotent-rerun path no longer trusts a single short-circuit signal (whichever of `publish.yaml`/`queue-state.json` fires first) as proof that all durable state is in sync. It now independently verifies all three durable artifacts — queue-state's `linked_issue`, `publish.yaml`'s `issue-created` record, and `runs.jsonl`'s `issue-created` event — and restores whichever are missing, using the same write helpers the first-run success path already uses.
- Fixes field incident from 2026-07-19 (publishing G530 as issue #1164): a concurrent host-main advance forced a stash+ff-sync mid-publish, leaving `publish.yaml` intact but queue-state's `linked_issue` and the runs event both reverted. The prior idempotent rerun reported `durable_state_synced: true` anyway — a false positive that required manual hand-restoration from the preserved stash.
- `durable_state_synced: true` is now reported only once all three artifacts verify (whether already correct or freshly restored this run); `gh` is never re-invoked on any idempotent rerun.
- A genuine contradiction between `publish.yaml` and `queue-state.json` (different issue numbers) fails loud rather than silently trusting either side.
- When restoration is impossible, the command exits non-zero naming exactly the missing/inconsistent artifacts and the recovery command (`issue publish-flow ... --write` retry, or `automation publish-recovery ... --write`).
- `automation publish-recovery` reports the identical gap on the same fixture shape (as an unsafe stop, since it requires PR evidence publish-flow doesn't need for its own prior record) — proving consistency rather than silence.
- ja/en developer-reference docs updated with the new artifact-verification checklist.

## Test plan

- [x] Field-incident regression fixture: create → reset queue-state/runs.jsonl to pre-publish snapshot (stash+ff-sync simulation) → rerun restores both from publish.yaml, then reports synced — verified against the real production binary with a stubbed `gh`
- [x] Restoration-impossible fixture: exits non-zero, names the exact missing artifact and recovery command
- [x] Contradiction fixture: publish.yaml and queue-state disagree on issue number → refuses, mutates nothing
- [x] Idempotent-restoration-is-itself-idempotent fixture: a third rerun after restoration makes zero further writes
- [x] `automation publish-recovery` consistency fixture: same durable-state shape reports the same execution unit as a gap (unsafe stop), not silence
- [x] All pre-existing publish-flow/publish-recovery fixtures updated/still green
- [x] Full solution test suite: 3600 CLI tests + all other projects passed, 1 pre-existing skip, 0 failures
- [x] Manual replay of the exact field sequence against the built CLI binary

Closes #1174

### DIFF
--- a/docs/en/09-developer-reference.md
+++ b/docs/en/09-developer-reference.md
@@ -841,6 +841,46 @@ rewritten check:
   calling `gh issue create`**; **multiple** matches → fails closed,
   non-mutating, exit 1 — ambiguity is never resolved automatically.
 
+**Round-5 review repair — GitHub enumeration is real cursor pagination, not
+a raised limit; body normalization is stricter; queue-state cardinality is
+checked.** A further review round found that a fixed `--limit` (however
+high) is still a cap that can, in principle, silently drop a real
+duplicate once the filtered result exceeds it — and that a blanket
+`Trim()` on the candidate/expected body could accept Markdown with
+different leading indentation (e.g. a code block) as if it were identical.
+
+- The GitHub existence check now uses `gh api graphql` with a real
+  `search(... first: 100, after: $cursor)` cursor-pagination loop —
+  `state=all` is preserved (open and closed issues both participate), and
+  the loop continues while `pageInfo.hasNextPage` is true, accumulating
+  every page rather than stopping at a fixed count. A page reporting
+  `hasNextPage: true` with no `endCursor`, or a result set that doesn't
+  terminate within an internal safety ceiling (50 pages / 5,000
+  candidates), fails loud (`InvalidOperationException`) instead of
+  silently truncating.
+- Body normalization now does **only** line-ending conversion (`\r\n`/`\r`
+  → `\n`) and treats "ends with exactly one trailing newline" as
+  equivalent to "no trailing newline" (GitHub's own storage convention) —
+  it no longer calls `Trim()`. Leading indentation, inner spacing, and any
+  trailing whitespace *before* the newline are preserved and compared
+  exactly, so a body that differs by so much as a single leading or
+  interior space is correctly treated as a **different** issue, not a
+  match.
+- `queue-state.json`'s `ReadQueueSignal` now collects **every** item
+  matching the execution unit rather than returning on the first match —
+  a second item for the same unit fails closed
+  (`queue_state_duplicate_execution_unit`, naming every matching index and
+  its identity) regardless of whether the duplicate entries agree or
+  conflict; identity must never depend on JSON array order.
+- New tests exercise the **real** `GhCliExistingIssueChecker` production
+  class end-to-end (not just the `IGitHubExistingIssueChecker` interface
+  stub used at the command level) via a `PageFetcherOverride` test seam
+  that replaces only the literal `gh` process spawn with canned GraphQL
+  JSON — covering multi-page open+closed accumulation, the two
+  fail-loud-on-truncation paths, and the body-normalization matrix
+  (byte-identical / CRLF vs LF / single-trailing-newline / leading /
+  inner / trailing whitespace drift).
+
 ---
 
 ### Facet-aware context supply (G530)
--- a/docs/en/09-developer-reference.md
+++ b/docs/en/09-developer-reference.md
@@ -920,6 +920,31 @@ classification.
   title, null body, and URL mismatches — wrong repo, wrong number, wrong
   scheme, or null).
 
+**Round-7 review repair — the last two provider fail-closed gaps: a
+non-null-but-empty cursor, and JSON `null` in place of a "non-nullable"
+shape.** A further review round found that `hasNextPage=true` with an
+**empty or whitespace-only** `endCursor` slipped past the null-only check
+from round 6 — it would be sent back as `cursor=` on the next request and
+recorded into the seen-cursors set as if it were a real value. Separately,
+System.Text.Json silently assigns `null` to a declared-non-nullable
+reference-type property when the JSON value is `null` (C# doesn't enforce
+non-null at runtime) — so `pageInfo: null`, `nodes: null`, or a `null`
+entry inside `nodes` would previously degrade into an incidental
+`NullReferenceException` rather than an intentional provider diagnostic.
+
+- `endCursor` is now checked with `string.IsNullOrWhiteSpace` — null,
+  empty, and whitespace-only are all treated as "missing cursor" and fail
+  loud before the loop would fetch again.
+- `pageInfo`, `nodes`, and each individual node are explicitly checked for
+  `null` immediately after parsing — a `null` in any of these positions
+  fails loud with a specific diagnostic naming which piece was missing,
+  never an unhandled NRE.
+- New tests pin the full malformed-shape matrix through
+  `FetchAllCandidates` itself: empty/whitespace-only `endCursor`, a
+  `null`/wrong-type top-level envelope (`null`, `[]`, a bare number, a
+  bare string), empty process output, `pageInfo: null`, `nodes: null`, and
+  a `null` entry inside `nodes`.
+
 ---
 
 ### Facet-aware context supply (G530)
--- a/docs/en/09-developer-reference.md
+++ b/docs/en/09-developer-reference.md
@@ -784,6 +784,63 @@ so an operator (or a test) can directly compare a `publish-recovery`
 unsafe stop against what `issue publish-flow`'s own rerun independently
 detects and restores for the identical durable state.
 
+**Round-4 review repair — the canonical identity is a full tuple, not just
+a number.** A subsequent review round found that treating "same issue
+number" as sufficient still let contradictory or self-inconsistent data
+through: two artifacts naming the same issue *number* but a different
+*repo*, or a single artifact whose own repo/number/URL fields disagreed
+with each other, were silently accepted. The analyzer now requires every
+present signal to be:
+
+- **Internally self-consistent** — `queue-state.json`'s `linked_issue`
+  (which carries `repo`, `number`, and `url` as three separate fields) and
+  each `runs.jsonl` event (which may carry both a `linked_issue`
+  `repo#number` descriptor and a `reason` issue URL) must agree with
+  themselves before they're even accepted as a signal at all.
+- **A canonical GitHub issue URL** — `https://github.com/<owner>/<repo>/issues/<number>`
+  exactly; any `/issues/`-containing string that doesn't match this shape
+  is no longer accepted as if it were canonical.
+- **Checked against the confirmed target repo directly**, not merely
+  pairwise between artifacts — a signal whose repo doesn't match the
+  `--repo` this command run is scoped to is a contradiction even when it's
+  the ONLY present signal.
+
+A malformed/unreadable `queue-state.json` now also fails closed exactly
+like `publish.yaml`/`runs.jsonl` already did (`queue_state_malformed`),
+rather than being silently treated as absent — a surviving `publish.yaml`
+or `runs.jsonl` signal must never authorize restoration around a
+`queue-state.json` this analyzer couldn't actually read. `publish.yaml`'s
+own `execution_unit` field is validated against the unit the packet path
+is scoped to (`publish_yaml_malformed` on mismatch — data copied from
+another unit's packet is corruption, not a signal for this unit).
+
+**The GitHub existence check is now a classification (zero / exactly-one /
+multiple), not a bare boolean, and requires exact title AND body
+linkage.** The prior round's `gh issue list --search ... --limit 20` with
+prefix-boundary title matching could both miss a real duplicate beyond the
+first 20 results and match a merely similarly-titled unrelated issue. The
+rewritten check:
+
+- Retrieves candidates with `--limit 1000` (vs. the prior round's 20) so a
+  legitimate duplicate is never silently dropped by client-side
+  truncation for any realistic repo's issue history — GitHub's own
+  `in:title` search is still used as a fast pre-filter, not the identity
+  decision itself.
+- Requires the candidate's title to equal the resolved expected title
+  **exactly** (no prefix/boundary heuristic — a similarly-titled unrelated
+  issue can never match).
+- Additionally requires the candidate's body to match the local packet's
+  `github-body.md` content byte-for-byte (normalized for line endings) —
+  the same content that was (or would be) posted via `gh issue create
+  --body-file`, giving a genuine content-linkage check rather than a title
+  guess alone.
+- Classifies the result: **zero** matches → safe to create; **exactly
+  one** → that confirmed GitHub identity is fed directly into the same
+  `PublishDurableArtifactAnalyzer`-backed restoration path used for a
+  local-signal rerun, restoring all three local artifacts **without ever
+  calling `gh issue create`**; **multiple** matches → fails closed,
+  non-mutating, exit 1 — ambiguity is never resolved automatically.
+
 ---
 
 ### Facet-aware context supply (G530)
--- a/docs/en/09-developer-reference.md
+++ b/docs/en/09-developer-reference.md
@@ -881,6 +881,45 @@ different leading indentation (e.g. a code block) as if it were identical.
   (byte-identical / CRLF vs LF / single-trailing-newline / leading /
   inner / trailing whitespace drift).
 
+**Round-6 review repair — the GraphQL provider now fails closed on
+authoritative-response defects, not just structural pagination gaps.** A
+further review round found that the checker still trusted an
+authoritative-looking response even when it wasn't: a GraphQL response can
+carry non-empty `errors` alongside otherwise-plausible `data`; a
+misbehaving server could return the same `endCursor` twice, looping the
+fetch forever short of the safety cap; the search `type: ISSUE` field
+actually matches both issues AND pull requests unless the query itself
+excludes PRs; and an individual candidate could be incomplete (null body,
+null/empty title, non-positive number, or a URL that doesn't exactly match
+the requested repo) without being rejected before it reached
+classification.
+
+- Every GraphQL response is checked for a non-empty `errors` array
+  **before** its `data` is ever read — the spec permits both to be present
+  simultaneously, and partial `data` alongside an error is never treated as
+  authoritative.
+- The search query now includes `is:issue` (`repo:<repo> <unit> in:title
+  is:issue`) so a similarly-titled pull request is excluded server-side
+  rather than risking an empty/default-deserialized node; `state:` remains
+  deliberately absent so both open and closed issues stay in scope. The
+  exact literal query string is pinned by a test.
+- Each page's `endCursor` is tracked in a seen-cursors set; a repeated
+  cursor value fails loud immediately (`InvalidOperationException`) rather
+  than looping until the 50-page safety cap.
+- Every fetched candidate is validated **before** it is accumulated (not
+  after, and never discovered only once classification or a restoration
+  write is already underway): a positive issue number, a non-null/
+  non-empty title, a non-null body (a null body is an invalid provider
+  response — never silently substituted with empty text), and a URL
+  matching the canonical `https://github.com/<requested repo>/issues/<number>`
+  shape **exactly** for the repo this check was scoped to.
+- New production-provider tests cover GraphQL errors alongside partial
+  data, repeated-cursor detection, the literal `is:issue`/no-`state:`
+  query pin, page-fetcher failure propagation, malformed JSON, and every
+  candidate-validation failure mode (non-positive number, null/empty
+  title, null body, and URL mismatches — wrong repo, wrong number, wrong
+  scheme, or null).
+
 ---
 
 ### Facet-aware context supply (G530)
--- a/docs/en/09-developer-reference.md
+++ b/docs/en/09-developer-reference.md
@@ -688,72 +688,101 @@ that has not been through `request-update`.
 
 ### `issue publish-flow` idempotent rerun independently verifies and restores all three durable artifacts (G536)
 
-Field incident (2026-07-19, publishing G530 as issue #1164): host `main`
-advanced concurrently after the GitHub issue was created, forcing a stash +
-fast-forward sync mid-publish. `publish.yaml` survived with its
-`issue-created` record intact, but `queue-state.json`'s `linked_issue` and
-the `runs.jsonl` `issue-created` event both reverted to their pre-publish
-(absent) state. The idempotent rerun that followed reported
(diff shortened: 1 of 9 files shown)


tomohisa:
G536 implementation finding — the rerun analyzer is incomplete and can still mutate on dry-run or create a duplicate issue.

The new `ExecuteIdempotentRerun` is entered only when publish.yaml or queue linked_issue is recognized. It ignores runs and GitHub-existing issue as authoritative identity sources. Therefore the required missing combination “only one canonical issue-created runs event survives” falls through to the normal create path and can create a second issue; zero/duplicate-identical/conflicting events are collapsed by `RunsLogHasIssueCreatedEvent` to a bool that checks only unit+event and ignores repo/number/URL. Malformed runs also becomes “missing” and is appended to rather than failing closed.

Additionally, the rerun helper does not receive the parsed `write` mode: it hardcodes `write: true` and calls queue/publish/runs write helpers unconditionally. An idempotent `--dry-run` with a missing artifact therefore mutates durable state. Its success is based on helper return values, with no post-write re-read of all three artifacts, so `durable_state_synced:true` is not proven after partial/faulted writes.

The required publish-flow/publish-recovery consistency is also not implemented as a shared analyzer/gap contract. The added publish-recovery test merely expects its existing unrelated `no-closing-pr-for-published-issue` unsafe stop; it neither compares the same three artifact gaps nor gives publish-flow and publish-recovery common identifiers/recovery semantics.

Required repair: build one shared, fail-closed analyzer that independently parses queue exact unit+repo+number+URL, publish unit/status/number/URL, and every canonical issue-created run event (zero / exactly one / duplicate-identical / conflicting), corroborates a noncontradictory identity with the existing GitHub issue without ever creating on rerun, and emits identical stable gaps to both surfaces. Dry-run must plan only. Write must restore every missing combination without duplicating the runs event, then re-read all three and set durable_state_synced only if the shared analyzer returns no gaps; malformed/unreadable and any cross-artifact contradiction must exit nonzero without choosing the first signal. Pin #1164 and #1166 reset/concurrent fixtures plus partial-write/retry faults and a fully converged rerun with zero create/mutation calls.

Packet/dispatch basis: G536 AC requires independent verification/restoration, no duplicate issue, shared recovery gap identifiers, fail-closed malformed data, and post-write verified truthfulness. Focused tests pass 55/55 and diff check is clean, but the tests currently omit these load-bearing combinations.

tomohisa:
Repair round 3 pushed at c7c120f (CI green: https://github.com/J-Tech-Japan/intent-system/actions/runs/29684826199).

- New shared, fail-closed `PublishDurableArtifactAnalyzer` — the single source of truth `issue publish-flow`'s idempotent rerun and `automation publish-recovery` both consult. It independently parses queue-state `linked_issue`, `publish.yaml`'s `issue-created` record, and every canonical `runs.jsonl` `issue-created` event (classified zero / exactly-one / duplicate-identical / conflicting), then resolves one canonical identity or fails closed.
- Fixes the runs-only-signal gap (field incident G531/#1166): a unit whose *only* surviving signal was `runs.jsonl` no longer falls through to the create path — the rerun trigger now checks all three artifacts uniformly via the shared analyzer.
- Malformed `publish.yaml` / `runs.jsonl` fail closed as a distinct `*_malformed` state, never silently collapsed into "missing" (which would invite an unsafe overwrite).
- Restoration now touches only genuinely-missing artifacts (per the analyzer's gap list), then **re-invokes the same analyzer a second time** against the freshly-written files before ever reporting `durable_state_synced: true` — a write helper's boolean return is never trusted alone.
- Dry-run reports a new `would_restore` field (the gap list a subsequent `--write` would restore) and never writes.
- New `IGitHubExistingIssueChecker` / `GhCliExistingIssueChecker` (`gh issue list --search`) corroboration check runs before `gh issue create` whenever the analyzer finds zero identity across all three local artifacts — refuses (never creates) if a title match already exists on GitHub, closing the last duplicate-creation path.
- `automation publish-recovery`'s unsafe stops now carry a `durable_artifact_gaps` field from the same shared analyzer; the G536 fixture test now asserts byte-for-byte parity between `publish-recovery`'s gaps and `issue publish-flow`'s `would_restore` for the identical durable state, replacing the previous same-command-only proxy assertion.
- New regression tests: the #1166 runs-only-signal case (proves zero duplicate `gh issue create` calls), conflicting vs duplicate-identical runs events, malformed publish.yaml/runs.jsonl, the GitHub existence-refusal path and its safe-to-create counterpart, and dry-run byte-for-byte no-mutation with `would_restore`.
- Full solution suite green (3608 passed, 1 pre-existing skip), `git diff --check` clean, ja/en developer-reference docs updated.

Ready for another review pass.

tomohisa:
G536 remaining implementation finding — canonical issue identity is still not authoritative/fail-closed.

The shared analyzer and dry-run/write reanalysis repair the previous structural gaps, but the analyzer currently treats issue number equality as sufficient and drops required identity fields:

- `ReadQueueSignal` catches malformed/unreadable queue JSON and returns absent, so a surviving publish/runs signal can cause the write path to overwrite/repair around unreadable queue state instead of failing closed. It also does not validate the linked_issue repo or canonical URL; same-number cross-repo linkage is accepted.
- `ReadPublishSignal` does not verify `execution_unit`, and cross-artifact comparison checks only distinct numbers. A publish artifact for another unit or URL/repo contradiction with the same number can be accepted.
- runs parsing accepts any text before the last `#`, treats arbitrary `/issues/<n>` URLs as canonical, and deduplicates/conflicts only by number. Thus `other/repo#1164`, the target repo #1164, and a mismatched reason URL can collapse to one “identical” identity instead of a contradiction.

The GitHub-only guard is also still loose: `gh issue list --search <unit> in:title --limit 20` returns only number/title, accepts any title prefix with a boundary, has no exact expected title/body linkage check, no all-result pagination beyond 20, and returns a bool rather than zero/one/multiple identities. A similarly prefixed unrelated issue can block a legitimate new issue, while an exact existing issue outside the first 20 can be missed and duplicated. It also only refuses; it cannot establish and restore the sole authoritative GitHub identity required by the rereview contract.

Please make the shared identity a canonical tuple (exact repo, positive number, canonical URL, execution unit) for queue/publish/runs, fail closed on malformed/unreadable queue as already done for publish/runs, compare every present field and reject same-number cross-repo/URL/unit contradictions. Replace the GitHub bool search with state=all exhaustive candidate retrieval and deterministic zero/exactly-one/multiple classification based on exact expected issue title plus execution-unit/body linkage; a similar title must not match. Feed the unique GitHub identity into the same analyzer so the GitHub-only case restores missing local artifacts without a create, while multiple/search failure remains nonmutating fail-closed. Pin sole queue/publish/runs/GitHub and all contradiction/cross-repo/similar-title/pagination cases.

The focused suites pass 63/63 and `git diff --check` is clean, but these identity holes violate the G536 requirement that no contradictory or weak signal can authorize recovery or duplicate creation.

tomohisa:
Repair round 4 pushed at 20419ad (CI green: https://github.com/J-Tech-Japan/intent-system/actions/runs/29685415628).

- Canonical identity is now a full **(exact repo, positive number, canonical GitHub issue URL, execution unit)** tuple, not "same number is enough."
- `queue-state.json`'s `linked_issue` (separate `repo`/`number`/`url` fields) and each `runs.jsonl` event (`linked_issue` descriptor + `reason` URL) are validated for **internal self-consistency** before being accepted as a signal at all.
- Any `/issues/`-containing string must match the canonical `https://github.com/<owner>/<repo>/issues/<number>` shape exactly — arbitrary URLs are no longer accepted as canonical.
- Every present signal's repo is checked **directly against the confirmed `--repo` target**, not merely pairwise between artifacts — a signal whose repo disagrees with the command's own scope is a contradiction even when it's the only signal present.
- `queue-state.json` now fails closed on malformed/unreadable content (`queue_state_malformed`), matching the existing `publish.yaml`/`runs.jsonl` behavior, instead of being silently treated as absent.
- `publish.yaml`'s own `execution_unit` field is validated against the packet path it's scoped to — data copied from another unit's packet now fails closed as `publish_yaml_malformed` rather than being accepted because the issue number looked plausible.
- The GitHub existence guard is replaced with a **zero / exactly-one / multiple** classification: candidates are retrieved with `--limit 1000` (vs. the previous round's 20), then filtered to an **exact** title match (no prefix/boundary heuristic) **and** exact local `github-body.md` content linkage (normalized for line endings) — a merely similarly-titled unrelated issue can never match. A unique match feeds its confirmed identity into the same `PublishDurableArtifactAnalyzer`-backed restoration path used for local-signal reruns (restoring local artifacts **without** ever calling `gh issue create`); multiple matches fail closed, non-mutating.
- New regression tests: cross-repo queue linked_issue, publish.yaml execution_unit mismatch, cross-repo runs.jsonl event, GitHub-unique-match restoration (no create call), and GitHub-multiple-match refusal.
- Full solution suite green (3612 passed, 1 pre-existing skip), `git diff --check` clean, ja/en developer-reference docs updated.

Ready for another review pass.

tomohisa:
G536 remaining implementation finding — GitHub enumeration/body normalization and queue cardinality are still not fail-closed.

1. `GhCliExistingIssueChecker` uses `gh issue list ... --limit 1000` and comments that 1000 is sufficient for a “realistic” repository. That is not exhaustive: when the filtered result exceeds the cap, an exact existing issue or a second exact duplicate can be omitted, changing None/Unique/Multiple and potentially authorizing a duplicate create. Use an API/GraphQL pagination loop with an explicit `hasNextPage`/cursor until exhaustion, or detect any incomplete/truncated result and fail loud. Keep state=all so closed published issues participate.

2. `NormalizeBody` calls `.Trim()`. The permitted GitHub normalization is line-ending conversion and trailing final newline handling; `Trim()` also removes leading spaces/tabs and trailing non-newline whitespace. Leading indentation can change Markdown semantics (e.g. an indented code block), so a semantically drifted GitHub body can be accepted as exact. Normalize CRLF/CR to LF and only the known trailing newline convention; preserve all other authored bytes. Pin originally submitted body acceptance plus leading/inner/trailing-space drift rejection.

3. `ReadQueueSignal` returns from the first matching execution-unit item and never detects a second queue item with the same unit. That makes identity depend on JSON order and can hide agreeing or conflicting duplicate entries. Collect all exact-unit items and fail closed on cardinality >1 with every index/identity, independent of whether the duplicate tuples agree.

The new tuple validation, malformed-file refusals, GitHub-only restore, similar-title filtering, shared gaps, dry/write/postverify, and focused tests otherwise look aligned. Please add production-provider tests (not only injected classification stubs) pinning multi-page open+closed enumeration, truncation/error, exact title/body normalization, and duplicate queue order. Focused suites pass 67/67 and `git diff --check` is clean; the three holes above remain capable of a false None/Unique identity decision.

tomohisa:
Repair round 5 pushed at fea67fd (CI green: https://github.com/J-Tech-Japan/intent-system/actions/runs/29685897571).

- **GitHub enumeration is now real cursor pagination, not a raised limit.** `GhCliExistingIssueChecker` shells to `gh api graphql` with a `search(... first: 100, after: $cursor)` loop, accumulating every page while `pageInfo.hasNextPage` is true (`state=all` preserved — open and closed issues both participate). A page reporting `hasNextPage: true` with no `endCursor`, or a result set that doesn't terminate within an internal 50-page/5,000-candidate safety ceiling, fails loud instead of silently truncating.
- **Body normalization is stricter.** Replaced the blanket `Trim()` with normalization limited to line-ending conversion (`\r\n`/`\r` → `\n`) plus treating "ends with exactly one trailing newline" as equivalent to "no trailing newline" (GitHub's own convention). Leading indentation, inner spacing, and any other trailing whitespace are now preserved and compared exactly — a body differing by a single leading/interior space is correctly rejected as a different issue, not accepted as a match.
- **`queue-state.json` cardinality is now checked.** `ReadQueueSignal` collects every item matching the execution unit instead of returning on the first match; a second item — whether it agrees or conflicts with the first — fails closed (`queue_state_duplicate_execution_unit`, naming every matching index/identity) rather than depending on JSON array order.
- **New production-provider tests** (`GhCliExistingIssueCheckerTests.cs`) exercise the real `GhCliExistingIssueChecker` class end-to-end via a `PageFetcherOverride` seam that replaces only the literal `gh` process spawn — not a hand-written interface stub — covering multi-page open+closed accumulation, both fail-loud-on-truncation paths, similar-title rejection, and the full body-normalization matrix (byte-identical / CRLF vs LF / single-trailing-newline / leading / inner / trailing whitespace drift).
- New regression test for duplicate queue-state execution-unit entries.
- Full solution suite green (3626 passed, 1 pre-existing skip), `git diff --check` clean, ja/en developer-reference docs updated.

Ready for another review pass.

tomohisa:
G536 remaining implementation finding — the GraphQL provider still does not fail closed on authoritative-response defects.

The cursor loop, narrow body normalization, and duplicate queue detection resolve the prior findings. However `GhCliExistingIssueChecker` still accepts incomplete/malformed authoritative data in several cases required by this rereview:

- `GraphQlResponse` models only `data`; a response containing GraphQL `errors` plus partial `data.search` is silently accepted. Any non-empty errors array must fail loud even when data is present.
- repeated `endCursor` values are not detected; the loop re-fetches the same page until the 50-page cap. Track seen cursors and fail immediately with a repeated-cursor diagnostic (retain the safety-cap failure too).
- search is `type: ISSUE` but the query lacks `is:issue`; GitHub search ISSUE covers issues and pull requests. The fragment yields no Issue fields for PR nodes, which can deserialize as empty/default entries rather than being explicitly excluded/fail-closed. Add `is:issue` to the search query and pin the literal production arguments. Open+closed remains correct by omitting state filtering.
- candidate `body` is nullable and classification substitutes null with empty text. A null body is an invalid provider response and must fail loud, not become an exact match when the expected body is empty or a harmless nonmatch otherwise. Likewise validate every candidate has positive number, exact canonical `https://github.com/<requested repo>/issues/<number>` URL, and non-null title before classification; do not write a fetched but malformed tuple and discover it only after local mutation/postverify.

Please extend the production-provider tests for partial-data GraphQL errors, null/malformed data/nodes/body/title/url/number, repeated cursor, query `is:issue`, nonzero process failure, and malformed JSON. Candidate accumulation and exact body/title matching should run only after every fetched page has passed these checks.

Focused G536 suites pass 81/81 and `git diff --check` is clean; all other reviewed contracts look resolved, but these provider cases can still turn an unreliable GitHub response into false None/Unique and authorize create or recovery.

tomohisa:
Repair round 6 pushed at 9eb7e5d (CI green: https://github.com/J-Tech-Japan/intent-system/actions/runs/29686296377).

- **GraphQL `errors` now checked before `data`.** A response can carry non-empty `errors` alongside otherwise-plausible partial `data` (the spec permits both); that partial data is never treated as authoritative — checked immediately after deserialization, before `search` is ever read.
- **Repeated-cursor detection.** Each page's `endCursor` is tracked in a seen-cursors set; a repeated cursor value (a misbehaving server re-serving the same page) fails loud immediately instead of looping until the 50-page safety cap.
- **`is:issue` added to the search query.** `type: ISSUE` in GraphQL search actually matches both issues and pull requests; the query is now `repo:<repo> <unit> in:title is:issue` so a similarly-titled PR is excluded server-side rather than risking an empty/default-deserialized node. `state:` remains deliberately omitted so open+closed issues stay in scope — the exact literal query string is now pinned by a test.
- **Every fetched candidate is validated before accumulation** (never discovered as malformed only after classification or a restoration write): positive issue number, non-null/non-empty title, non-null body (a null body is an invalid provider response, never silently substituted with empty text), and a URL matching the canonical `https://github.com/<requested repo>/issues/<number>` shape exactly for the repo this check was scoped to.
- Extended `GhCliExistingIssueCheckerTests.cs`: GraphQL errors alongside partial data, repeated-cursor detection, the literal `is:issue`/no-`state:` query pin, page-fetcher failure propagation, malformed JSON, and every candidate-validation failure mode (non-positive number, null/empty title, null body, URL mismatches for wrong repo/number/scheme/null).
- Full solution suite green (3642 passed, 1 pre-existing skip), `git diff --check` clean, ja/en developer-reference docs updated.

Ready for another review pass.

tomohisa:
G536 provider fail-closed coverage is still incomplete at the exact head. In FetchAllCandidates, hasNextPage=true rejects only a null endCursor; an empty endCursor is accepted, sent as cursor= on the next request, and entered into seenCursors. Treat null/empty (preferably whitespace-only as malformed too) as a missing cursor and fail loudly before another fetch, with a production-path regression test. Please also pin the delegated malformed-shape matrix through FetchAllCandidates itself (null/wrong-type envelope, null pageInfo/nodes/node, and empty output), so nullable GraphQL shapes cannot regress into incidental NullReferenceException behavior rather than an intentional provider diagnostic. The focused checker suite currently passes 29 tests and git diff --check is clean, but these required pagination/provider boundaries are not yet satisfied.

tomohisa:
Repair round 7 pushed at 9ce571c (CI green: https://github.com/J-Tech-Japan/intent-system/actions/runs/29686720576).

- **Empty/whitespace-only `endCursor` now rejected, not just null.** `hasNextPage=true` is now checked with `string.IsNullOrWhiteSpace(endCursor)` — an empty or whitespace-only cursor is a missing cursor just as much as `null`, and previously would have been sent back as `cursor=` on the next request and recorded into the seen-cursors set as if it were a real value.
- **Explicit `null` guards for `pageInfo`, `nodes`, and individual nodes.** System.Text.Json silently assigns `null` to a declared-non-nullable reference-type property when the JSON value is `null` (C# doesn't enforce this at runtime), so `"pageInfo": null`, `"nodes": null`, or a `null` entry inside `nodes` previously degraded into an incidental `NullReferenceException` rather than an intentional provider diagnostic. Each is now checked immediately after parsing and fails loud naming exactly what was missing.
- Extended `GhCliExistingIssueCheckerTests.cs` to pin the full malformed-shape matrix through `FetchAllCandidates` itself: empty/whitespace-only `endCursor`, a `null`/wrong-type top-level envelope (`null`, `[]`, a bare number, a bare string), empty process output, `pageInfo: null`, `nodes: null`, and a `null` entry inside `nodes`.
- Full solution suite green (3652 passed, 1 pre-existing skip), `git diff --check` clean, ja/en developer-reference docs updated.

Ready for another review pass.

PR #1177, opened 2026-07-19
G537: Canonical publish-order override — honor queue priority in next-slice selection

### LINKED ISSUE #1176: G537: Canonical publish-order override — honor queue priority in next-slice candidate selection
# G537: Canonical publish-order override — honor queue priority in next-slice candidate selection

## Goal

Make publish ordering an expressible decision: a bounded canonical transition sets the `priority` of a queued, not-yet-published execution unit (with a runs event recording the change and reason), and next-slice / publish candidate selection orders eligible candidates priority-first (high > normal > low) with authoring order as the tiebreaker — while dependency and WIP gates keep absolute precedence. With no priorities set, selection stays byte-identical to today.

## Why This Slice Exists Now

Field incident (intent-cli-dev orchestrator, 2026-07-19): after the G529 closeout, the orchestrator had ruled — with justification — to publish the field-impact fixes G532 and G534 ahead of the G530/G531 continuation. The canonical comm-based candidate selection chose G530 by authoring order, and NO installed surface could express the ruling. The orchestrator faced the forbidden choice (hand-edit ordering state or abandon the ruling), correctly abandoned it, and reported the gap. Queue items already carry a `priority` field (values like `high` observed in the field), but nothing consults it for publish selection — the schema promises what the selector ignores.

## Current Observed State

- Publish/next-slice candidate selection derives order from authoring order (comm-based backlog vs completed); the queue `priority` field is ignored.
- No transition mutates `priority` after enqueue; changing it would require the forbidden hand-edit.
- The 2026-07-19 wake report records the concrete unapplied ruling ("G534 could not be reordered ahead of it").

- Dependency gates, WIP cap, clarification gates, and lifecycle (retired) checks all dominate priority — priority only orders ALREADY-ELIGIBLE candidates.
- Design owns packet content; the orchestrator owns moving READY packets — prioritization of ready packets is squarely an orchestrator/operator decision, so a bounded transition (not a packet edit) is the right surface.
- Runs events are the audit trail for every durable mutation; a priority change without a recorded reason is not acceptable.
- G534 (lifecycle-aware selector) may land before or after; both touch selection but compose (retired-skip is an eligibility rule, priority is an ordering rule).

## Target Repo / Path / Part

Target part: Queue priority transition + next-slice/publish candidate ordering + docs.

## In Scope

- Bounded transition (e.g. `queue transition --execution-unit <id> --priority <high|normal|low> --reason <text> --write`, or an equivalent `queue reprioritize` surface — child picks the idiomatic shape): sets priority on a queued, not-yet-published unit; appends a runs event with old/new priority and reason; refuses on published/in-flight/completed/retired units; dry-run default.
- Selection ordering: priority class first, authoring order within class; fixture reproducing the field scenario (G530 authored earlier, G532 marked high → selector returns G532).
- Gate precedence fixture: high-priority unit with incomplete dependencies is not selected.
- No-priority regression: existing ordering fixtures byte-identical.
- ja/en docs: transition, ordering, precedence.

## Out Of Scope

- Priority influencing anything other than publish/next-slice candidate ordering (no scheduling, no WIP-cap changes, no review ordering).
- Automatic prioritization heuristics; priorities on packets rather than queue items.
- G536 (idempotent rerun) and G534 (eligibility) content.

The child PR adds a bounded canonical queue-priority transition (dry-run default; sets high/normal/low with a recorded reason on queued not-yet-published units only; runs event appended; refusals for published/in-flight/completed/retired) and makes next-slice/publish candidate selection order eligible candidates priority-first with authoring order as the in-class tiebreaker, while dependency/WIP/clarification/lifecycle gates keep absolute precedence. Fixtures reproduce the field reordering scenario (later-authored high-priority unit selected first), prove gate precedence over priority, and prove byte-identical selection when no priorities are set; ja/en developer reference documents the surface. No scheduling, WIP, review-order, or packet-schema changes.

## Acceptance Criteria

- Priority transition: sets priority with reason + runs event on a queued unit; refuses (naming why) on published/in-flight/completed/retired; dry-run lists the exact mutation.
- Field-scenario fixture: G530-analog authored earlier, G532-analog set high → selector returns the high-priority unit.
- Gate-precedence fixture: high priority never overrides incomplete dependencies or WIP/clarification/lifecycle gates.
- No-priority selection byte-identical to current behavior (existing fixtures).
- ja/en docs present; focused tests green.

## Verification

- Run the focused queue/selection suites including the three fixture classes and show them green.
- Manual replay of the field scenario on a fixture host.
- `git diff --check` clean.

## Related Links

- Field incident: orchestrator agmsg reports 2026-07-19 02:16Z (ruling) and 04:04Z (ruling unappliable; "G534 could not be reordered ahead of it").
- G534 — lifecycle-aware selector (composes: eligibility vs ordering); G536 — sibling from the same wake report.
- intents/intent-cli/intent-tree/means/01-queue-topology.md

## Knowledge Maintenance

- Intent placement: `intents/intent-cli/intent-tree/means/01-queue-topology.md` (existing node).
- ADR candidate: none — completes an existing schema field with a bounded transition; gate precedence preserves all stronger invariants.
- Diagram candidate: none.
- Closeout writeback expected: no — revisit orchestrator-thread guide publish-order wording after landing.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1176

## Summary
- Adds `queue reprioritize <execution-unit> --priority <high|normal|low> --reason <text> [--write]`: a bounded canonical transition that sets `priority` on a queued, not-yet-published execution unit, with a required reason and a `priority-changed` runs event. Dry-run by default; refuses (naming why) on any non-queued state or once a GitHub issue is linked; idempotent no-op when the requested priority matches the current one.
- `intent next-slice` now orders eligible candidates priority-class-first (high > normal > low), with authoring order (queue-state array order) as the in-class tiebreak. This is a stable sort applied to the existing `queued` bucket *before* the per-candidate gate loop runs — every existing gate (packet directory presence, execution-unit regex, domain/repo filter, G534 lifecycle exclusion, legacy retirement marker) still runs exactly as before, per candidate. Priority only changes try-order among candidates that already pass every gate; it never lets a candidate skip one.
- A host with no priorities meaningfully set (everything at the enqueue default `"normal"`) produces byte-identical output to pre-G537 behavior — proven by the full existing 72-test `IntentNextSliceCommandTests` suite passing unchanged, plus a new explicit regression test.

## Field incident this closes
2026-07-19: after the G529 closeout, the orchestrator ruled — with justification — to publish field-impact fixes G532/G534 ahead of a G530/G531 continuation. No installed surface could express that ruling (`priority` existed in the schema but nothing consulted it), so the orchestrator faced the forbidden choice of hand-editing queue-state or abandoning the ruling, correctly abandoned it, and reported the gap.

## Scope note for reviewers
`next-slice`'s existing candidate loop has no dependency (`Dependencies`/`BlockedBy`) gate today — that logic lives only in the separate, simpler legacy `QueueSelection.SelectNext` (`queue next`) surface, which this PR does not touch (out of scope per the field incident, which was specifically about next-slice/publish ordering). The "gate precedence" fixture therefore demonstrates precedence over next-slice's actual gates (G534's lifecycle exclusion) rather than a literal dependency gate, since next-slice doesn't have one to test against.

## Test plan
- [x] New fixtures: field-incident reorder scenario, gate-precedence (lifecycle exclusion beats priority), explicit no-priority-set regression
- [x] Full `QueueReprioritizeCommand` test matrix: dry-run reporting, `--write` mutation + runs event, refusal on non-queued state, refusal on already-published unit, refusal on unknown unit, refusal without `--reason`, refusal on unsupported priority value, idempotent no-op
- [x] Manual end-to-end replay of the field scenario against the built CLI binary (dry-run → write → next-slice re-selection)
- [x] Full solution test suite green (3663 passed, 1 pre-existing skip)
- [x] ja/en developer-reference docs updated

### DIFF
--- a/docs/en/09-developer-reference.md
+++ b/docs/en/09-developer-reference.md
@@ -1015,9 +1015,27 @@ is reversed — the runs event is appended **first**, `queue-state.json`
   fails, the audit trail already proves the attempted change and its
   reason even though the state file doesn't yet reflect it — never a
   silent, unaudited mutation. Re-running the exact same command detects
-  the already-recorded event (matched on execution unit + event name +
-  the deterministic reason text) and retries **only** the `queue-state.json`
+  the already-recorded event and retries **only** the `queue-state.json`
   write, so convergence never produces a duplicate event.
+
+**Round-2 review repair — the dedup match is bound to the queue-state
+generation, not merely the reusable transition text.** Matching only on
+execution unit + event name + the deterministic reason text has a real
+collision: replaying the exact same transition later (e.g. `normal→high`
+reason `R`, then `high→normal` reason `S`, then `normal→high` reason `R`
+again — a genuine third mutation) produces a reason string
+byte-identical to the first event's, so the naive dedup would wrongly
+treat that stale historical event as the pending audit for the third
+mutation and skip appending its own — violating "one reasoned event per
+mutation." The match now additionally requires the candidate event's
+`Ts` to be **at or after** the `UpdatedAt` value read from
+`queue-state.json` at the top of the current invocation. Every successful
+write this command makes advances `UpdatedAt` to the same timestamp used
+for its own event — so an event from any prior generation is necessarily
+older than the current `UpdatedAt` and can never be mistaken for a
+pending retry, while an event genuinely written moments ago by an
+immediately-preceding failed attempt (against this same still-unmutated
+generation) always satisfies the bound.
 
 ---
 
--- a/docs/en/09-developer-reference.md
+++ b/docs/en/09-developer-reference.md
@@ -947,6 +947,56 @@ entry inside `nodes` would previously degrade into an incidental
 
 ---
 
+### Canonical publish-order override — queue priority (G537)
+
+Field incident (2026-07-19): after the G529 closeout, the orchestrator
+ruled — with justification — to publish field-impact fixes G532/G534
+ahead of a G530/G531 continuation. `queue-state.json`'s `priority` field
+(values like `high` observed in the field) already existed, but no
+selection surface consulted it — the orchestrator faced the forbidden
+choice of hand-editing ordering state or abandoning the ruling, correctly
+abandoned it, and reported the gap.
+
+**`intent-cli queue reprioritize <execution-unit> --priority
+<high|normal|low> --reason <text> [--write]`** is the bounded canonical
+transition that closes this gap:
+
+- Only ever mutates a **queued, not-yet-published** item's `priority` —
+  refuses (no mutation, naming why) when the item's state isn't `queued`,
+  or when it already has a linked GitHub issue.
+- `--reason <text>` is required — a priority change without a recorded
+  reason is never permitted.
+- **Dry-run by default.** Without `--write`, the command reports the
+  exact mutation that would happen (old priority, requested priority,
+  whether anything would actually change) without touching
+  `queue-state.json`. `--write` is required to mutate and append the
+  `priority-changed` runs event (old/new priority plus the operator's
+  reason).
+- Requesting the item's current priority is a no-op (idempotent) — no
+  write, no runs event, `changed: false`.
+
+**`intent next-slice` orders eligible candidates priority-class-first
+(high > normal > low), with authoring order (queue-state array order) as
+the in-class tiebreak.** Every existing eligibility gate — packet
+directory presence, execution-unit namespace regex, domain/repo filter,
+G534's lifecycle-aware exclusion, and the legacy-retirement-marker check
+— still runs exactly as before, per candidate, inside the same loop;
+priority never lets a candidate skip a gate it would otherwise fail. It
+only reorders which already-eligible candidate is tried, and in which
+order, before that per-candidate gate loop runs. Because the reorder uses
+a **stable** sort, a host where every item carries the enqueue default
+(`"normal"`) — i.e. no priorities meaningfully set — produces
+byte-identical output to pre-G537 behavior.
+
+`QueueItem.Priority` remains a plain, unvalidated `string` at the schema
+level (unchanged) — `queue reprioritize` is the only writer that
+normalizes and validates it (`high`/`normal`/`low`, case-insensitive);
+`next-slice`'s ranking function treats any unrecognized/missing value as
+`normal` rather than erroring, so hand-authored or historical
+`queue-state.json` files never fail closed on this field.
+
+---
+
 ### Facet-aware context supply (G530)
 
 Building on G529's four semantic facets (`vocabulary`, `invariant`,
--- a/docs/en/09-developer-reference.md
+++ b/docs/en/09-developer-reference.md
@@ -1027,35 +1027,49 @@ reason string byte-identical to the first event's, so a naive dedup on
 execution unit + event name + reason text alone would wrongly treat that
 stale historical event as the pending audit for the third mutation.
 
-**Round-3 review repair — wall-clock ordering is not a safe generation
-marker; the dedup match is now a content fingerprint, not a timestamp
-comparison.** The round-2 `Ts >= UpdatedAt` bound breaks in three ways: at
-timestamp equality (`UtcNowFactory` fixed to one value across several
-operations — or genuine same-tick production writes) it cannot
-distinguish "my own immediately-preceding failed attempt" from an older,
-unrelated transition; a clock rollback can move a later write's
-`UpdatedAt` *behind* an earlier one; and the write path never guaranteed
-`changedAt` strictly advances past the current `UpdatedAt` in the first
-place. The generation marker is now a **SHA-256 digest of the exact
-pre-mutation `queue-state.json` bytes**, read at the top of the current
-invocation, appended to the recorded reason (e.g. `... (generation
-3f9a2b7c1d4e5f60)`). This needs no clock at all:
+**Round-3 review repair (superseded by round 4 below) — the dedup match
+was next bound to a SHA-256 content fingerprint of the pre-mutation
+`queue-state.json` bytes,** to eliminate all wall-clock dependence from
+round 2's `Ts >= UpdatedAt` bound (which broke at timestamp equality, on
+clock rollback, and because the write path never guaranteed `changedAt`
+strictly advances past `UpdatedAt` in the first place).
 
-- A genuine retry (failed attempt, then re-run) starts from the
-  IDENTICAL un-mutated file — the failed attempt never wrote it — so the
-  retry computes the exact same fingerprint and finds its own
-  already-recorded event, regardless of what any clock did in between.
-- Any other, older transition necessarily read `queue-state.json`
-  *before* at least one subsequent successful write changed its content
-  (this command's own transitions always change the item's `priority`
-  field, and typically `updated_at` too) — so its fingerprint differs
-  and can never collide with the current generation's fingerprint. Same-
-  tick collisions and clock rollback/future-skew are structurally
-  impossible to trigger, since no timestamp is compared at all.
-- A historical event with no fingerprint tag at all (data predating this
-  fix, or hand-edited) can never exact-match a freshly-computed tagged
-  reason, so it is correctly treated as unrelated history rather than a
-  pending retry.
+**Round-4 review repair — a content fingerprint identifies BYTES, and
+this state machine can revisit identical bytes; the dedup token is now a
+durable, injective `priority_revision` counter, not a fingerprint of
+anything.** The round-3 fingerprint is collision-resistant for different
+content, but genuinely revisitable: `normal→high(R)` then `high→normal(S)`
+under one fixed clock reproduces the *exact original file bytes* (same
+priority, same `updated_at`, same everything) — a real revisit, not a
+hypothetical one. A subsequent `normal→high(R)` request then computes the
+identical fingerprint and tagged reason as the first event, so the
+fingerprint-based dedup wrongly treats that stale first event as pending
+for the third, genuinely distinct mutation.
+
+`QueueItem` now carries `priority_revision` — a plain `int`, deliberately
+NOT `required`, so a legacy `queue-state.json` predating this field
+simply deserializes it as `0` (the correct migration semantics: revision
+counting starts from the first `queue reprioritize` ever applied to a
+given item). Every successful `--write` bumps it by exactly 1. The
+recorded reason now carries the `fromRevision->toRevision` pair (e.g.
+`... (revision 0->1)`), and the dedup match is an exact match on that
+tagged reason (plus execution unit + event name), same as before — only
+the tag's *source* changed:
+
+- `toRevision` can mathematically never be produced by two distinct
+  successful mutations of the same item: each mutation strictly consumes
+  the "next" integer in the durably-persisted sequence, and once
+  consumed it is never the "next" one again — **regardless of whether
+  every other field of `queue-state.json` later cycles back to
+  byte-identical content**, since the counter itself is part of that same
+  durable content and only ever moves forward.
+- A genuine retry (failed queue-state write, then re-run) reads the SAME
+  `fromRevision` from the still-unmutated file both times — the failed
+  attempt never wrote the bump — so it computes the identical
+  `fromRevision->toRevision` pair and finds its own already-recorded
+  event.
+- A historical event with no revision tag at all (data predating this
+  fix, or hand-edited) can never exact-match a freshly-tagged reason.
 
 ---
 
--- a/docs/en/09-developer-reference.md
+++ b/docs/en/09-developer-reference.md
@@ -979,12 +979,19 @@ transition that closes this gap:
 (high > normal > low), with authoring order (queue-state array order) as
 the in-class tiebreak.** Every existing eligibility gate — packet
 directory presence, execution-unit namespace regex, domain/repo filter,
-G534's lifecycle-aware exclusion, and the legacy-retirement-marker check
-— still runs exactly as before, per candidate, inside the same loop;
-priority never lets a candidate skip a gate it would otherwise fail. It
-only reorders which already-eligible candidate is tried, and in which
-order, before that per-candidate gate loop runs. Because the reorder uses
-a **stable** sort, a host where every item carries the enqueue default
+**dependency completeness / non-empty `blocked_by`** (review repair: the
+same rule `QueueSelection.SelectNext` already enforces — every
+`dependencies` entry must be `completed`, and `blocked_by` must be
+empty), G534's lifecycle-aware exclusion, and the legacy-retirement-marker
+check — still runs exactly as before, per candidate, inside the same
+loop; priority never lets a candidate skip a gate it would otherwise
+fail. A "high" priority queued unit with an incomplete dependency or a
+non-empty `blocked_by` is never selected ahead of an eligible
+lower-priority unit — the loop simply tries the next candidate in
+priority/authoring order, same as any other gate failure. Priority only
+reorders which already-eligible candidate is tried, and in which order,
+before that per-candidate gate loop runs. Because the reorder uses a
+**stable** sort, a host where every item carries the enqueue default
 (`"normal"`) — i.e. no priorities meaningfully set — produces
 byte-identical output to pre-G537 behavior.
 
@@ -994,6 +1001,23 @@ normalizes and validates it (`high`/`normal`/`low`, case-insensitive);
 `next-slice`'s ranking function treats any unrecognized/missing value as
 `normal` rather than erroring, so hand-authored or historical
 `queue-state.json` files never fail closed on this field.
+
+**Review repair — `queue reprioritize --write` uses a fail-closed,
+repairable write order.** Writing `queue-state.json` before appending the
+required `priority-changed` runs event could leave a durable priority
+mutation with no audit record if the append step then failed. The order
+is reversed — the runs event is appended **first**, `queue-state.json`
+**second**:
+
+- If the event append fails, `queue-state.json` is never touched at all
+  — no durable change happened, and a plain retry starts fresh.
+- If the event append succeeds but the `queue-state.json` write then
+  fails, the audit trail already proves the attempted change and its
+  reason even though the state file doesn't yet reflect it — never a
+  silent, unaudited mutation. Re-running the exact same command detects
+  the already-recorded event (matched on execution unit + event name +
+  the deterministic reason text) and retries **only** the `queue-state.json`
+  write, so convergence never produces a duplicate event.
 
 ---
 
--- a/docs/en/09-developer-reference.md
+++ b/docs/en/09-developer-reference.md
@@ -1107,6 +1107,29 @@ against a concurrent writer.**
   copy from the top of `Execute`), so an unrelated concurrent change to
   any *other* field or item is preserved rather than clobbered.
 
+**Round-6 review repair — the round-5 "re-read + compare" was still a
+TOCTOU check, not authoritative mutual exclusion.** Two concurrent
+invocations could both read the same `priority_revision`, both classify
+zero event claims, both append their own event, both re-read and see the
+still-unchanged revision, and both then commit — with identical requests
+that duplicates the audit trail; with different requests it silently
+produces last-writer-wins state with a conflicting orphaned event.
+
+`--write` now acquires a **non-blocking, OS-level exclusive lock**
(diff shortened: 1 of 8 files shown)


tomohisa:
Two G537 contract findings remain at exact head 134136111030ded68395d731101dbe713ccd4abc:

1. The lifecycle-only precedence fixture cannot replace the packet/issue-required dependency fixture. `.intent-cli/issues/G537/packet.yaml` explicitly requires that a high-priority unit with incomplete dependencies is never selected, and ICL.M.TOPOLOGY says the queue-aware supervisor considers dependency and priority. `IntentNextSliceCommand` currently never reads `Dependencies` or `BlockedBy`; after priority sorting, such a high-priority queued item can become `candidate`. Implement the dependency/BlockedBy eligibility gate on this actual next-slice path (or obtain an authored design clarification changing the AC), and pin high-priority blocked vs lower-priority ready fixtures. WIP/clarification/lifecycle must remain dominant too.

2. `queue reprioritize --write` writes queue-state before appending the required `priority-changed` event. If the runs directory/create/append fails, the command leaves a durable priority mutation with no required reason event. Add a fail-closed/repairable write strategy and fault fixtures proving the operator cannot silently receive an unaudited half-transition; idempotent retry must converge without duplicate events.

Focused next-slice/reprioritize tests pass (83/83), `git diff --check` is clean, exact head is stable, and CI is green; these are semantic contract blockers, so merge/closeout was not performed.

tomohisa:
Repair round 1 pushed at c1a6108 (CI green: https://github.com/J-Tech-Japan/intent-system/actions/runs/29687975607).

Addresses both findings from the previous review round:

1. **Authoritative dependency/blocked-by eligibility gate added to `IntentNextSliceCommand`.** The `queued` candidate loop now enforces the same rule `QueueSelection.SelectNext` already applies — every `Dependencies` entry must be `Completed`, and `BlockedBy` must be empty. A "high" priority queued unit that fails this gate is never selected ahead of an eligible lower-priority unit; the loop simply tries the next candidate in priority/authoring order, exactly like any other gate failure. New fixtures pin: a high-priority unit with an incomplete dependency (skipped, normal-priority unit selected instead), a high-priority unit with a non-empty `blocked_by` (same outcome), and the counterpart where the dependency has reached `completed` (the high-priority unit becomes eligible and correctly wins).

2. **`queue reprioritize --write` now uses a fail-closed, repairable write order.** The `priority-changed` runs event is appended **before** `queue-state.json` is written (reversed from the original order) — so a failure at either step never leaves a silent, unaudited mutation:
   - Event-append failure → `queue-state.json` is never touched; a plain retry starts fresh.
   - `queue-state.json` write failure (after the event was recorded) → the audit trail already proves the attempted change and its reason; re-running the *exact same* command detects the already-recorded event (matched on execution unit + event name + the deterministic reason text) and retries *only* the `queue-state.json` write — no duplicate event on convergence.
   - New fault-injection tests (via new `AppendPriorityChangedEventOverride`/`WriteQueueStateOverride` test seams) pin both failure points and the idempotent-retry convergence path.

Full solution suite green, `git diff --check` clean, ja/en developer-reference docs updated.

Ready for another review pass.

tomohisa:
The dependency/BlockedBy finding is resolved, but the event-first recovery still has a stale-event collision at exact head c1a61086bf0f916963be7ca1da885cfbaf0f3373. `RunsLogHasMatchingPriorityChangedEvent` treats any historical event with the same execution unit + event name + deterministic reason as the pending event from the immediately preceding failed write. Reproduce: complete normal→high with reason R; complete high→normal with reason S; request normal→high with reason R again. The first historical event matches, so the third durable priority mutation skips appending its own audit event. A stale identical transition is therefore misclassified as recovery, violating one reasoned event per mutation. Bind dedup to the specific queue-state generation/operation (not merely the reusable transition text), and add fixtures for this round-trip stale collision plus wrong reason, wrong direction, wrong unit, and the genuine immediate retry. Focused tests pass 89/89, CI is green, `git diff --check` is clean, and head is stable. Merge/closeout remains blocked until every new durable priority mutation has its own event while genuine retry still deduplicates.

tomohisa:
Repair round 2 pushed at a597f4f (CI green: https://github.com/J-Tech-Japan/intent-system/actions/runs/29688391818).

Fixes the stale round-trip collision: `RunsLogHasMatchingPriorityChangedEvent` matched only on execution unit + event name + the deterministic reason text, which is not enough to tell "the pending audit event from my own immediately-preceding failed attempt" apart from "a genuinely historical, fully-completed transition that happens to share the same old/new priority and reason text." Concretely: `normal→high` reason `R` completes, `high→normal` reason `S` completes, then `normal→high` reason `R` is requested again — a real, distinct third mutation whose reason text is byte-identical to the first event's. The naive dedup wrongly matched the stale first event and skipped appending the third mutation's own event, violating one-reasoned-event-per-mutation.

The match now additionally requires the candidate event's `Ts` to be **at or after** `queue-state.json`'s `UpdatedAt` as read at the top of the current invocation. Every successful write this command makes advances `UpdatedAt` to the same timestamp used for its own event, so:
- An event from any prior generation is necessarily older than the current `UpdatedAt` and can never be mistaken for a pending retry (fixes the exact collision above — the third mutation now correctly gets its own new event).
- An event genuinely written moments ago by an immediately-preceding failed attempt (against this same still-unmutated generation) always satisfies the bound, so the fail-closed/repairable convergence from round 1 is preserved.

New fixtures pin: the exact round-trip stale-collision reproduction (three distinct mutations → three distinct events), a stale event with a different reason text, a stale event recording the opposite direction, a stale event for a different execution unit (none of these ever suppress the new event), and a direct fixture for the genuine-retry convergence path.

Full solution suite green, `git diff --check` clean, ja/en developer-reference docs updated.

Ready for another review pass.

tomohisa:
The stale-event repair is still not generation-safe at exact head a597f4fd81dd87f656afd6340dd7048347f837d2. `Ts >= queueState.UpdatedAt` does not distinguish generations at timestamp equality, and the write path does not guarantee that `changedAt` strictly advances beyond the current `UpdatedAt`. Reproduce the same round trip with UtcNowFactory returning one identical timestamp for all three successful operations: normal→high(R), high→normal(S), normal→high(R). Queue UpdatedAt stays equal throughout, so the first event satisfies `>=` during operation three and suppresses the third audit event. Clock rollback/future-skew has the same structural problem: a successful write can move UpdatedAt backward, while a failed event older than a future UpdatedAt is not recognized on genuine retry. Use a strictly monotonic generation/operation identity (or equivalent unambiguous recovery marker), not non-strict wall-clock ordering. Pin same-Ts/multiple-events, future queue UpdatedAt + immediate retry, clock rollback, and malformed/missing generation handling; stale events must never pass and genuine retries must not duplicate. Dependency/BlockedBy and prior G537 repairs remain resolved. Focused tests pass 94/94, CI is green, diff check clean, and head stable; merge/closeout remains blocked on this audit correctness boundary.

tomohisa:
Repair round 3 pushed at 98e3bc6 (CI green: https://github.com/J-Tech-Japan/intent-system/actions/runs/29688910702).

The round-2 fix (`Ts >= UpdatedAt`) was still wall-clock ordering and broke in exactly the three ways flagged: at timestamp equality (`UtcNowFactory` fixed to one value across operations, or genuine same-tick production writes) it couldn't distinguish a pending retry from an unrelated older transition; a clock rollback could move a later write's `UpdatedAt` behind an earlier one; and the write path never guaranteed `changedAt` strictly advances past the current `UpdatedAt` at all.

**The generation marker is now a SHA-256 digest of the exact pre-mutation `queue-state.json` bytes**, computed at the top of the current invocation and appended to the recorded runs-event reason (e.g. `... (generation 3f9a2b7c1d4e5f60)`). This eliminates wall-clock dependence entirely:
- A genuine retry (failed attempt, then re-run) starts from the identical un-mutated file — the failed attempt never wrote it — so the retry computes the exact same fingerprint and finds its own already-recorded event, regardless of what any clock did in between.
- Any other, older transition necessarily read `queue-state.json` *before* a subsequent successful write changed its content (this command's own transitions always change the item's `priority` and typically `updated_at` too) — so its fingerprint can never collide with the current generation's fingerprint, no matter what any clock did.
- A historical event with no fingerprint tag at all (data predating this fix, or hand-edited) can never exact-match a freshly-computed tagged reason.

New/updated fixtures: the review's exact counter-example (all three operations under one **identical, fixed** clock value — still produces three distinct events), a clock-rollback scenario, stale events differing by reason/direction/execution-unit re-seeded with *matching* fingerprints to isolate each dimension independently, a historical event missing the fingerprint tag entirely, and the genuine-retry convergence fixture re-seeded with a matching content fingerprint instead of a matching timestamp.

Full solution suite green, `git diff --check` clean, ja/en developer-reference docs updated.

Ready for another review pass.

tomohisa:
The SHA-256 prestate fingerprint is still not an injective operation/generation identity at exact head 98e3bc65845f44d9e0c76b8472f34fd5f3e40f52. It identifies content, and this state machine can revisit identical content. The new same-fixed-clock fixture does not prove otherwise: BuildQueueState starts with UpdatedAt in May while fixedNow is July, so after normal→high(R), high→normal(S), the returned normal state has July UpdatedAt and is not byte-identical to the original May prestate. Set initial UpdatedAt equal to fixedNow (or continue another full high/normal cycle at the same clock): serializer output, priority, UpdatedAt, and all other fields return to identical bytes, so the original fingerprint and tagged R reason recur; the later normal→high(R) then reuses the stale first event and suppresses its own audit. A content hash is collision-resistant for different bytes but cannot distinguish separate visits to the same bytes. Use a durable injective operation/generation token or another protocol that can prove a specific event is pending rather than historical. Pin an assertion that the revisited prestate bytes are exactly equal before exercising the repeated transition, alongside missing fingerprint, wrong unit/reason/direction, clock skew, and genuine fault retry. Dependency gates and other G537 AC remain resolved. Focused tests pass 97/97, CI is green, diff check clean, and head stable; merge/closeout remains blocked.

tomohisa:
Repair round 4 pushed at 1c28703 (CI green: https://github.com/J-Tech-Japan/intent-system/actions/runs/29689363955).

Confirmed and fixed: a content fingerprint identifies *bytes*, and this state machine can genuinely revisit identical bytes — `normal→high(R)` then `high→normal(S)` under one fixed clock reproduces the exact original file (same priority, same `updated_at`, everything). The prior round's fixture accidentally avoided this because the initial seed's `UpdatedAt` (May) differed from the fixed clock (July); it did not prove the fingerprint was safe in general.

**Replaced the fingerprint with a durable, injective `QueueItem.PriorityRevision` counter** — a plain `int`, deliberately not `required`, so legacy `queue-state.json` files without the field simply deserialize it as `0` (correct migration semantics: revision counting starts from the first `queue reprioritize` ever applied to a given item). Every successful `--write` bumps it by exactly 1, and the recorded reason now carries the `fromRevision->toRevision` pair (e.g. `... (revision 0->1)`).

`toRevision` can mathematically never be produced by two distinct successful mutations of the same item: each mutation strictly consumes the next integer in the durably-persisted sequence, and once consumed it's never reused — **regardless of whether every other field of `queue-state.json` later cycles back to byte-identical content**, since the counter itself is part of that same durable content and only ever moves forward. A genuine retry (failed write, then re-run) reads the same `fromRevision` from the still-unmutated file both times and finds its own already-recorded event.

New fixtures pin exactly what was asked:
- The revisited-prestate reproduction, with an **explicit assertion** that the prestate bytes are exactly equal (aside from `priority_revision` itself) before exercising the repeated transition — proving the revisit is real and the fix still produces three distinct events.
- Clock rollback.
- Stale events differing by reason/direction/execution-unit, re-seeded with *matching* revision pairs to isolate each dimension independently.
- A historical event missing the revision tag entirely.
- Genuine-retry convergence.
- Legacy migration (no `priority_revision` field at all → treated as 0).

Full solution suite green, `git diff --check` clean, ja/en developer-reference docs updated.

Ready for another review pass.

tomohisa:
The durable revision direction is sound, and the revisited-state fixture now correctly proves equality of every field except priority_revision. Two protocol boundaries remain blocking at exact head 1c28703295aa3e8dab1c6e6d03c987e0a4bc9e89:

1. `PriorityRevision` is an unconstrained `int`; negative values deserialize successfully, and `fromRevision + 1` is unchecked. At `int.MaxValue` it wraps to `int.MinValue`, directly violating the documented monotonic/injective invariant and eventually permitting revision-pair reuse. Fail closed before any event or queue mutation on revision < 0 or exhausted/max revision, with malformed/negative/max fixtures. Legacy missing=0 remains valid.

2. Recovery uses `events.Any(...)`. Two identical matching N→N+1 events are silently accepted as one pending attempt, while a conflicting event claiming the same unit/revision pair with different direction/reason is ignored and a new event is appended. The revision pair is the operation identity, so cardinality and ownership must be validated: exactly zero means append, exactly one exact match means retry, duplicate-identical or any conflicting claim for that pair must fail closed with queue untouched and actionable diagnostics. Pin reversed/order-independent duplicates and conflicts.

Also protect the read→event→queue write boundary from a stale concurrent queue writer, or explicitly fail closed via an authoritative compare/lock before committing so unrelated queue changes cannot be overwritten after the event was derived. Dependency/BlockedBy and all earlier G537 repairs remain resolved. Focused tests pass 97/97, CI is green, diff check clean, and head stable; merge/closeout remains blocked.

tomohisa:
Repair round 5 pushed at 7a84861 (CI green: https://github.com/J-Tech-Japan/intent-system/actions/runs/29689888858).

Three protocol boundaries fixed:

1. **`PriorityRevision` input validation.** It was an unconstrained `int` — negative values deserialized successfully, and `fromRevision + 1` was unchecked arithmetic that would silently wrap to `int.MinValue` at `int.MaxValue`, violating the monotonic/injective invariant. Both dry-run and `--write` now validate `PriorityRevision >= 0` and compute `toRevision` with `checked` arithmetic **before** previewing or mutating anything — a negative or exhausted revision fails closed with no event and no queue-state write.

2. **Explicit recovery classification, replacing `events.Any(...)`.** Since the revision pair *is* the operation identity, a bare existence check silently accepted two identical events claiming the same pair as "one pending attempt" (masking a real duplication bug), and silently ignored a genuinely conflicting event (same pair, different reason/direction), appending a second, different event right past it. Recovery is now classified explicitly: **zero** matches → safe to append; **exactly one exact match** → the pending audit for an in-progress retry; **two or more identical matches**, or **any mismatched-reason match** on the same pair → fails closed (exit 1, queue-state untouched, naming the conflicting/duplicate event).

3. **Concurrent-writer protection on the read → event → queue-write boundary.** Immediately before the final `queue-state.json` write, the file is now re-read fresh; if the target item's `priority_revision` no longer equals the `fromRevision` this attempt started from, the write refuses (the audit event is already durably recorded, so this is never silent) rather than blindly overwriting whatever a concurrent writer produced. The mutation is applied onto that **fresh** re-read, not the stale copy from the top of `Execute`, so an unrelated concurrent change to any other item/field is preserved rather than clobbered.

New fixtures: conflicting same-pair events (differing reason, opposite direction, and an order-independent variant), duplicate-identical same-pair events, negative/exhausted revision in both dry-run and write, malformed `priority_revision` type, and a concurrent-writer simulation proving the final write refuses and never clobbers the concurrent change.

Full solution suite green, `git diff --check` clean, ja/en developer-reference docs updated.

Ready for another review pass.

tomohisa:
Negative/exhausted/malformed revision validation and event-claim cardinality are resolved, but concurrency remains blocking at exact head 7a84861c25a925d4cdffa748d50b0a9059957c29. The implementation performs a fresh re-read and revision comparison, then a separate `File.WriteAllText`; that is still a TOCTOU check, not an authoritative CAS/lock. Two writers A/B can both read revision N, both classify no event claim, both append events, both perform the final re-read while the file still shows N, and both then write N+1. With identical requests this leaves duplicate audit events; with different reason/priority it leaves conflicting claims and last-writer-wins state. The fixture only simulates a writer completing before this invocation final re-read, so it cannot reproduce the compare-pass/write race. Hold an interprocess lock across authoritative re-read + event classification/append + final state commit, or use a true atomic compare-and-swap protocol. Tests must coordinate two real command invocations so both reach the old compare point, and prove only one mutation/event wins while the loser fails closed. Preserve unrelated concurrent fields by rebuilding under the same lock/current state; a pre-lock fresh copy is insufficient. Focused tests pass 104/104, CI is green, diff check clean, head stable; dependency gates and all other G537 AC remain resolved. Merge/closeout remains blocked.

tomohisa:
Repair round 6 pushed at 3ff0994 (CI green: https://github.com/J-Tech-Japan/intent-system/actions/runs/29690346411).

Confirmed: the round-5 "re-read + compare" immediately before the final write was still a TOCTOU check, not authoritative mutual exclusion. Two concurrent invocations could both read the same `priority_revision`, both classify zero event claims, both append their own event, both re-read and see the still-unchanged revision, and both then commit — duplicating the audit trail on identical requests, or producing silent last-writer-wins state plus a conflicting orphaned event on different ones.

**`--write` now acquires a non-blocking, OS-level exclusive lock** (`FileShare.None` on a stable sibling file next to `queue-state.json`, e.g. `queue-state.reprioritize.lock`) **before** the authoritative queue-state/runs.jsonl read, and holds it across revision validation, event-claim classification/append, the fresh re-read, and the final commit — released only once the invocation is completely done.

- A second, concurrent invocation that cannot acquire the same lock fails closed **immediately** (no wait, no retry) rather than racing to the compare point at all.
- Dry-run never mutates and never takes the lock.
- The round-5 fresh-re-read-and-rebuild is retained *underneath* the lock as defense-in-depth against a non-cooperating writer that bypasses the lock entirely (e.g. a different tool directly mutating `queue-state.json`) — the lock is what makes two *cooperating* `queue reprioritize` invocations mutually exclusive.

New tests: a deterministic two-invocation lock-contention fixture (via a new `OnLockAcquiredForTest` seam that runs a second, fully independent `Execute` call while the first still holds the lock — proving the loser fails closed and exactly one mutation/event lands, without depending on genuine OS-thread-scheduling nondeterminism), and a dry-run regression proving it never contends for the lock.

Full solution suite green, `git diff --check` clean, ja/en developer-reference docs updated.

Ready for another review pass.

tomohisa:
The OS-level FileShare.None lock now closes the cooperating-writer TOCTOU race, but one explicit exception-release boundary remains at exact head 3ff0994d8287208dadb95f2ea38a8b8660746fbd. After successful `TryAcquireLock`, `OnLockAcquiredForTest?.Invoke()` runs before entering the `try/finally` that disposes `lockStream`. If that callback throws, the acquired handle is not deterministically disposed; a subsequent independent Execute can remain locked out until GC/finalization. Move every post-acquisition operation, including the callback, inside the try/finally (or use an immediate `using` scope), and pin: callback throws, first call exits exceptionally, second call immediately acquires the same lock and succeeds, with queue/runs unchanged by the failed first call. This is the requested using/finally exception-release guarantee. Contention, dry-run no-lock, stale lock-file/open-handle semantics, revision validation/cardinality, dependency gates, and prior G537 repairs otherwise check out. Focused tests pass 106/106, CI is green, diff check clean, head stable; merge/closeout remains blocked on deterministic lock release.

tomohisa:
Repair round 7 pushed at ce1cd06 (CI green: https://github.com/J-Tech-Japan/intent-system/actions/runs/29690718179/job/88202822049).

Fixed the exception-release boundary flagged in round 7: `OnLockAcquiredForTest?.Invoke()` previously ran *before* entering the `try`/`finally` that disposes the acquired `FileStream` lock handle. If that callback (or any future post-acquisition step placed ahead of the `try`) threw, the OS-level lock would leak — a subsequent independent invocation would stay locked out until GC/finalization eventually closed the handle, an unbounded, non-deterministic window.

Every post-acquisition operation, including the test callback, now runs inside the `try`/`finally` that disposes `lockStream` — acquisition and the guarded region are adjacent, with nothing but the callback invocation between them. An exception there still releases the lock immediately as it unwinds.

New deterministic test `Execute_CallbackThrowsAfterLockAcquired_LockStillReleasedDeterministically`: seeds a throwing callback, asserts the first call propagates the exception with queue-state/runs.jsonl left byte-unchanged (no partial write, no orphaned event), then asserts a second, fully independent call immediately acquires the same lock and completes normally.

Full solution suite green (3687 passed, 1 pre-existing skip, up from 3686), `git diff --check` clean, ja/en developer-reference docs updated with the round-7 subsection.

Ready for another review pass.

PR #1199, opened 2026-07-27
G548: Guard every queue-state write with a no-item-loss invariant and stale-base re-application

### LINKED ISSUE #1198: G548: Queue-state writers must never silently drop other items — no-item-loss invariant and stale-base detection
# G548: Queue-state writers must never silently drop other items — no-item-loss invariant and stale-base detection

## Goal

Protect the shared `queue-state.json` from concurrent-writer lost updates: every canonical mutation (1) verifies before persisting that it removes **no queue item it was not explicitly asked to remove**, aborting loud with the exact execution units that would be lost, (2) detects a **stale base** (on-disk file changed since read) and re-applies its scoped mutation on the fresh state instead of persisting the stale copy, and (3) keeps mutations item-scoped on re-application. A regression fixture reproduces the field incident where a sekiban write silently erased an intent-cli item.

## Why This Slice Exists Now

Field incident, 2026-07-23 (host commit `2ab082cf`): a sekiban-domain canonical write recorded a G841 PR linkage from a stale in-memory base and **silently dropped the intent-cli G545 queue item** seeded an hour earlier. Nothing errored; the commit message claimed only the linkage change. The loss stayed invisible for four days, then surfaced as `closeout-plan host-metadata-blocked` and combined with the `pr-is-draft` recovery gate into a circular deadlock; restoration took three canonical surfaces plus an operator (host commit `c0897649`).

`queue-state.json` is one shared file for **all domains** on a multi-domain host, written concurrently by several loops from different checkouts — read-modify-write races are structural. G536 hardened publish-flow's own rerun; this incident came through a *different* writer path (another session, another domain), so the protection belongs at the shared writer layer.

## Current Observed State

- Canonical writers deserialize the whole file, mutate in memory, and reserialize the whole file; whatever the in-memory copy lacks is silently erased.
- No writer compares against the on-disk state at persist time; no invariant checks item survival.
- The v0.5.0 serializer normalization (priority_revision backfill) demonstrated whole-file rewrites are routine, magnifying the blast radius of a stale base.

- Explicit removals stay legitimate: retire (G525), completed-item lifecycle, and any operation whose contract names the item may remove/rewrite it. The invariant targets **unrequested** loss only.
- git-level conflict handling is out of scope: the incident happened within a fast-forward-clean history; the defense must work at the writer layer before commit.
- The restoration path (queue-seed → publish-flow idempotent rerun → write-recovered-linkage) exists and works (proven 2026-07-27); the fail-loud message should name it.
- Per-domain queue files are a possible future design; explicitly out of scope here (recorded as escalation criterion).

## Target Repo / Path / Part

Target part: The shared queue-state persistence layer used by all canonical mutations, plus developer-reference docs.

## In Scope

- **No-item-loss invariant** at persist time: diff the on-disk item set against the outgoing set; abort loud naming lost execution units unless the operation explicitly requested each removal.
- **Stale-base detection**: compare `updated_at`/content hash captured at read; on mismatch, re-read and re-apply the scoped mutation to the fresh state (recording the re-application in output) instead of persisting stale memory.
- **Item-scoped re-application**: a mutation for execution unit X touches item X (plus `updated_at`) only.
- Regression fixture reproducing the 2ab082cf shape (two writers, same base; both changes survive).
- Fail-loud message naming the canonical recovery surfaces.
- ja/en docs: the invariant, re-application, and multi-writer expectations for shared hosts.

## Out Of Scope

- Per-domain queue file split (future design; escalation criterion recorded in closeout learning).
- git merge/rebase strategies, file locking daemons, or cross-process mutexes.
- Runs-log protection (G542 covers runs auditing) and repair-staleness detection (G546).

The child PR hardens the shared queue-state persistence layer so every canonical mutation aborts loudly — naming the exact execution units — if persisting would remove any queue item the operation was not explicitly asked to remove or retire; detects at persist time that the on-disk file changed since read (updated_at/content-hash comparison) and re-reads then re-applies its item-scoped mutation to the fresh state rather than persisting the stale in-memory copy, recording the re-application in its output; and keeps re-applied mutations scoped to the named execution unit plus updated_at. A regression fixture reproduces the 2026-07-23 two-writer incident (an unrelated linkage write on a stale base) and proves both changes survive; the abort message names queue-seed-from-packet, the publish-flow idempotent rerun, and closeout-plan --write-recovered-linkage as the recovery path; ja/en developer reference documents the invariant and multi-writer expectations. No per-domain file split, no locking daemons, no runs-log changes.

## Acceptance Criteria

- Persisting a state that would drop an unrequested item aborts loud, naming the execution units and the recovery surfaces (fixture).
- Stale-base fixture: writer A seeds an item, writer B (stale base) records an unrelated linkage → B re-applies on fresh state, both changes present, re-application recorded in B's output.
- Explicit removals (retire, completed lifecycle) still work unchanged (fixtures).
- Re-applied mutations touch only the named item + `updated_at` (byte-level assertion on unrelated items).
- ja/en docs present; focused tests green.

## Verification

- Run the focused queue-persistence suites including the two-writer regression; show green.
- Manual replay of the 2ab082cf shape on a fixture host.
- `git diff --check` clean.

## Related Links

- Field incident: host commits 2ab082cf (loss, 2026-07-23) and c0897649 (restoration, 2026-07-27); orchestrator STOP report 2026-07-27 09:08Z (circular deadlock).
- G536 — publish-flow rerun truthfulness (same family, different writer path); G542 — runs-log auditing sibling.

## Knowledge Maintenance

- Intent placement: `intents/intent-cli/intent-tree/means/03-state-and-audit-strategy.md` (existing node).
- ADR candidate: none — invariant enforcement at an existing layer; the per-domain-split decision, if ever taken, gets its own ADR.
- Diagram candidate: none.
- Closeout writeback expected: no — recurrence after landing escalates the per-domain-file design question.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1198

## What

`queue-state.json` is **one file shared by every domain** on a multi-domain host, written concurrently by several loops from different checkouts. Every canonical writer deserialized the whole file, mutated in memory, and reserialized the whole file — so a read-modify-write race did not merely conflict, it **silently erased** whatever the stale in-memory copy lacked.

This adds `QueueStatePersistence`, the single shared layer **all 19 canonical queue-state mutations now write through**.

## The incident being closed

2026-07-23 (host commit `2ab082cf`): a sekiban-domain write recorded a G841 PR linkage from a base read an hour earlier and dropped the intent-cli G545 item seeded in between. Nothing errored; the commit message claimed only the linkage change. The loss stayed invisible for four days, then surfaced as `closeout-plan host-metadata-blocked` and combined with the `pr-is-draft` recovery gate into a circular deadlock. Restoration took three canonical surfaces plus an operator (host commit `c0897649`).

## Three guarantees, one layer

**1. Stale-base detection and re-application.** The state the caller *read* is compared against what is on disk *now*, at persist time — through the same serializer round-trip, so pure formatting drift is never mistaken for a concurrent write. On a mismatch the caller's mutation is re-applied to the **fresh** state and the re-application is reported back so it is never invisible.

The mutation is **derived**, not supplied: `QueueStateItemDelta.Between(base, outgoing)` computes the item-level delta. That is why 19 existing writers gained stale-base recovery without any of them restructuring its own mutation logic — each one only had to hand the guard the base state it had already read.

**2. No-item-loss invariant.** Any execution unit present on disk but missing from the outgoing state, and not named as an expected removal, **aborts the write** — naming the exact units and the canonical recovery path — and leaves the file untouched. An on-disk state that cannot be *read* aborts too, since neither guarantee can be established against it.

**3. Item-scoped re-application.** A re-applied mutation touches only the units its delta covers, plus `updated_at`. Unrelated items are carried through byte-identically from the fresh state, **in the fresh state's order**, so a stale copy's older view of another item can never overwrite a newer one. Asserted rather than assumed — the failure it guards against is precisely the silent one.

## Explicit removals stay legitimate

Retire (G525), the completed-item lifecycle, and any operation whose contract names the item it may remove pass those units as `expectedRemovals`. The invariant targets **unrequested** loss only, and an allow-list entry excuses that unit and nothing else. Retire itself needs no entry at all — it rewrites the item as `state=retired` rather than removing it, which a fixture pins.

## Fixtures (15)

| fixture | proves |
| --- | --- |
| two-writer 2ab082cf regression | A seeds G545, B writes a linkage from the stale base → **both** survive, `reapplied` reported |
| item-scope, byte-level | unrelated items byte-identical to the fresh state; the stale copy's older view does not overwrite a newer one |
| fresh-order preservation | concurrently-added items keep their on-disk order |
| unrequested loss | aborts naming `G545, G546` + all three recovery surfaces; file byte-identical |
| stale base that *would* have dropped an item | repaired by re-application, **not** aborted — the abort is the last line of defense, not the first |
| unreadable on-disk state | aborts rather than overwriting |
| explicit removal, clean base | allowed |
| explicit removal, stale base | removes only that unit; concurrent seed survives |
| allow-list does not over-excuse | still aborts on the *other* lost unit |
| retire-style rewrite | never treated as loss, needs no allow-list |
| clean base / missing file / concurrent add | ordinary writes unaffected |
| formatting-only difference | not mistaken for a concurrent write |
| delta derivation | captures only the units actually changed |

## Verification

- Focused `QueueStatePersistenceTests`: **15/15 pass**.
- Full solution suite: **4115 passed, 1 skipped, 0 failed** across 11 assemblies — every existing writer's own suite still green, which is the regression bar for a 19-call-site change.
- Manual replay on a multi-domain fixture host: a real `intent-cli queue transition SKS-G841 completed` through the built binary left the intent-cli `G545` item untouched. (The *interleaved* two-writer case is covered by the deterministic fixture rather than a manual run — a real race is not reproducible on demand.)
- `git diff --check` clean.
- ja/en `09-developer-reference.md` document the invariant, re-application, and multi-writer expectations.

## Multi-writer expectations

Concurrent canonical writers are supported and expected: a losing writer is **repaired**, not rejected, so no loop has to serialize against another. What is not supported is a writer that bypasses the guard — hand-editing the file, or a new command calling `File.WriteAllText` directly. There are now **zero** direct `File.WriteAllText(..., QueueStateSerializer.Serialize(...))` call sites outside the guard.

Out of scope and unchanged: per-domain queue file split (future design; a recurrence after this lands is the escalation criterion), file-locking daemons, cross-process mutexes, git merge strategy, and runs-log handling.

### DIFF
--- a/docs/en/09-developer-reference.md
+++ b/docs/en/09-developer-reference.md
@@ -1809,8 +1809,12 @@ days, then surfaced as `closeout-plan host-metadata-blocked` and combined
 with the `pr-is-draft` recovery gate into a circular deadlock. Restoration
 took three canonical surfaces plus an operator (host commit `c0897649`).
 
-Every canonical mutation now writes through one shared guard
-(`QueueStatePersistence.Persist`), which enforces three things:
+Every canonical mutation now writes through one shared guard,
+`QueueStatePersistence`. It lives in **`IntentSystem.Supervisor`** — the
+assembly that owns the queue-state model and serializer, and the only one
+*both* `IntentSystem.Cli` and `IntentSystem.Drift` reference — so "every
+canonical writer" means every writer in the solution, including the drift
+service's corrective enqueue. It enforces three things:
 
 1. **Stale-base detection and re-application.** The state the caller *read*
    is compared against what is on disk *now*, at persist time (through the
@@ -1840,11 +1844,34 @@ only; an allow-list entry excuses that unit and nothing else. Retire itself
 needs no entry — it rewrites the item as `state=retired` rather than removing
 it.
 
+**Re-application is reported, never silent.** A canonical command whose
+write was re-applied says so in its own output, naming the execution units it
+was re-applied for — a writer that quietly repaired itself against a
+concurrent write teaches an operator nothing about the contention that caused
+it. `queue transition`, for example, prints `note: queue-state changed after
+it was read (a concurrent canonical write); this transition was re-applied to
+the current state for <units> and no other item was modified.`
+
+**The one raw-text writer.** `metadata update` is the bounded controlled
+metadata writer: it mutates queue-state as raw JSON so it never rewrites a
+field it does not own, and it accepts documents that need not satisfy the
+full `QueueItem` contract. It uses `PersistRawJson`, which checks the
+invariant on `items[].execution_unit` read straight out of the JSON — never
+by deserializing — and on a clean base writes the caller's own text verbatim.
+Only if a concurrent write is actually detected does it need the model to
+re-apply; when the document cannot round-trip through the model it **aborts
+loud** ("re-run this command against the current state") rather than
+normalizing a file it exists not to normalize, or overwriting a change it
+cannot see. Nothing is written either way.
+
 **Multi-writer expectations for shared hosts.** Concurrent canonical writers
 are supported and expected: a losing writer is repaired (re-applied), not
 rejected, so no loop has to serialize against another. What is *not*
 supported is a writer that bypasses the guard — hand-editing
-`queue-state.json`, or a new command calling `File.WriteAllText` directly.
+`queue-state.json`, or a new command calling `File.WriteAllText` directly — a
+source-level fixture (`QueueStateWriterCoverageTests`) fails with the file and
+line of any writer added anywhere in `src/` that bypasses the guard, so the
+all-writers claim cannot regress by hand-verification again.
 Deliberately out of scope, and unchanged by this slice: a per-domain queue
 file split (a future design decision; a recurrence after this lands is the
 escalation criterion), file-locking daemons, cross-process mutexes, and
--- a/docs/en/09-developer-reference.md
+++ b/docs/en/09-developer-reference.md
@@ -1792,6 +1792,66 @@ deliberately left on the legacy whole-file behavior (out of scope for this
 slice); `RunLogSerializer` / the RunEvent required-field contract are
 unchanged.
 
+### Queue-state writes are guarded: no-item-loss invariant and stale-base re-application (G548)
+
+`queue-state.json` is **one file shared by every domain** on a multi-domain
+host, written concurrently by several loops from different checkouts. Every
+canonical writer deserializes the whole file, mutates in memory, and
+reserializes the whole file — so a read-modify-write race does not merely
+conflict, it **silently erases** whatever the stale in-memory copy happened
+not to contain.
+
+**Field incident, 2026-07-23** (host commit `2ab082cf`): a sekiban-domain
+write recorded a G841 PR linkage from a base read an hour earlier and dropped
+the intent-cli G545 queue item seeded in between. Nothing errored; the commit
+message claimed only the linkage change. The loss stayed invisible for four
+days, then surfaced as `closeout-plan host-metadata-blocked` and combined
+with the `pr-is-draft` recovery gate into a circular deadlock. Restoration
+took three canonical surfaces plus an operator (host commit `c0897649`).
+
+Every canonical mutation now writes through one shared guard
+(`QueueStatePersistence.Persist`), which enforces three things:
+
+1. **Stale-base detection and re-application.** The state the caller *read*
+   is compared against what is on disk *now*, at persist time (through the
+   same serializer round-trip, so pure formatting drift is never mistaken for
+   a concurrent write). On a mismatch the caller's mutation — derived
+   automatically as an item-level delta between its base and its outgoing
+   state — is re-applied to the **fresh** state instead of persisting the
+   stale copy, and the re-application is reported back so it is never
+   invisible.
+2. **No-item-loss invariant.** Any execution unit present on disk but missing
+   from the outgoing state, and not named as an expected removal, **aborts
+   the write** — naming the exact units and the canonical recovery path
+   (`queue-seed-from-packet` → idempotent `issue publish-flow` rerun →
+   `closeout-plan --write-recovered-linkage`). The file is left untouched.
+   An on-disk state that cannot be *read* also aborts, since neither
+   guarantee can be established against it.
+3. **Item-scoped re-application.** A re-applied mutation touches only the
+   units its delta actually covers, plus `updated_at`. Every unrelated item
+   is carried through byte-identically from the fresh state, in the fresh
+   state's order — so a stale copy's older view of another item can never
+   overwrite a newer one.
+
+**Explicit removals stay legitimate.** Retire (G525), the completed-item
+lifecycle, and any operation whose contract names the item it may remove pass
+those units as expected removals. The invariant targets **unrequested** loss
+only; an allow-list entry excuses that unit and nothing else. Retire itself
+needs no entry — it rewrites the item as `state=retired` rather than removing
+it.
+
+**Multi-writer expectations for shared hosts.** Concurrent canonical writers
+are supported and expected: a losing writer is repaired (re-applied), not
+rejected, so no loop has to serialize against another. What is *not*
+supported is a writer that bypasses the guard — hand-editing
+`queue-state.json`, or a new command calling `File.WriteAllText` directly.
+Deliberately out of scope, and unchanged by this slice: a per-domain queue
+file split (a future design decision; a recurrence after this lands is the
+escalation criterion), file-locking daemons, cross-process mutexes, and
+git-level merge strategy — the 2ab082cf loss happened inside a
+fast-forward-clean history, so the defense has to sit at the writer, before
+anything reaches a commit.
+
 ---
 
 ## Version flow
--- a/docs/en/09-developer-reference.md
+++ b/docs/en/09-developer-reference.md
@@ -1858,11 +1858,16 @@ field it does not own, and it accepts documents that need not satisfy the
 full `QueueItem` contract. It uses `PersistRawJson`, which checks the
 invariant on `items[].execution_unit` read straight out of the JSON — never
 by deserializing — and on a clean base writes the caller's own text verbatim.
-Only if a concurrent write is actually detected does it need the model to
-re-apply; when the document cannot round-trip through the model it **aborts
-loud** ("re-run this command against the current state") rather than
-normalizing a file it exists not to normalize, or overwriting a change it
-cannot see. Nothing is written either way.
+When a concurrent write *is* detected, the re-application happens at the
+**JSON level** too — items this writer did not touch are carried across as
+the fresh document's own nodes, so fields the model does not know about
+survive on both sides and a stale copy's older view of another item can never
+overwrite a newer one.
+
+Because `metadata update` is the bounded **linkage** writer — the role writer
+B played in the 2ab082cf incident — a re-application on this path is reported
+in its own result: `queue_state_reapplied` / `queue_state_reapplied_execution_units`
+in JSON, and a `queue_state_reapplied:` block in the text output.
 
 **Multi-writer expectations for shared hosts.** Concurrent canonical writers
 are supported and expected: a losing writer is repaired (re-applied), not
--- a/docs/ja/09-developer-reference.md
+++ b/docs/ja/09-developer-reference.md
@@ -1996,12 +1996,16 @@ writer です: 所有していない field を書き換えないよう queue-sta
 として変更し、完全な `QueueItem` 契約を満たさない document も受け付けます。
 この writer は `PersistRawJson` を使い、invariant を JSON から直接読んだ
 `items[].execution_unit` で検査し(deserialize は一切しません)、base が clean
-なら caller 自身のテキストをそのまま書き込みます。model が必要になるのは並行
-書き込みを実際に検出して再適用する場合だけで、document が model を round-trip
-できないときは、normalize しないために存在する writer が normalize することも、
-見えていない変更を上書きすることもせず、**loud に中止**します
-(「current state に対して再実行してください」)。いずれの場合もファイルへの
-書き込みは行われません。
+なら caller 自身のテキストをそのまま書き込みます。並行書き込みを実際に検出した場合の再適用も **JSON レベル**で行われます —
+この writer が触れていない item は fresh document 自身の node としてそのまま
+引き継がれるため、model が知らない field も両側で保持され、stale copy が持つ
+他 item の古い姿が新しいものを上書きすることはありません。
+
+`metadata update` は bounded **linkage** writer — 2ab082cf incident における
+writer B の役割そのもの — であるため、この経路での再適用は自身の result に
+記録されます: JSON では `queue_state_reapplied` /
+`queue_state_reapplied_execution_units`、text 出力では `queue_state_reapplied:`
+ブロックです。
 
 **共有 host における multi-writer の期待。** canonical writer の並行実行は
 サポートされ、想定されています: 競争に負けた writer は拒否されるのではなく修復
--- a/docs/ja/09-developer-reference.md
+++ b/docs/ja/09-developer-reference.md
@@ -1951,7 +1951,12 @@ commit message は linkage 変更のみを主張していました。この loss
 (host commit `c0897649`)。
 
 現在は、すべての canonical mutation が 1 つの共有 guard
-(`QueueStatePersistence.Persist`)経由で書き込み、次の 3 点を強制します:
+`QueueStatePersistence` 経由で書き込みます。この guard は
+**`IntentSystem.Supervisor`** — queue-state の model と serializer を所有し、
+`IntentSystem.Cli` と `IntentSystem.Drift` の*両方*が参照する唯一の assembly —
+に置かれているため、「すべての canonical writer」は CLI 内だけでなく solution
+全体の writer(drift service の corrective enqueue を含む)を意味します。
+強制する内容は次の 3 点です:
 
 1. **stale-base の検出と再適用。** caller が*読んだ*状態と、persist 時点で
    ディスク上に*実際にある*状態を比較します(同一の serializer round-trip を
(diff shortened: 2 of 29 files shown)


tomohisa:
Semantic review of `0402fba62aa807e37d690195497b4670d6e0a1a6`: request-update

Two blocking contract gaps remain:

1. The no-item-loss guard does not cover every canonical queue-state writer. `src/IntentSystem.Cli/Commands/MetadataUpdateCommand.cs:321` still persists `newQueueText` directly, and that command is explicitly documented as the bounded controlled metadata writer. `src/IntentSystem.Drift/IntentDriftService.cs:93` also directly serializes and writes the shared queue state when it enqueues a corrective item. Either path can still reproduce the stale whole-file overwrite that G548 exists to prevent. Please route both through the shared invariant or place the invariant at a lower shared layer that both assemblies can use, and add a source/call-path fixture so the all-writers claim cannot regress.

2. Stale-base re-application is not recorded in the invoking writer output. Every production call to `QueueStatePersistence.Persist(...)` discards `QueueStatePersistResult`; only the persistence unit test observes `ReappliedOnFreshBase` / `ReappliedExecutionUnits`. The issue contract requires writer B to record re-application in its output. Please propagate the result through canonical command output for a stale-base writer and add a command-level fixture that proves the observable output names the re-applied execution unit.

The base branch policy is correct (`main`, `direct-main`), CI is green, and the core sequential stale-base/item-scope fixtures are otherwise aligned with the intent. These two gaps block approval because they leave canonical loss paths unguarded and the required recovery signal invisible.

tomohisa:
## Repair round 2 — shared-assembly guard, full writer coverage, observable re-application

Fixed head: `9bbacee3d97bdd0b653c1453c586b52108864a90` (CI: **pass**, 1m43s — [run 30316261283](https://github.com/J-Tech-Japan/intent-system/actions/runs/30316261283/job/90142336149), merge state CLEAN).

Addresses both blocking findings in https://github.com/J-Tech-Japan/intent-system/pull/1199#issuecomment-5098173083.

### 1. Every canonical writer, including the two that were missed

The `IntentDriftService` gap was **structural, not accidental**: it lives in `IntentSystem.Drift`, which cannot reference `IntentSystem.Cli` — where the guard was. No amount of call-site editing could have closed it. So the guard moved down a layer, exactly as the review suggested:

`QueueStatePersistence` now lives in **`IntentSystem.Supervisor`** — the assembly that owns the queue-state model and serializer, and the only one *both* writer assemblies reference — and is public.

- **`IntentDriftService.cs:93`** — the corrective enqueue writes the same shared multi-domain file every CLI writer does, and now goes through `Persist(base, outgoing)` like the rest.
- **`MetadataUpdateCommand.cs:321`** — the bounded controlled metadata writer needed different treatment, and finding out why was the substantive part of this round. It mutates raw JSON so it never rewrites a field it does not own, **and it accepts queue documents that do not satisfy the full `QueueItem` contract**. Routing it through the model-based guard broke 9 existing `MetadataUpdateCommandTests` with `JsonException: missing required properties including 'dependencies', 'blocked_by', 'clarification_return_path', 'packet_paths'` — the guard would have rejected the very files this writer exists to handle.

  It therefore gets `PersistRawJson`, which enforces the invariant on `items[].execution_unit` read **straight out of the JSON, never through the model**, and on a clean base writes the caller's own text **verbatim** — preserving the bounded-writer property completely. Only a *detected* concurrent write needs the model in order to re-apply the delta; when the document cannot round-trip through it, the writer **aborts loud** (`the file changed after it was read … re-run this command against the current state`) rather than normalizing a file it exists not to normalize, or overwriting a change it cannot see. Nothing is written on either branch.

**The all-writers claim is now a fixture, not hand-verification** — `QueueStateWriterCoverageTests`:

| test | pins |
| --- | --- |
| `EveryQueueStateWriteInSource_GoesThroughTheSharedGuard` | scans every `.cs` under `src/`, failing with **file and line** of any queue-state write bypassing the guard (matches both an explicit `QueueStateSerializer.Serialize` write and a `File.WriteAllText` whose target names a queue path — the two shapes of the round-1 gap) |
| `TheGuardItselfIsInASharedAssemblyBothWriterAssembliesCanReach` | the guard stays in `IntentSystem.Supervisor` and that project stays `ProjectReference`-free, so it cannot drift back up a layer and recreate the structural gap |
| `KnownQueueStateWriters_AllReferenceTheGuard` | names both round-1 gaps explicitly, so a rename or refactor cannot quietly drop either |

### 2. Re-application is now recorded in the writer's own output

Every production call discarded `QueueStatePersistResult`. `queue transition` now surfaces it:

> `note: queue-state changed after it was read (a concurrent canonical write); this transition was re-applied to the current state for SKS-G841 and no other item was modified.`

Proving that at the **command** level needs an interleaving a single-process test cannot otherwise produce, so the guard gained a `BeforePersistHook` test seam that fires in exactly the window a concurrent canonical write occupies: after the command has read queue-state, before the guard reads it again. The fixture uses it to seed another domain's item in that window, then asserts the command's observable output names the re-applied unit — and that both items survive with the right states. A companion fixture asserts a clean-base run says nothing about re-application.

### Fixtures (10 new, 25 across the two suites)

- Command-level: stale-base output names the re-applied unit; clean-base output stays quiet.
- Raw JSON: clean-base verbatim write; unrequested-loss abort; stale-base re-application; a **partial document** checked on identity alone (the case that would have been rejected by the model); the partial-document stale abort leaving the file untouched.
- Source/call-path coverage: the three tests above.

### Verification

- Focused `QueueStatePersistenceTests` + `QueueStateWriterCoverageTests`: **25/25 pass**.
- Full solution suite: **4123 passed, 1 skipped, 0 failed** across 11 assemblies — including the 9 `MetadataUpdateCommandTests` that the model-based routing broke and the raw-text guard restores.
- `git diff --check` clean.
- ja/en docs updated for the shared-assembly placement, the reported re-application, and the raw-text writer's semantics.

tomohisa:
Semantic re-review of `9bbacee3d97bdd0b653c1453c586b52108864a90`: request-update (round 3)

The writer-coverage finding is resolved: the guard now lives in `IntentSystem.Supervisor`, both previously missed writers route through it, raw JSON preserves the bounded metadata writer semantics, and the source/call-path fixtures pin the coverage.

One blocking acceptance gap remains. The packet and review context require the 2ab082cf regression shape itself: writer B starts stale, records an unrelated **linkage**, both changes survive, and **B output records the re-application**. At this head, only `QueueTransitionCommand` consumes `QueueStatePersistResult`; every other production call still discards it. In particular, `MetadataUpdateCommand` — the bounded writer that adds `linked_pr` / completed closeout — calls `PersistRawJson(...)` without retaining its result, so a successful stale re-application on the incident-shaped linkage path is still silent. The new command-level fixture exercises a state transition, not the required linkage writer, and therefore does not close this packet clause.

Please retain the `PersistRawJson` result in `MetadataUpdateCommand`, expose the re-application and exact execution units in that command output (text and structured JSON as applicable), and add a command-level stale-base fixture where the metadata/linkage writer is B and a concurrently seeded unrelated item survives. The queue-transition notice may remain as useful additional coverage, but it cannot substitute for the explicitly required linkage regression.

Pinned evidence: CI `Build and test (source contract)` is green for this head, base policy is correct (`main` / `direct-main`), and `git diff --check` is clean. This is an implementation-finding tied directly to packet AC #2 and the review-context two-writer delta.

tomohisa:
## Repair round 3 — the linkage writer records its own re-application

Fixed head: `195efc04162d31b309f87a04cebd82521df9cdf7` (CI: **pass**, 1m48s — [run 30316831652](https://github.com/J-Tech-Japan/intent-system/actions/runs/30316831652/job/90144048002), merge state CLEAN).

Addresses the remaining acceptance gap in https://github.com/J-Tech-Japan/intent-system/pull/1199#issuecomment-5098281839: the packet requires the 2ab082cf shape with the **linkage** writer as B, and B's output recording the re-application.

### `metadata update` now retains and reports its result

`MetadataUpdateResult` gains `queue_state_reapplied` and `queue_state_reapplied_execution_units` (with the camelCase aliases this command's contract uses), and the text output gains a `queue_state_reapplied:` block naming the units plus the explanatory note. `queue transition`'s notice stays as additional coverage.

### Making that path actually re-apply required fixing a real gap the requirement exposed

Wiring the result through would have been cosmetic on its own: `PersistRawJson` re-applied a stale-base change **through the model**, and aborted when the document could not round-trip. The host-shaped documents this bounded writer exists to accept routinely do not satisfy the full `QueueItem` contract — that is why it works on raw text at all — so **the incident's own file shape would have taken the abort path rather than recovering**. The round-2 fixture I wrote pinned that abort as if it were the intended behaviour; it was really the limitation showing through.

Re-application for the raw writer now happens **at the JSON level**, with no model involvement anywhere in this writer:

- the delta is derived by comparing `items[]` by `execution_unit` between base and outgoing;
- it is applied onto the **fresh document's own nodes**, so items this writer did not touch are carried across verbatim, in the fresh document's order;
- fields the model does not know about survive on both sides — B's own item keeps its host-specific fields, and A's concurrently-seeded item keeps its own;
- `updated_at` is taken from the writer's outgoing document; every other top-level field stays as the fresh document has it.

The round-2 abort fixture is replaced by one asserting this recovery on a partial document carrying host fields the model has never heard of.

### The required regression, at the command level, with the right writer

`Execute_StaleBase_LinkageWriteIsReappliedAndRecordedInOutput_ReproducesIncident2ab082cf_G548`:

| assertion | covers |
| --- | --- |
| `queue_state_reapplied: true`, `queue_state_reapplied_execution_units: ["SKS-G841"]` | **B's output records the re-application** and names the exact unit |
| `G545` present in the persisted file | the concurrently seeded unrelated item **survives** |
| `SKS-G841` is `completed` with `linked_pr.number: 841` | **B's own linkage change survives** |
| `G545` still `queued`, with **no** `linked_pr` | B touched nothing but its own item |

The interleaving uses the guard's `BeforePersistHook`, which fires in exactly the window a concurrent canonical write occupies: after `metadata update` has read queue-state, before the guard reads it again. A clean-base counterpart asserts no re-application is reported.

### Verification

- Focused: `QueueStatePersistenceTests` **22/22**, `MetadataUpdateCommandTests` **21/21** (including the 9 that the model-based routing broke in round 2).
- Full solution suite: **4125 passed, 1 skipped, 0 failed** across 11 assemblies.
- `git diff --check` clean.
- ja/en docs updated for the JSON-level re-application and the reported linkage recovery.

tomohisa:
Semantic re-review of `195efc04162d31b309f87a04cebd82521df9cdf7`: request-update (round 4)

The round-3 linkage/output acceptance gap is resolved in source: `MetadataUpdateCommand` now retains `PersistRawJson` results, emits reapplied units in JSON/text output, and the incident-shaped linkage fixture proves both the linkage and concurrent G545 seed survive.

One blocking deterministic-test finding remains. The two new command-level interleaving fixtures share the process-global static `QueueStatePersistence.BeforePersistHook`, but they live in different xUnit classes and run in parallel. Each test overwrites or clears the other test hook, so the required focused suites are not green when executed together.

Reproduction at the pinned head:

`dotnet test tests/IntentSystem.Cli.Tests/IntentSystem.Cli.Tests.csproj --no-build --filter 'FullyQualifiedName~QueueStatePersistenceTests|FullyQualifiedName~QueueStateWriterCoverageTests|FullyQualifiedName~MetadataUpdateCommandTests' --logger 'console;verbosity=minimal'`

Observed twice:
- run 1: 45 passed / 1 failed — `QueueTransitionCommand_OnAStaleBase_ReportsTheReapplicationAndNamesTheUnit_G548` did not observe the reapply notice;
- run 2: 44 passed / 2 failed — that test plus `Execute_StaleBase_LinkageWriteIsReappliedAndRecordedInOutput_ReproducesIncident2ab082cf_G548` (`QueueStateReapplied` false).

Control: running `QueueStatePersistenceTests` alone passes 22/22. This isolates the failure to cross-class hook contention, not the production reapply logic.

Please make the test seam isolation-safe (for example, place every class using this static hook in one non-parallel xUnit collection, or replace the global seam with a scoped/path-specific mechanism), then run the combined focused command in one invocation and report the total. This is an implementation-finding tied to the packet Verification requirement that the focused persistence suites be green.

tomohisa:
## Repair round 4 — the before-persist test seam is isolation-safe

Fixed head: `10c5c83d3afab50dd16154b1e9e6cdc8bc1252b7` (CI: **pass**, 1m41s — [run 30317293995](https://github.com/J-Tech-Japan/intent-system/actions/runs/30317293995/job/90145499136), merge state CLEAN).

The finding was correct and the diagnosis exact: `BeforePersistHook` was a single process-global mutable static, and the two command-level interleaving fixtures live in different xUnit classes, which run in parallel. Each cleared or overwrote the other's hook — hence 22/22 alone, 45/1 then 44/2 together.

### The seam is now keyed by queue-state path

`RegisterBeforePersistHook(path, hook)` returns a disposable registration, and the guard invokes only a hook registered for the exact path it is about to write. Each fixture owns its own temp workspace, so two fixtures can never address the same key — the contention is removed **structurally**.

I chose this over the non-parallel-collection option the review offered because collection serialization fixes the symptom by coupling every future fixture in this area to every other one; path scoping fixes the cause and leaves the suites parallel.

Registering a second hook for the same path now **throws** rather than silently replacing the first — that silent replacement is precisely what broke these fixtures, so it should not be available.

### Regression tests for the fix itself

| test | pins |
| --- | --- |
| `BeforePersistHook_IsScopedByPath_SoParallelFixturesCannotContend_G548` | two hooks on different paths fire independently; only the one for the written path fires |
| `BeforePersistHook_RefusesASecondRegistrationForTheSamePath_G548` | duplicate registration is an error, not a silent replace |
| `BeforePersistHook_IsRemovedOnDispose_G548` | disposal removes the registration |

### The reviewer's combined command, one invocation

```
dotnet test tests/IntentSystem.Cli.Tests/IntentSystem.Cli.Tests.csproj --no-build \
  --filter 'FullyQualifiedName~QueueStatePersistenceTests|FullyQualifiedName~QueueStateWriterCoverageTests|FullyQualifiedName~MetadataUpdateCommandTests' \
  --logger 'console;verbosity=minimal'
```

**Passed! — Failed: 0, Passed: 49, Skipped: 0, Total: 49.**

Run five times total across the fix (three at 46 before the three new regression tests were added, twice at 49 after); green every time, with no ordering dependence.

### Verification

- Combined focused command: **49/49**, repeatedly.
- Full solution suite: **4128 passed, 1 skipped, 0 failed** across 11 assemblies.
- `git diff --check` clean.
- No production behaviour change: the only touched production path is how the guard looks a hook up.

tomohisa:
Approved semantic review for G548 at pinned head `10c5c83d3afab50dd16154b1e9e6cdc8bc1252b7`.

Packet and intent evidence:
- Checked G548 packet scope/pre-post contract: every canonical shared queue-state mutation must prevent unrequested item loss, stale writers must re-apply only their scoped mutation and report it, and explicit lifecycle rewrites/removals must remain valid.
- Consistent with `intents/intent-cli/intent-tree/means/03-state-and-audit-strategy.md`: queue decisions and repair/closeout provenance remain durable and explainable.

Acceptance mapping:
1. All canonical writers route through the shared `IntentSystem.Supervisor.QueueStatePersistence`; coverage fixtures include the previously missed metadata and drift writers.
2. The 2ab082cf incident shape is covered at command level: stale `MetadataUpdateCommand` writer B records the linkage, concurrent G545 survives, and structured/text output reports `queue_state_reapplied` plus `SKS-G841`.
3. Model and raw-JSON re-application are item-scoped; unrelated fresh items, unknown host fields, and ordering survive, with only touched units plus `updated_at` changed.
4. Unrequested loss aborts with exact units and all three recovery surfaces; retire and explicit-removal fixtures remain green.
5. EN/JA developer references document the invariant and shared-host behavior.
6. The round-4 path-scoped disposable hook removes cross-class test contention; same-path duplicate registration fails loud.

Verification:
- Reviewer-run combined focused command: 49 passed / 0 failed / 0 skipped.
- GitHub Actions `Build and test (source contract)`: SUCCESS for this exact head (run 30317293995).
- Implementer full suite: 4128 passed / 1 skipped / 0 failed across 11 assemblies.
- `git diff --check origin/main..10c5c83d`: clean.
- Base policy: `main` matches `direct-main`; merge state CLEAN.

Tests are necessary but not sufficient; approval also rests on the packet/intent mapping above and direct source inspection through all four repair rounds. Out-of-scope boundaries were preserved: no per-domain split, lock daemon/mutex, git merge strategy, or runs-log protection expansion. `Closes #1198` matches the linked source issue.

PR #1252, opened 2026-08-02
G574: park blocked stalled-work units

### LINKED ISSUE #1251: G574: backlog-ready-idle excludes blocked units — stalled-work must never recommend publishing a parked packet
# G574: backlog-ready-idle excludes blocked units — stalled-work must never recommend publishing a parked packet

## Goal

Fix the G544 `backlog-ready-idle` detector so units whose queue entry is `state=blocked` with non-empty `blocked_by` are reported as an **informational `blocked-parked`** item (block reason + age-since-block, no publish recommendation) instead of `backlog-ready-idle`; half-converged entries (blocked_by/state disagreeing) become a distinct non-informational **state-drift** item pointing at the canonical block/unblock surfaces; after a canonical unblock (including the G561 pre-publish path) normal idle accounting resumes with age measured from the unblock. Genuinely queued units keep byte-identical behavior.

## Why This Slice Exists Now

Field incident (2026-08-01): with G570/G571/G573 canonically parked behind the `v0.7.1-release-gate` (the established priority-encoding pattern), `stalled-work` reported `backlog-ready-idle G570 age=45m` with the recommendation to run `publish-flow` — following it blindly would have published a deliberately parked unit and defeated the operator-approved sequencing. Both the host watchdog and the orchestrator had to install manual suppressions, which is exactly the failure smell (a detector whose output must be second-guessed is worse than no detector).

This is the third instance of the same class: G553 excluded converged-blocked units from WIP counting, G545 excluded them from `claimed-but-silent`; the G544 detector predates pre-publish blocking (first used in the G559 wake, canonicalized by G561) and never gained the exclusion.

## Current Observed State

- Reproduced on live state: `G570 state=blocked blocked_by=['v0.7.1-release-gate …']` while `stalled-work` returns `backlog-ready-idle G570` with a publish recommendation.
- Orchestrator and host watchdog are running documented manual suppressions until this merges.

- Blocked semantics (G553 convergence, G561 pre-publish block/unblock) are correct; only the detector lags.
- Informational items already flow through `stalled-work` and `heartbeat` (G546 lineage) — reuse that shape for `blocked-parked`.
- Sequenced on the v0.8.0 line after the release-gate lifts (it will be parked behind the same gate it fixes the reporting for — which is fine and self-demonstrating).

## Target Repo / Path / Part

Target paths: `src/IntentSystem.Cli/**, tests/**`

Target part: the stalled-work backlog-ready-idle detection path + heartbeat carriage + tests.

## In Scope

- Exclusion of converged-blocked units from `backlog-ready-idle`; informational `blocked-parked` item (reason, age-since-block, no publish recommendation); heartbeat carriage.
- Distinct non-informational state-drift item for half-converged entries with a canonical-surface recommendation (never publish).
- Post-unblock accounting: idle age from the unblock timestamp, fixture-proven via the G561 pre-publish path.
- Regression fixture reproducing the field shape; byte-unchanged behavior for genuinely queued units.

## Out Of Scope

- Selector/publish-flow changes; block/unblock semantics; guide content; other stalled-work kinds.

The child PR changes `automation stalled-work` (and its `automation heartbeat` carriage) so that a unit whose queue entry is `state=blocked` with non-empty `blocked_by` is never classified `backlog-ready-idle` and instead yields an informational `blocked-parked` item naming the block reason and age since blocking with no publish recommendation; yields a distinct non-informational state-drift item, whose recommended action points at the canonical block/unblock surfaces and never at publish-flow, when an entry is half-converged (blocked_by without `state=blocked` or `state=blocked` with empty blocked_by); resumes normal backlog-ready-idle accounting after a canonical unblock — including the G561 pre-publish path — with idle age measured from the unblock; and proves via a regression fixture reproducing the field incident (packet directory present, unit blocked pre-publish with a release-gate reason) that the detector reports `blocked-parked` and not `backlog-ready-idle`, while genuinely queued units keep byte-identical detection output. The full suite is green and the diff is bounded to the detection paths and tests. No selector, publish-flow, block/unblock-semantics, or guide changes.

## Acceptance Criteria

- Converged-blocked → informational blocked-parked (reason + age-since-block, no publish rec); heartbeat carries it.
- Half-converged → non-informational state-drift with canonical-surface recommendation.
- Post-unblock idle age from unblock; field-shape regression fixture green; queued units byte-unchanged.
- Full suite green.

## Verification

- Run the regression and detection fixtures plus the full suite; show green.
- `git diff --check` clean.

## Related Links

- Field incident: intent-cli-dev agmsg 2026-08-01 (backlog-ready-idle G570 under release-gate block; orchestrator + watchdog suppressions).
- Lineage: G544 (the detector), G553 (WIP blocked exclusion), G545 (claimed-but-silent exclusion), G561 (pre-publish block/unblock), G546 (informational kinds).
- intents/intent-cli/intent-tree/means/01-queue-topology.md

## Knowledge Maintenance

- Intent placement: node 03 primary, node 01 supporting.
- ADR candidate: none.
- Diagram candidate: none.
- Docs update: none.
- Closeout writeback expected: no — the next parked sequence validates (and the host watchdog's temporary filter is removed).

## Base Branch Policy

Policy: `direct-main`

tomohisa:
## Summary
- classify converged pre-publish blocked units as informational blocked-parked with the recorded reason and age since block
- classify both half-converged state/blocked_by shapes as actionable state-drift without recommending publish-flow
- carry the new classifications through heartbeat and restart idle age from the canonical pre-publish unblock

## Verification
- focused detector/heartbeat/block tests: 150 passed
- full suite: 4,565 total, 0 failed, 1 skipped
- hermetic restore/build/full suite: passed
- touched-file formatting and git diff --check: passed

Closes #1251

### DIFF
--- a/src/IntentSystem.Cli/Commands/AutomationStalledWorkCommand.cs
+++ b/src/IntentSystem.Cli/Commands/AutomationStalledWorkCommand.cs
@@ -206,6 +206,22 @@ internal static class AutomationStalledWorkCommand
     public const string KindBacklogReadyIdle = "backlog-ready-idle";
 
     /// <summary>
+    /// G574: a pre-publish queue item whose blocked representation is fully
+    /// converged (<c>state=blocked</c> and non-empty <c>blocked_by</c>).
+    /// Informational only: parking is intentional, so this kind names the
+    /// recorded reason and never recommends publishing or changing state.
+    /// </summary>
+    public const string KindBlockedParked = "blocked-parked";
+
+    /// <summary>
+    /// G574: a queue item whose two blocked fields disagree. This is
+    /// actionable drift rather than an intentional park; its recommendation
+    /// points at the canonical block/clear surface without choosing the
+    /// operator's intended direction.
+    /// </summary>
+    public const string KindStateDrift = "state-drift";
+
+    /// <summary>
     /// G544: default <c>--backlog-idle-minutes</c> threshold (45 minutes) —
     /// below G523's typical single-message recovery latency, matching
     /// <see cref="AutomationHeartbeatCommand.DefaultStaleMinutes"/>, so a
@@ -1477,6 +1493,15 @@ internal static class AutomationStalledWorkCommand
             return;
         }
 
+        // G574: inspect the two-field blocked representation before asking
+        // next-slice for a publishable candidate. The selector's fallback can
+        // surface state=blocked packets, while the reverse half-converged
+        // shape (non-blocked + blocked_by) is filtered out entirely. Reading
+        // queue-state directly here is therefore necessary both to suppress
+        // the unsafe publish recommendation and to keep reverse drift visible.
+        var nonPublishableUnits = CollectBacklogBlockedState(
+            context, domain, repo, now, items, excluded);
+
         IntentNextSliceResult nextSlice;
         try
         {
@@ -1506,6 +1531,13 @@ internal static class AutomationStalledWorkCommand
         }
 
         var executionUnit = nextSlice.Candidate.ExecutionUnit;
+        if (nonPublishableUnits.Contains(executionUnit))
+        {
+            // blocked-parked/state-drift already explains why this packet is
+            // not publishable. Never append the contradictory G544 action.
+            return;
+        }
+
         var runLogPath = context.GetRunLogPath();
         if (!File.Exists(runLogPath))
         {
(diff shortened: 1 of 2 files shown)


tomohisa:
## Semantic review — G574 / PR #1252 @ `c6b55b3` — **PASS**

Read-only review at frozen exact head `c6b55b3583606a37c6557d003daec640d673232b` (== PR tip, confirmed) in the managed worktree `.intent-cli/worktrees/review-G574`. `design_alignment_checked = true`.

**Checked sources:** `.intent-cli/issues/G574/{packet.yaml,implementation.md,review-context.md}`; issue #1251 (AC + Out Of Scope); `intents/intent-cli/intent-tree/means/03-state-and-audit-strategy.md`; `intents/intent-cli/intent-tree/means/01-queue-topology.md`; the exact two-file diff; `IntentNextSliceCommand` and `AutomationHeartbeatCommand` as unchanged collaborators; exact-head CI.

### Lens 1 — no publish recommendation can reach a blocked unit

I traced this to the mechanism rather than trusting the tests. `intent-cli issue publish-flow` appears as a recommendation at exactly **one** site in the whole command (`AutomationStalledWorkCommand.cs:1609`), and it is unreachable without first passing the new `nonPublishableUnits.Contains(executionUnit)` early return. Every other exit from `CollectBacklogReadyIdle` — WIP non-empty, next-slice throw, null candidate, missing/unparseable/empty runs.jsonl, below-threshold idle — produces no recommendation at all.

I also confirmed **why** the field incident was possible, which is what makes the fix correct rather than incidental: `IntentNextSliceCommand` only puts `QueueItemState.Queued` items into `queuedItemByUnit`, so its all-directories fallback (`Directory.EnumerateDirectories(packetRoot)`) re-applies the `blocked_by` eligibility gate **only** to queued units. A `state=blocked` unit has no entry there, so it slipped past the gate and became the candidate — exactly the shape the new code comment describes. Reading queue-state directly in `CollectBacklogBlockedState` is therefore genuinely necessary; a selector-side assumption would not have closed it.

The fail-safe direction is right too: `nonPublishableUnits.Add(...)` happens **before** the run-log transition lookup, so a unit with no usable age evidence still suppresses publishing and lands in `excluded[]` rather than silently becoming publishable again. That is the correct bias for this defect class.

### Lens 2 — visibility preserved, not deleted

`blocked-parked` is emitted as a real item with `is_informational: true`, the `blocked_by` reason joined into the recommended-action text, and age-since-block. I verified heartbeat carriage is structural rather than incidental: `AutomationHeartbeatCommand` counts and renders purely on `StalledWorkItem.IsInformational` with no kind allow-list, which is why the new kinds flow through with **zero** heartbeat code change — and the regression test proves the end-to-end result (`0 pending transition(s), 1 informational note(s)`, `FYI:`, the release-gate reason present, `publish-flow` absent).

### Lens 3 — age semantics

`Execute_PrePublishUnblock_RestartsBacklogIdleAgeAtCanonicalTransition_G574` does not hand-write the post-unblock fixture: it invokes the **canonical** `automation issue-block G570 --clear --pre-publish --write` surface (the G561 path) at `now-10m` over a unit blocked at `now-120m`, then asserts `backlog-ready-idle` with `age_minutes == 10` and no residual `blocked-parked`/`state-drift`. That is the AC's "measured from the unblock, not from packet creation", proven through the real surface. The underlying mechanism is the pre-existing "most recent `ts` across every row in runs.jsonl" baseline, which the canonical unblock advances — which is why this AC needs no production change.

### Lens 4 — byte-unchanged for genuinely queued units

`Execute_GenuinelyQueuedUnit_RetainsExactJsonBytes_G574` asserts the **entire** detector JSON document byte-for-byte, not a field subset — threshold fields, the single `backlog-ready-idle` item with its `publish-flow` recommendation, empty `excluded`/`warnings`, trailing newline included. That is the strongest available form of this AC.

### Acceptance criteria → evidence

| AC | Evidence |
|---|---|
| 1 — converged blocked → informational `blocked-parked`, reason + age-since-block, no publish rec, heartbeat carries | field-shape regression test (G570, `v0.7.1-release-gate`, age 45, `is_informational: true`, no `publish-flow`, no `backlog-ready-idle`) + heartbeat assertions |
| 2 — half-converged → non-informational `state-drift` with canonical-surface recommendation | `[Theory]` covering **both** shapes (`blocked` without `blocked_by`; `queued` with `blocked_by`), asserting `automation issue-block` / `--clear --pre-publish` and no `publish-flow` |
| 3 — post-unblock age from the unblock | canonical G561 `--clear --pre-publish` fixture, `age_minutes == 10` |
| 4 — field-shape fixture + queued byte-unchanged | the two tests above |
| 5 — full suite green, diff bounded | full suite green; diff is exactly two files |

### Boundaries

Diff touches **exactly two files** — `AutomationStalledWorkCommand.cs` and its test. `IntentNextSliceCommand` (selector), `AutomationIssueBlockCommand` (block/unblock semantics), `AutomationHeartbeatCommand`, `issue publish-flow`, guide content, and every other stalled-work kind are **untouched**, satisfying the issue's Out Of Scope verbatim. G573, dogfood, and release-prep surfaces are untouched. No AI provider launched from `intent-cli`; no GitHub or parent state mutated. `Closes #1251` present and matching the linked source issue (G311). Base `main`, OPEN, non-draft, MERGEABLE, CLEAN.

The new `Path.Combine(context.RepoRoot, ".intent-cli", "issues", …, "packet.yaml")` resolution matches the established convention — I checked it against ~10 existing call sites including four already in this same file — so it is not a new hardcoding.

### Design alignment

Node 03 makes the audit intent 「なぜ止まったか」を後から追跡できること. Reporting the park as an informational item that *names its reason* and rides the heartbeat serves that directly; silently dropping parked units would have satisfied the bug report while violating the node. Node 01 treats Clarify as a `selective block` under its own state machine, which is exactly why the implementation excludes `ClarifyBlocked` from G574 classification instead of mislabelling it drift.

### Evidence

| Check | Result |
|---|---|
| Full solution, 11/11 test projects | **4,565 total, 0 failed, 1 skipped** — matches the PR body |
| Focused detector/heartbeat/block filter | **150 passed, 0 failed** — reproduces the PR body's 150 exactly |
| G574-specific tests | 5 cases passed |
| `git diff --check` | clean |
| CI at exact head | **SUCCESS** — run 30732250595 / job 91454590940 |

**Tests-pass is necessary, not sufficient.** This PASS rests on the mechanism trace above — one publish site behind one guard, the selector-fallback root cause confirmed in `IntentNextSliceCommand`, heartbeat's kind-agnostic informational carriage, and the pre-existing run-log baseline that makes AC 3 hold — plus node 03 / node 01 alignment. The tests corroborate it; they are not the argument.

### Non-blocking follow-up (does not fail any acceptance criterion of this PR)

The suppression set is built only from queue items whose state is `Queued` or `Blocked`. A unit in **any other state** that still carries `blocked_by` is therefore neither classified nor suppressed, so if the selector's fallback surfaced it, `publish-flow` could still be recommended for a `blocked_by`-bearing unit. Scoping the residual honestly:

- `completed` — the dominant real case (32 of 1,514 entries in this host's queue-state carry `blocked_by` while not blocked; **all** are completed) is excluded upstream by the fallback's own `completed` guard, so it cannot reach the recommendation.
- `active` / `review` / `fixing` — normally unreachable because `DomainWipIsEmpty` gates the whole detector before this point.
- `clarify-blocked` — the genuine residual. The implementation deliberately excludes it with a documented rationale (a different state machine, per node 01), which I agree is right for *classification*: calling it `state-drift` would be a false alarm recommending `issue-block` on a clarification hold. But the same predicate also governs *suppression*, so the exclusion removes it from both. Note this is a narrower reading than AC 2's literal wording ("`blocked_by` without `state=blocked`"), which would textually include it. It has never occurred in this domain: **0 of 1,514** queue entries have ever been `clarify-blocked`.

I am not treating this as a blocker: every contracted AC is met and fixture-proven, the residual is pre-existing rather than introduced here, and widening suppression into the clarification lane would cross this issue's "no other kinds touched" boundary. Recording it so the operator can decide whether a follow-up packet should split "is it drift?" from "may it be published?" into two predicates.

### False-claim boundaries

- I did **not** execute the built binary against a live `automation stalled-work` run: the detector calls GitHub for its WIP gate, and the field-incident shape is only reachable through the test suite's `CandidateListerFactory` seam. Everything I report about blocked-parked, state-drift, and heartbeat behavior was verified by **reading the implementation and running the project's own fixtures**, not by observing a production stall report. No live orchestrator run was reproduced and none is claimed.
- The queue-state statistics above are from this host repository's `.intent-cli/queue-state.json` at the time of review; they bound observed history, not what is theoretically possible.
- No release, tag, publish, roll, or dogfood action was performed, and G573 was not re-reviewed here.

No merge, label, workflow, or parent-state mutation was performed by this review.

PR #1344, opened 2026-08-05
feat: add registry-limited topology field update

### LINKED ISSUE #1343: G620: A field that was never recorded
# G620: A field that was never recorded

## Goal

Add a **registry-limited field update path** to `session-layer topology` so a recorded role can adopt a declarative field it never carried — requiring the role, an **allowed** field name (initially only `delivery_method`), an explicit **current value that may be stated as absent**, the new value, an explicit confirmation, and `--dry-run|--write`. `record` stays strictly idempotent and keeps refusing conflicts; `update-kind` keeps working; arbitrary JSON paths are never addressable.

## Why This Slice Exists Now

**G619 shipped a delivery method that no existing team can turn on.** Measured on this host:

- `record --delivery-method file-backed` against the existing `review` role returns `conflict: true` — refusing to replace the recorded shape.
- The same call against a fresh, unrecorded role succeeds.
- The role's record carries **no `delivery_method` field at all** — it is *absent*, not set to `inline`.

So the remedy for a failure this team paid for three times is unreachable for every team that already recorded its topology, which is every team using it.

**The design question is sharper than a missing subcommand.** G592's conflict rule does not distinguish **changing a recorded value** from **declaring a field that was never recorded**. The second overwrites nothing. If absent-to-declared stays a conflict, every future optional field is unreachable for existing records, and G614's `update-kind` becomes a per-field pattern repeated without bound.

Option (c) chosen, verbatim: **keep `record` strict and idempotent, refusing conflicts** — it has caught real mistakes — and add a **registry-limited general update path**. It requires the role, an allowed field, an explicit `current` that may be stated as **absent**, the new value, confirmation, and `--dry-run|--write`. **Initially only `delivery_method` is allowed.** `update-kind` stays compatible. **Arbitrary JSON paths are never allowed.** Mechanism, docs/tests, and the post-merge adoption on this team's Copilot seat ship as one slice.

## Current Observed State

- No path exists to declare a previously-absent topology field on a recorded role.
- `update-kind` covers exactly one field; `record` refuses; hand-editing is prohibited where a command should exist.
- This team's `review` role therefore cannot adopt `file-backed` delivery.

- G592's refusal semantics for `record` are correct and must not be weakened.
- G614's `update-kind` is correct and stays; this slice does not replace it.
- G619's `delivery_method` semantics are correct; this slice only makes them adoptable.

## Target Repo / Path / Part

Target part: the topology update surface, its allowed-field registry, and ja/en guidance.

## In Scope

- **The registry-limited update path**: requires domain, team, role, an allowed field name, the stated current value (**`absent` is a valid statement**), the new value, an explicit confirmation flag, and `--dry-run|--write`. It refuses when the stated current value does not match reality — including stating a value when the field is absent, and stating absent when a value exists.
- **The allow-list registry**: initially `delivery_method` only. A field not in the registry is refused by name. **No arbitrary JSON path is addressable.**
- **Narrowness**: only the named field changes; every other field of that role stays byte-identical. `record` is untouched and gains no force flag; `update-kind` keeps working unchanged.
- **Adoption in the same slice**: after merge, this team's Copilot review seat declares `delivery_method: file-backed` through the new path, and the result is verified — one line reaching the pane, task file written before the pointer.
- **ja/en guidance** covering when the update path applies (declaring a previously-absent field, or changing an allowed one), why `record` still refuses, and why the registry exists.

## Out Of Scope

- Weakening `record`'s conflict refusal or adding `--force` to it.
- Allowing fields beyond the registry, or any arbitrary-path mutation.
- Changing `update-kind`, `delivery_method` semantics, delivery semantics, or recipient identity.

The child PR adds a registry-limited field update path to session-layer topology that requires the domain, team and role together with an allowed field name, the stated current value where the literal absent is a valid statement, the new value, an explicit confirmation flag and either dry-run or write, and that refuses when the stated current value does not match reality including the cases of stating a value while the field is absent and stating absent while a value exists. The allowed-field registry initially contains only delivery_method, a field outside the registry is refused by name, and no arbitrary JSON path is addressable. Only the named field changes while every other field of that role remains byte-identical, record is untouched and gains no force flag, and update-kind continues to work unchanged. After merge this team's Copilot review seat declares delivery_method file-backed through the new path and the result is verified with exactly one line reaching the pane and the task file written before the pointer. EN and JA guidance explains when the update path applies, why record still refuses conflicts, and why the registry exists; tests cover the absent-to-declared case, the mismatch refusals in both directions, the registry refusal, field-preservation of all other fields, and dry-run parity; ja/en stay in parity under the G613 terminology policy and the full suite is green.

## Acceptance Criteria

- Update path with role + allowed field + stated current (`absent` valid) + new value + confirmation + dry-run/write.
- Mismatch refused in both directions (value stated when absent; absent stated when present).
- Registry initially `delivery_method` only; out-of-registry field refused by name; no arbitrary paths.
- Only the named field changes; all other fields byte-identical; `record` untouched, no force flag; `update-kind` unchanged.
- Post-merge adoption on this team's seat verified: one line to the pane, file written before pointer.
- ja/en guidance; parity under G613; full suite green.

## Verification

- Declare `delivery_method` on a role where it is absent with `current=absent`: succeeds; diff shows only that field added.
- State a value while absent, and state absent while present: both refused.
- Attempt a field outside the registry: refused naming the field.
- Attempt `record --delivery-method` on a recorded role: still refuses (unchanged).
- Adoption on the live seat: measure the line count reaching the pane.

## Related Links

- G619 — the delivery method this slice makes adoptable; G614 — `update-kind`, the narrow precedent; G592 — the refusal semantics kept intact.
- Measurement, 2026-08-05: `record --delivery-method` on the recorded review role returns `conflict: true`; the field is absent, not `inline`.

## Knowledge Maintenance

- Intent placement: node 08 primary (topology surface), node 03 supporting (declared-state discipline).
- ADR candidate: none.
- Diagram candidate: none.
- Closeout writeback expected: yes — node 08 records that declaring an absent field is not the same as changing a recorded one, that the update path is registry-limited by design, and that `record` stays strict.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1343

## G620 contract
- Adds session-layer topology update-field, requiring domain, team, role, registry field, stated current value (including absent), new value, explicit confirmation, and dry-run or write.
- The registry admits only delivery_method; unknown and dotted names are refused. The field accepts only inline or file-backed and preserves every sibling field raw JSON value.
- Stale current values refuse in both directions. Dry-run wins over write in either flag order.
- record was not changed and remains strict/idempotent; update-kind, recipient identity, and delivery semantics are unchanged.
- EN/JA doc 12 explains the absent-field declaration path, why the registry exists, and why record conflict refusal remains.

## Verification
- Focused SessionLayerTopologyG604Tests and AgentMessageOrchestrationDocsTests: 37 passed.
- Full: dotnet test IntentSystem.sln --no-restore --verbosity quiet passed.
- git diff --check clean.

## Post-merge adoption
The issue requires the existing Copilot review seat to declare delivery_method file-backed after this command is merged. This child PR intentionally does not inspect or mutate host machine-local topology before merge. The required operator verification is: update the recorded role with current=absent and new=file-backed, then use canonical notify to measure the one-line pane pointer and confirm the durable task file exists before that pointer is sent. Existing G619 coverage pins file-before-pointer delivery; G620 coverage pins the new declaration path.

### DIFF
--- a/docs/en/12-agent-message-orchestration.md
+++ b/docs/en/12-agent-message-orchestration.md
@@ -219,8 +219,9 @@ herdr agent start <logical-role> --kind copilot --pane <pane-id> -- --model clau
   still measures 842 characters over 14 lines and can itself be pasted. It
   reduces duplication, not a paste-sensitive wedge; G619 owns the transport-layer
   remedy.
-- **Task-envelope delivery method.** A paste-sensitive herdr seat declares
-  `delivery_method: file-backed`: `notify` writes the unchanged envelope under
+- **Task-envelope delivery method.** A paste-sensitive herdr seat with an
+  existing record declares `delivery_method: file-backed` through the
+  registry-limited topology field update: `notify` writes the unchanged envelope under
   host `.intent-cli/tasks/<domain>/<team>/<task-id>-<nonce>.md` before sending
   the pane one line, `Read task envelope: <path>`. Declare `inline` explicitly
   when desired; an absent declaration preserves existing inline delivery.
@@ -410,6 +411,7 @@ by hand:
 intent-cli session-layer topology record --domain <domain> --team <team> --role <role> --resident herdr --workspace-id <workspace-id> --pane-id <pane-id> --cwd <role-cwd> [--kind <agent-kind>] --write
 intent-cli session-layer topology record --domain <domain> --team <team> --role <role> --resident external --reader <routing-root-relative-path> [--frontend <frontend>] --write
 intent-cli session-layer topology update-kind --domain <domain> --team <team> --role <role> --current-kind <kind> --new-kind <kind> --confirm-update-kind --write
+intent-cli session-layer topology update-field --domain <domain> --team <team> --role <role> --field delivery_method --current <absent|inline|file-backed> --new <inline|file-backed> --confirm-update-field --write
 intent-cli session-layer topology retire-legacy --domain <domain> --team <team> --evidence <named-fleet-migration-evidence> --confirm-retire-legacy --write
 intent-cli session-layer topology validate --domain <domain> --team <team> --format json
 intent-cli session-layer topology show --domain <domain> --team <team> --format json
@@ -418,8 +420,8 @@ intent-cli session-layer topology show --domain <domain> --team <team> --format 
 Agent kind is whatever herdr can start: Claude, Codex, Copilot, Cursor, OpenCode,
 and others are examples, not a supported-set restriction. Logical role defaults
 are `implementation`, `review`, `interview`, and `clarify`; existing explicit
-role mappings, including legacy product-named mappings, remain valid. Both new
-mutation commands emit JSON and support only `--format json`. For `update-kind`,
+role mappings, including legacy product-named mappings, remain valid. The three
+update/retire mutation commands emit JSON and support only `--format json`. For `update-kind`,
 an explicit `--dry-run` takes precedence over `--write` in either flag order and
 never writes.
 
@@ -443,6 +445,16 @@ the mapping exists or herdr-only requires it, and notify topology refusals point
 back to `topology validate` / `record` as the remedy. Invalid state always stays
 fail-closed; these commands add knowledge and a controlled writer, never a
 fallback.
+
+`update-field` is the narrow path for declaring a field that a recorded role
+never carried, or for changing its already recorded value. It requires the
+role, the field name, the stated current value, the new value, explicit
+confirmation, and `--dry-run` or `--write`; state `--current absent` only when
+the field is actually absent. A stale statement is refused in both directions.
+The registry initially permits only `delivery_method`, so an unknown or dotted
+name is refused rather than becoming an arbitrary JSON-path editor. The command
+changes only that field. It does not relax `record`: re-recording a different
+shape still refuses its conflict and has no force flag.
 
 #### Visible, generated mode markers
 
@@ -556,8 +568,10 @@ The measured limit matters: a minimal canonical `notify delegate` envelope is
 duplicated substance; it does not prevent a paste-sensitive seat from wedging.
 G619 owns the transport-layer remedy.
 
-For a paste-sensitive herdr seat, declare `delivery_method: file-backed` in its
-recipe. `notify` writes the unchanged envelope to an addressable durable task
+For a paste-sensitive herdr seat whose recorded role lacks `delivery_method`,
+use `topology update-field` with `--field delivery_method --current absent --new
+file-backed` and explicit confirmation; use the same path with the recorded
+current value for a later allowed change. `notify` writes the unchanged envelope to an addressable durable task
 file under `.intent-cli/tasks/<domain>/<team>/<task-id>-<nonce>.md` before it
 sends the pane the single-line `Read task envelope: <path>` pointer. The file is
 not deleted, so a restarted recipient can read the same task. Declare `inline`
--- a/docs/ja/12-agent-message-orchestration.md
+++ b/docs/ja/12-agent-message-orchestration.md
@@ -206,8 +206,8 @@ herdr agent start <logical-role> --kind copilot --pane <pane-id> -- --model clau
   remedy として扱ってはいけません。最小の canonical `notify delegate` envelope でも 842 文字・14 行で、
   これ自体が paste になります。これは重複を減らす discipline であり、paste-sensitive な wedge を防ぐ
   ものではありません。transport-layer の remedy は G619 が担当します。
-- **task-envelope delivery method。** paste-sensitive な herdr seat には
-  `delivery_method: file-backed` を宣言します。`notify` は unchanged な envelope を host の
+- **task-envelope delivery method。** existing record を持つ paste-sensitive な herdr seat では、
+  registry-limited topology field update で `delivery_method: file-backed` を宣言します。`notify` は unchanged な envelope を host の
   `.intent-cli/tasks/<domain>/<team>/<task-id>-<nonce>.md` に書いてから、pane には
   `Read task envelope: <path>` という 1 行だけを送ります。明示的に選ぶなら `inline` を宣言します。
   宣言がなければ既存の inline delivery をそのまま維持します。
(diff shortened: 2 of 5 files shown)


tomohisa:
## G620 review verdict — PR #1344 at head `8424629f96d7287f10146fb42b64a7be4304ca9c`

**Verdict: mechanism approved; slice not closeable yet.** Every mechanism boundary in the packet holds under independent measurement. One acceptance criterion (live-seat adoption) is stated as an operator instruction rather than a measurement, so G620 does not close on merge alone. Two minor findings below; neither blocks merge.

Verification was run in a worktree pinned to the exact head, against a scratch routing root — not from the PR's own test assertions alone.

### Packet-versus-PR delta, as required by review-context

**1. `record` untouched — cited grep.** `git diff HEAD~1 HEAD -- src/ | grep -iE 'force|Record|Conflict'` returns only two `record`-touching lines, both the subcommand usage string:

```
-  "Usage: ... topology record|show|validate|update-kind|retire-legacy [options]";
+  "Usage: ... topology record|show|validate|update-kind|update-field|retire-legacy [options]";
```

No `--force` flag is introduced anywhere; `SessionLayerTopologyWriter.Record` has no removed lines; no relaxed conflict path, no new branch that lets `record` write over a recorded value. Measured live at head, after `update-field` had already written the field: `record --delivery-method file-backed` against the recorded role still returns `conflict: true, applied: false`, file byte-unchanged. The justification for the new path holds.

**2. Both mismatch directions refuse — both exercised.**

- Value stated while field absent: `--current inline` on an absent field → `conflict: true`, `Role 'review' records delivery_method 'absent', not stated current value 'inline'. Refusing update-field.` File unchanged.
- `absent` stated while a value exists: `--current absent` against a recorded `file-backed` → `conflict: true`, `records delivery_method 'file-backed', not stated current value 'absent'`. File unchanged.
- Third direction also refuses: wrong value stated while a different value exists (`--current inline` vs recorded `file-backed`).

The stale-belief hole G614 closed for `kind` is closed here in both directions, not one.

**3. The registry is a real gate — refusal names the field, no path is addressable.** Every attempt below was refused with `conflict: true` and the offending name echoed in both the `field` key and the summary, and the topology file was byte-identical afterward:

| attempted `--field` | result |
|---|---|
| `kind` | `Field 'kind' is not in the topology update registry. Only 'delivery_method' is allowed.` |
| `cwd` | refused, names `cwd` |
| `roles.review.delivery_method` (dotted) | refused, names the dotted string |
| `roles/review/kind` (nested) | refused, names the string |
| `Delivery_Method` (case) | refused — `StringComparison.Ordinal`, no case-folding widening |

No dotted path, nested key, or sibling field name is addressable. The narrow tool did not become an editor. `--new bogus` is also refused before any file read.

**4. Field preservation is byte-level — cited diff.** Before/after the successful `--current absent --new file-backed --write`:

```
10c10,11
<       "kind": "copilot"
---
>       "kind": "copilot",
>       "delivery_method": "file-backed"
```

Per-key raw-JSON comparison of the role object: added `{delivery_method}`, removed `{}`; `resident`, `workspace_id`, `pane_id`, `cwd`, `kind` each byte-identical; top-level `domain`, `team`, `workspace_id` unchanged. Dry-run parity confirmed in both flag orders (`--write --dry-run` and `--dry-run --write` both yield `mode: dry-run`, `applied: false`, zero bytes written).

**5. The adoption is asserted, not measured — outstanding.** The PR body states the operator procedure as a requirement ("The required operator verification is: update the recorded role... then use canonical notify to measure...") and explicitly declines to inspect host state. It reports no line count and no file-before-pointer ordering. Confirmed independently on the host: `.intent-cli/topology/intent-cli/intent-cli-dev.json` role `review` (`kind: copilot`) still carries **no** `delivery_method` key, and the installed `intent-cli session-layer topology` usage still advertises only `record|show|validate|update-kind|retire-legacy` — no `update-field`.

This is a genuine gap against acceptance criterion 4, but it is structurally post-merge: the criterion itself says "after merge", the host record is machine-local and outside this repo, and the command does not exist in the installed CLI until this lands. The PR's refusal to mutate host state pre-merge is the correct call. **The consequence is scoping, not code:** merging #1344 satisfies criteria 1, 2, 3 and 5 only. G620 stays open until the host operator runs `update-field --field delivery_method --current absent --new file-backed` on the review seat and records the actual measurement — exactly one line reaching the pane, and the task file existing before the pointer is sent.

**6. Nothing else moves.** Diff is 5 files, +264/−12, single commit on merged `main` (`36b89ac9`). `src/` change is confined to `SessionLayerTopologyCommand.cs`; no notify, delivery, or recipient-identity file is touched. `update-kind` still applies cleanly at head (measured). `delivery_method` semantics, delivery semantics and recipient identity are unchanged. EN and JA doc 12 both updated with the same substance (when the path applies, why the registry exists, why `record` still refuses); ja/en parity per G613 holds and is pinned by a docs test asserting the paired strings in both languages. No AI provider is launched from `intent-cli`; no GitHub or parent-state mutation; no contract section skipped; no scope widening beyond the issue contract.

### Suite

- Focused `SessionLayerTopologyG604Tests` + `AgentMessageOrchestrationDocsTests`: **37 passed, 0 failed** — matches the PR's claim exactly.
- Full `dotnet test IntentSystem.sln`: **green**, `IntentSystem.Cli.Tests` 4473 passed / 1 skipped, all other assemblies passing.
- `git diff --check`: clean.

### Findings

**F1 — low, non-blocking. A non-string `delivery_method` escapes `Validate` and throws uncaught, breaking the JSON-only output contract.**

`UpdateField` reads the field with `fieldNode?.GetValue<string>()`. The surrounding `catch` covers `IOException`, `UnauthorizedAccessException` and `JsonException` only, so a non-string value throws `InvalidOperationException` straight through. Measured with a hand-edited `"delivery_method": 42`:

```
$ ... update-field --field delivery_method --current absent --new inline --confirm-update-field --write --format json
EXIT=1
STDOUT: (empty)
STDERR: An element of type 'Number' cannot be converted to a 'System.String'.
```

`topology validate` reports that same record as `"valid": true`, so the guard the code relies on ("Topology record is invalid; refusing update-field") does not catch it. It does fail closed — no write, non-zero exit — so this is a contract-shape issue, not a safety hole, and it is only reachable by the hand-editing this slice exists to eliminate. The fix is small and the code already anticipates the case: `var recorded = hasField ? actual ?? "invalid" : "absent"` handles JSON `null` correctly (measured: refuses with `records delivery_method 'invalid'`). Reading via `fieldNode is JsonValue v && v.TryGetValue<string>(out var s)`, or adding `InvalidOperationException`/`FormatException` to the catch, routes a numeric value down the same structured refusal.

**F2 — cosmetic. No-op re-declare reports `applied: true, changed: true` and rewrites the file.** With `delivery_method` already `inline`, `--current inline --new inline --confirm-update-field --write` reports `applied: true, changed: true` and `Updated only delivery_method...`, rewriting identical content. `record` distinguishes this case with `already_recorded`. A `changed: false` when the new value equals the recorded one would make the audit line honest and match `record`'s idempotence reporting.

**Nit (no action needed):** the sibling-preservation test iterates `beforeRole` and compares each key into `afterRole`, so it pins that nothing is changed or dropped but would not catch an unexpected *additional* sibling. The behaviour is correct — verified independently above that `delivery_method` is the only added key.

### Knowledge writeback (G461)

The closeout to node 08 is correctly identified and should record all three: declaring an absent field is not the same as changing a recorded one; the update path is registry-limited by design; `record` stays strict. Not yet written at this head — expected at closeout, not in this PR.

**Not merged**, per task scope.

PR #1356, opened 2026-08-06
G626: report observed working as in progress

### LINKED ISSUE #1355: G626: Working is not a failure
# G626: Working is not a failure

## Goal

Make a delivery that reached an observed working transition report a **successful non-terminal dispatch state** instead of the failure-shaped `working-did-not-settle` / `not-observed-within-bound` pair. The 10-second bound stays a fixed internal detail — **not** configurable, **not** recipe-declared — because the defect is the reporting, not the duration.

## Why This Slice Exists Now

**Measured, and reported independently by another team.** `NotifyTransport` uses a hard-coded `BoundedPromptTimeoutMilliseconds = "10000"`. Any recipient that works longer than ten seconds — which is every real review and every real implementation turn — produces:

```
receiver outcome: working-did-not-settle
settle outcome: not-observed-within-bound
Resend is forbidden because the recipient may still be performing this work.
```

The Zero4Racer team saw it on five of five delegates; this team has seen it on every delegate all session. **The normal, fully successful path always reports in failure vocabulary.**

**Why this is a contract defect, not a cosmetic one.** G598 exists precisely to separate *never submitted* from *submitted and busy*, because those two demand opposite responses. A hard bound that turns "submitted and busy" into `did-not-settle` **quietly re-merges the two** in the reader's mind: the operator, the orchestrator, and any automated caller all learn to ignore a warning that fires on success — and an ignored warning cannot signal the real failure it was built for.

**Do not make the settle wait configurable or recipe-declared.** Return **observed-working as a successful non-terminal dispatch state**. This keeps G598's separation intact and adds no knob: a caller that observed the unattended working transition has the information it needs, and lengthening a timeout would only move the same false alarm later.

## Current Observed State

- `delivered: true` and an observed working transition still produce `working-did-not-settle` / `not-observed-within-bound`.
- `resend_permitted: false` is correct in that state, but is delivered alongside failure-shaped vocabulary.
- Callers cannot distinguish "delivered, still working" from "never settled for a bad reason" by outcome name alone.

- G598's separation is correct: `delivered` comes from the observed unattended working transition; settle is reported separately.
- `resend_permitted: false` after an observed working transition is correct and stays.
- The 10-second bound as an internal implementation detail is acceptable; it is not the defect.

## Target Repo / Path / Part

Target part: the dispatch outcome vocabulary and its documentation.

## In Scope

- **A successful non-terminal outcome**: when delivery reached an observed unattended working transition and the recipient had not settled within the internal bound, the reported outcome names that state as **success in progress**, not as a failure to settle. The machine field distinguishes it from a genuinely unobserved transition.
- **The genuine failure keeps its own name**: idle-stays-idle — never submitted — remains clearly distinguishable, with `resend_permitted: true` as before.
- **`resend_permitted` unchanged**: still `false` after an observed working transition, still `true` for never-submitted.
- **Human-readable text follows the machine state**: no sentence that reads as a warning when the state is success; the "resend is forbidden" line is phrased as a consequence of the recipient working, not as a caution about a missed observation.
- **ja/en guidance** updated so the outcome vocabulary is documented with which states are success, which are failure, and what each implies for resending.

## Out Of Scope

- Making the settle wait configurable or recipe-declared.
- Changing `delivered` semantics, recipient identity, or the observation mechanism.
- Changing the bound's value.

The child PR changes the dispatch outcome vocabulary so that a delivery which reached an observed unattended working transition but had not settled within the internal bound is reported as a successful non-terminal state rather than as a failure to settle, with a machine field that distinguishes it from a genuinely unobserved transition, while the never-submitted idle-stays-idle case keeps its own clearly distinguishable outcome. The resend_permitted values are unchanged, remaining false after an observed working transition and true for never-submitted, and the human-readable text follows the machine state so that no sentence reads as a warning when the state is success, with the resend-is-forbidden line phrased as a consequence of the recipient still working rather than as a caution about a missed observation. The settle wait remains a fixed internal detail and is neither made configurable nor recipe-declared, its value is unchanged, and delivered semantics, recipient identity and the observation mechanism are untouched. EN and JA guidance documents the outcome vocabulary including which states are success, which are failure, and what each implies for resending; tests cover the observed-working success state, the never-submitted failure state, the resend_permitted values in both, and the absence of warning-shaped text in the success path; ja/en parity under the G613 terminology policy and the full suite is green.

## Acceptance Criteria

- Observed working transition without settle ⇒ successful non-terminal outcome, distinguishable in machine output from an unobserved transition.
- Never-submitted keeps a distinct failure outcome.
- `resend_permitted` values unchanged in both states.
- Human text matches the machine state; no warning-shaped sentence on the success path.
- Bound unchanged and not exposed as configuration.
- ja/en outcome-vocabulary documentation; parity under G613; full suite green.

## Verification

- Dispatch to a recipient that works longer than the bound: assert a success-shaped outcome and `resend_permitted: false`.
- Dispatch to an idle recipient that never starts: assert the failure outcome and `resend_permitted: true`.
- Diff the human-readable text in the success path for warning vocabulary.
- Assert no new configuration surface appears for the bound.

## Related Links

- Field report, 2026-08-05: five of five delegates reporting the failure pair on success; verified here as `BoundedPromptTimeoutMilliseconds = "10000"` in `NotifyTransport`.
- G598 — the separation this defect quietly re-merges.

## Knowledge Maintenance

- Intent placement: node 08 primary (delivery outcome vocabulary), node 03 supporting (a warning that fires on success stops being a signal).
- ADR candidate: none.
- Diagram candidate: none.
- Docs update: ja/en delivery-outcome guidance (in scope).
- Closeout writeback expected: yes — node 08 records that observed-working is a success state, and node 03 that a warning which fires on the normal path teaches readers to ignore it.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1355

## Summary
- report observed working without a later settle acknowledgement as `working-observed-in-progress` / `pending`
- preserve idle-stays-idle as the distinct never-submitted outcome and keep resend permissions unchanged
- document the success-in-progress outcome in EN/JA and the installed herdr-only guide

## Verification
- `dotnet test tests/IntentSystem.Cli.Tests/IntentSystem.Cli.Tests.csproj -c Release --filter FullyQualifiedName~SessionLayerPreflightG594Tests` (19 passed)
- `dotnet test IntentSystem.sln -c Release`
- `git diff --check`


tomohisa:
## Review verdict: APPROVE — G626 / PR #1356 (Issue #1355)

**Execution unit / packet contract**: `G626`, packet `.intent-cli/issues/G626/`. Head `e7643bc`, one commit, 6 files, base `main` @ `06f1a71` (= current tip). Packet pre/post: the observed-working dispatch state changes from failure-shaped to success-shaped vocabulary, with `resend_permitted`, the bound, `delivered` semantics, recipient identity and the observation mechanism all held fixed.

I did not review this from the diff alone. `HerdrNotifyTransport` resolves its executable through `INTENT_CLI_HERDR_EXECUTABLE`, so I drove **the real CLI end-to-end against a stub herdr** — no live pane touched — and ran the identical scenario against a binary built from `06f1a71` for a true before/after.

---

## The defect, reproduced and fixed — identical scenario, two binaries

Recipient enters working, then does not settle within the bound (the everyday path):

```
########## BASE 06f1a71 (current main)
  receiver_state_outcome = working-did-not-settle
  settle_outcome         = not-observed-within-bound
  resend_permitted       = False
  active_phase.status    = not-observed
  summary = Team 'intent-cli-dev' workspace 'wH' pane 'wH:p1' entered unattended working, but did not
            return to idle/done/blocked within 10000ms ({"error":{"code":"timeout",...}}). Delivered
            after the observed working transition; settled acknowledgement was not observed within the
            bound. Resend is forbidden because the recipient may still be performing this work.

########## HEAD e7643bc
  receiver_state_outcome = working-observed-in-progress
  settle_outcome         = pending
  resend_permitted       = False
  active_phase.status    = in-progress
  summary = Delivery succeeded to team 'intent-cli-dev' workspace 'wH' pane 'wH:p1' after the observed
            unattended working transition. The recipient is still working and its settled
            acknowledgement is pending; do not resend while it is working.
```

**Corroboration from this seat**: the BASE output above is byte-for-byte what every `intent-cli notify report` I have run this session has returned — including the ones reporting *successful* G623/G624/G625 reviews. The field report's "five of five delegates" matches my own five of five. This is the normal path, and on `main` it speaks in failure vocabulary.

---

## Acceptance criteria

| # | Criterion | Evidence |
|---|---|---|
| 1 | Observed working ⇒ successful non-terminal outcome, distinguishable in machine output from an unobserved transition | live run above: `working-observed-in-progress` / `observed` / `pending` / `delivered: true`. Differs from the unobserved case in **four** fields (see row 2), so an automated caller can branch on the outcome name alone — lens 2 satisfied |
| 2 | Never-submitted keeps a distinct failure outcome | live run, idle pane that never starts: `cause: receiver-transition-unobserved`, `receiver_state_outcome: idle-stays-idle`, `working_transition: not-observed`, `settle_outcome: not-applicable`, `delivered: false`, **exit 1**, and failure-shaped prose retained: *"Not delivered; resend is permitted because the required unattended working transition was never observed."* |
| 3 | `resend_permitted` unchanged in both states | verified by direct before/after comparison, not by reading: `False` on base and head for observed-working; `True` for never-submitted. The diff changes no `ResendPermitted` assignment anywhere (lens 3) |
| 4 | Human text matches machine state; no warning-shaped sentence on the success path | scanned both success strings for `but`, `did not`, `could not`, `not observed`, `fail`, `forbidden`, `warn`, `caution`, `missed`, `timeout`, `within the bound` → **clean**. Read aloud, nothing prompts an operator to investigate. Critically, the retained caution is phrased exactly as the packet requires — *"do not resend while it is working"* as a **consequence of the recipient working**, replacing *"Resend is forbidden because the recipient may still be performing this work"*, which was a caution about a missed observation |
| 5 | Bound unchanged and not exposed as configuration | `BoundedPromptTimeoutMilliseconds = "10000"` untouched — the diff's only `BoundedPromptTimeout` hits are a context line and a deleted line, no value change (lens 5). No flag, option, recipe field or config key added; the sole grep hit for configuration-shaped additions is guide prose (lens 4) |
| 6 | ja/en outcome-vocabulary documentation; parity under G613; full suite green | `docs/en/12` and `docs/ja/12` receive equivalent edits, both explicitly naming it **"a successful non-terminal dispatch"** / **「成功した non-terminal dispatch」**, with `resend_permitted` implications stated for each state; JA keeps field names and technical terms in English per G613. Suite below |

**Vocabulary migration is complete** — `working-did-not-settle` and `not-observed-within-bound` appear **nowhere** in the repository at this head (src, docs, tests). The third surface, `HerdrOnlyOperatingGuide`, was updated too, so the CLI's own operating guidance cannot contradict the new outcome names. That completeness is the failure mode I flagged on the sibling G625 PR, and it is not present here.

`ActiveInProgress = "in-progress"` is a **new** preflight status rather than a reuse of `ActiveNotObserved`, so the active-phase status distinguishes the states as well. The two remaining `ActiveNotObserved` uses are the never-submitted and prompt-failure paths — correct, and correctly left alone.

---

## Out-of-scope boundaries: not crossed

No configurable or recipe-declared bound; no bound-value change; no resend-semantics change; no `delivered`-semantics, recipient-identity or observation-mechanism change (the diff contains no `Delivered =`, `Resolved =`, `ResendPermitted =`, `WorkingTransition =`, pane/workspace-resolution or `agent prompt`/`agent wait` argument edits outside prose). No `.intent-cli/**` or `intents/**` path touched; no label mutation knowledge; no `intent-target`/`intent-pr-created`; no AI provider launched.

**Intent consistency**: `means/08-agent-message-orchestration.md` (delivery outcome vocabulary) primary, `means/03-state-and-audit-strategy.md` supporting. This is precisely node 03's point — a warning that fires on the normal path teaches readers to ignore it, and an ignored warning cannot signal the real failure it was built for. G598's separation is preserved: `delivered` still comes from the observed working transition, settle is still reported independently, and the never-submitted case keeps `resend_permitted: true`.

---

## Non-blocking finding — the settle probe that could not run

Not a merge blocker and **no action requested on this PR**; recording it because it is adjacent to the honesty principle this slice defends.

`NotifyTransport.cs:465` catches `InvalidOperationException` from the settle probe. That exception means the herdr process **could not be started at all** (`Notification transport '<x>' could not start: …`, from `Win32Exception`/`IOException`). The PR now reports that case with the same success payload as the healthy still-working case — `settle_outcome: pending`, *"The recipient is still working and its settled acknowledgement is pending"* — and **drops the `exception.Message`** that the previous code included.

So in the one case where the settle observation never executed, the output asserts a determined state about the recipient that was never observed, and the diagnostic explaining why is discarded. That is the same shape as the G594/G625 ruling this repo just merged: an unresolved lookup should be reported as unresolved, never as a determined value.

Why I am **not** blocking on it: AC 1's distinguishability requirement is scoped to "distinguishable from an unobserved transition", and it is; the two cases were already indistinguishable by outcome name *before* this PR, so the conflation is inherited, not introduced; and the path is rare, since herdr has already launched successfully twice in the same invocation by the time it is reached. The only true regression is the dropped diagnostic.

A follow-up could give this branch `settle_outcome: unchecked` with the launch failure retained in the summary — additive, and it touches neither `resend_permitted`, the bound, nor any configuration surface.

**Second, minor**: `idle-transitions` — the fully-settled terminal success state — is still not named in either doc (pre-existing; absent at base too). With G626 documenting the *non-terminal* success state explicitly, the terminal one is now the only outcome a reader cannot look up.

---

## Test commands executed

```
dotnet test IntentSystem.sln -c Release
  → exit 0; IntentSystem.Cli.Tests: 4496 passed, 1 skipped, 0 failed
    (skip is the pre-existing TaskingHandoffCommandTests.Execute_TolerantToSubAnalyzerFailure_…)
  → ConceptIntake 36, Supervisor 66, Projection 80, Review 35, Workflow 21, DomainBinding 13,
    Clarify 47, Drift 6, DogfoodingBridge 10, WorkerAdapter 16 — all pass, 0 failures overall
git diff --check 06f1a71 e7643bc → clean (exit 0)
```
CI "Build and test (source contract)" passes. `Closes #1355` matches `publish.yaml` `created_issue_number: 1355` (G311). Base branch `main` per `direct-main`.

The total test count is unchanged because the PR **retargets** existing assertions rather than adding methods. I checked that this still satisfies the packet's coverage clause: the success state, the never-submitted state and `resend_permitted` in both are covered by the four-row `_G598` theory (now carrying a `working-observed-in-progress` row asserting `pending` / `false` / `in-progress`), and the renamed `HerdrDelivery_ObservedWorkingWithoutFreshSettledAck_IsSuccessfulAndInProgress_G626` adds the warning-vocabulary negatives — `DoesNotContain("Resend is forbidden")` and `DoesNotContain("not observed", OrdinalIgnoreCase)` — which is the assertion that would catch a regression into failure vocabulary.

**Tests-pass is necessary but not sufficient** — the green suite is paired above with a live before/after through the real binary on both the success and the never-submitted path, plus the packet-clause mapping, because a vocabulary contract is exactly the kind of change a passing suite can accommodate while still reading wrong to an operator.

Approving.

PR #1358, opened 2026-08-06
G627: retire legacy topology compatibility read

### LINKED ISSUE #1357: G627: The compat read has no readers left
# G627: The compat read has no readers left

## Goal

Remove the legacy `role-pane-mapping.json` compatibility read and move its ledger disposition off `retain-through-1.x`, so the 1.0 line does not commit to maintaining a path with no remaining consumers. The removal is expressed as a **fail-closed error naming the recovery commands**, never a silent stop: a host that has only the legacy file and no per-team record is told to run `topology record` or `retire-legacy` rather than discovering that delivery stopped resolving.

## Why This Slice Exists Now

**Operator decision, 2026-08-05**: include the retirement in 0.12.0 rather than deferring it, and fix the 1.0 feature set at that release.

**The ledger's condition is now satisfied, and it was measured rather than assumed.** The row reads `retain-through-1.x` with the note that *G604 migration/re-record evidence is required before any retirement is eligible*. Since then:

- The Zero4Racer team ran `session-layer topology retire-legacy` with named evidence, producing **entry one** in `legacy-topology-retirements.jsonl`. They verified before deleting that the legacy file held nothing the per-team record lacked — four roles, pane, cwd and workspace all identical, no other team present.
- **All four hosts on this machine are already free of the legacy file**: this team's design host (retired during the G604 migration), the two Sekiban hosts that share it, and the Zero4Racer host. **The compat read has no observed consumer.**

**Keeping it costs more than removing it.** Carrying the path into 1.0 means maintaining it through the whole 1.x line — the next removal opportunity is 2.0, which is where remote-herdr work is headed and therefore some distance away. It has already caused harm: the dual-location conflict pushed a team into hand-synchronising two files.

**The honest limit of the evidence**: what was measured is this machine's four hosts plus one recorded retirement. Environments elsewhere may still carry the file. That is exactly why the removal must fail closed with instructions rather than silently changing behaviour.

Split from the versioning-policy slice: this one is a removal with a fail-closed contract, and the freeze/preview policy is versioning substance. Combining them would let a reviewer approve the policy on the strength of this removal's evidence.

## Current Observed State

- The per-team record is authoritative; the legacy fixed path is read only when the per-team record is absent.
- The ledger row is `retain-through-1.x` pending migration evidence, which now exists.
- Nothing tells a host that still has only the legacy file what to do.

- G604's per-team topology is correct and stays; `retire-legacy` from G614 stays as the supported migration command.
- The retirement evidence ledger is the record the disposition change cites.
- Delivery semantics, recipient identity, and the settle contract are untouched.

## Target Repo / Path / Part

Target part: the legacy compat read, its ledger row, and ja/en guidance.

## In Scope

- **Remove the compat read**: the legacy fixed `role-pane-mapping.json` is no longer consulted when the per-team record is absent.
- **Fail closed with instructions**: when a team has no per-team record and a legacy file is present, every reader that previously fell back — topology show/validate, the session-layer preflight, notify's recipient resolution — reports a specific error naming the file it found, that the compat read was removed, and the two commands that resolve it (`topology record` to declare the current shape, or `retire-legacy` to retire the file with evidence). When neither is present, the existing configuration-incomplete report is unchanged.
- **Ledger disposition updated**: the `role-pane-mapping compatibility read` row moves off `retain-through-1.x` to reflect removal, citing the retirement-ledger evidence rather than asserting it.
- **ja/en guidance**: what changed, why, and the exact recovery steps for a host that still has the legacy file.

## Out Of Scope

- Removing `retire-legacy` itself — it remains the supported way to retire the file with recorded evidence.
- Changing the per-team record format, delivery semantics, recipient identity, or the settle contract.
- Auto-migrating anyone: no reader may write a per-team record on a host's behalf.

The child PR removes the legacy role-pane-mapping.json compatibility read so that the fixed file is no longer consulted when a team's per-team record is absent, and replaces the fallback with a fail-closed error in every reader that previously relied on it — topology show and validate, the session-layer preflight, and notify's recipient resolution — where the error names the legacy file that was found, states that the compatibility read has been removed, and names both topology record and topology retire-legacy as the commands that resolve the situation, while a host with neither a per-team record nor a legacy file keeps the existing configuration-incomplete report unchanged. The 1.0 compatibility ledger row for the role-pane-mapping compatibility read moves off retain-through-1.x to reflect the removal and cites the retirement evidence ledger rather than asserting that migration occurred. No reader writes a per-team record on a host's behalf, retire-legacy remains available as the evidence-recording retirement command, and the per-team record format, delivery semantics, recipient identity and settle contract are unchanged. EN and JA guidance explains what changed, why, and the exact recovery steps for a host that still has the legacy file; tests cover the fail-closed path with a legacy file and no per-team record, the unchanged configuration-incomplete path with neither, the normal path with a per-team record present, and the absence of any automatic migration; ja/en parity under the G613 terminology policy and the full suite is green.

## Acceptance Criteria

- Compat read removed; no reader falls back to the legacy fixed path.
- Legacy-present-without-record ⇒ fail-closed error naming the file, the removal, and both recovery commands, in every previously-falling-back reader.
- Neither present ⇒ existing configuration-incomplete report unchanged.
- Ledger row updated off `retain-through-1.x`, citing the evidence ledger.
- No automatic migration anywhere; `retire-legacy` still available.
- ja/en recovery guidance; parity under G613; full suite green.

## Verification

- Fixture with legacy file only: assert the fail-closed error in each reader and that both commands are named.
- Fixture with neither: assert the unchanged configuration-incomplete report.
- Fixture with a per-team record: assert normal resolution.
- Assert no code path writes a per-team record automatically.

## Related Links

- Zero4Racer retirement, 2026-08-05 — entry one in `legacy-topology-retirements.jsonl`, verified content-identical before deletion.
- Design measurement, 2026-08-05 — all four hosts on this machine already free of the legacy file.
- G604 — per-team topology; G614 — `retire-legacy`; G612 — the ledger rule this satisfies.

## Knowledge Maintenance

- Intent placement: node 08 primary (topology resolution), node 03 supporting (evidence-gated disposition change).
- ADR candidate: none.
- Diagram candidate: none.
- Docs update: ja/en topology and recovery guidance (in scope).
- Closeout writeback expected: yes — node 08 records that the compat read was removed on measured evidence and that its removal fails closed with the recovery commands named.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1357

Removes legacy role-pane-mapping fallback, fails closed with record/retire recovery commands, and updates EN/JA ledger guidance.

Verification: focused topology/preflight/JA guards; full suite; git diff --check.

### DIFF
--- a/docs/en/1.0-compatibility-ledger.md
+++ b/docs/en/1.0-compatibility-ledger.md
@@ -197,7 +197,7 @@ Every `cause` value actually emitted in a machine-consumed JSON payload is cover
 | per-team topology record | Canonical team topology record and its identity/path validation | `stable-at-1.0` | Record-first; no inferred mode change. |
 | judgment-wait record | `.intent-cli/operator-attention.json`; `open` → `resolved` or `superseded` | `stable-at-1.0` | G623 names the general durable record; its on-disk format and identifiers remain unchanged. |
 | operator-attention command alias | Alias for `judgment-wait`; no record schema or identifier change | `deprecate-with-alias` | Structured `deprecation_warning` names the replacement; alias is retained through 1.x and removal is only next MAJOR. |
-| role-pane-mapping compatibility read | Legacy fixed-file read when the current per-team record is absent | `retain-through-1.x` | G604 migration/re-record evidence is required before any retirement is eligible. |
+| role-pane-mapping compatibility read | Removed legacy fixed-file read; legacy-only hosts fail closed with record/retire recovery commands | `retire-before-1.0` | Retirement evidence: `legacy-topology-retirements.jsonl` entry one (Zero4Racer); all four measured hosts are legacy-free. |
 | runtime-state shapes | Runtime JSON state consumed by automation | `retain-through-1.x` | Inventory only; no automatic removal. |
 | packet-schema variants | Supported packet YAML/schema variants and state linkage | `retain-through-1.x` | A future `retire-before-1.0` row must name migration evidence. |
 | field aliases | Documented compatibility field aliases | `retain-through-1.x` | Inventory only; no inference or removal. |
--- a/docs/en/12-agent-message-orchestration.md
+++ b/docs/en/12-agent-message-orchestration.md
@@ -426,7 +426,11 @@ update/retire mutation commands emit JSON and support only `--format json`. For 
 an explicit `--dry-run` takes precedence over `--write` in either flag order and
 never writes.
 
-After a successful `retire-legacy`, the CLI appends one fleet-citable entry to
+The legacy fixed `role-pane-mapping.json` compatibility read is removed. If a
+host still has that file but lacks its per-team record, readers fail closed and
+name `topology record --domain <domain> --team <team> ... --write` and
+`topology retire-legacy --domain <domain> --team <team> --evidence <evidence>
+--confirm-retire-legacy --write`; no reader auto-migrates it. After a successful `retire-legacy`, the CLI appends one fleet-citable entry to
 `<host-repo>/.intent-cli/legacy-topology-retirements.jsonl`, outside the ignored
 machine-local topology directory. Its defined fields are `timestamp_utc`, `host`,
 `domain`, `team`, `retired_path`, and named `evidence`, so a later ledger decision
--- a/docs/ja/1.0-compatibility-ledger.md
+++ b/docs/ja/1.0-compatibility-ledger.md
@@ -198,7 +198,7 @@ machine-consumed JSON payload で実際に emit されるすべての `cause` va
 | per-team topology record | canonical team topology record と identity/path validation | `stable-at-1.0` | record-first; inferred mode change はしない。 |
 | judgment-wait record | `.intent-cli/operator-attention.json`; `open` → `resolved` または `superseded` | `stable-at-1.0` | G623 が general な durable record を命名し、on-disk format と identifier は変えない。 |
 | operator-attention command alias | `judgment-wait` の alias。record schema と identifier は変更しない。 | `deprecate-with-alias` | structured `deprecation_warning` が replacement を示し、alias は 1.x を通じて残り removal は next MAJOR のみ。 |
-| role-pane-mapping compatibility read | current per-team record が absent の場合の legacy fixed-file read | `retain-through-1.x` | G604 migration/re-record evidence がなければ retirement は eligible ではない。 |
+| role-pane-mapping compatibility read | removed になった legacy fixed-file read。legacy-only host は record/retire recovery command を示して fail-closed する | `retire-before-1.0` | retirement evidence: Zero4Racer の `legacy-topology-retirements.jsonl` entry one と、legacy-free を確認した 4 host。 |
 | runtime-state shape | automation が consume する runtime JSON state | `retain-through-1.x` | inventory のみ。automatic removal ではない。 |
 | packet-schema variant | supported packet YAML/schema variant と state linkage | `retain-through-1.x` | future `retire-before-1.0` row は migration evidence を名指しする。 |
 | field alias | documented compatibility field alias | `retain-through-1.x` | inventory のみ。inference や removal はしない。 |
--- a/docs/ja/12-agent-message-orchestration.md
+++ b/docs/ja/12-agent-message-orchestration.md
@@ -394,7 +394,7 @@ agent kind は herdr が起動できる任意の kind です。Claude、Codex、
 `update-kind` では明示した `--dry-run` が flag の順序にかかわらず `--write` より優先され、決して
 書き込みません。
 
-`retire-legacy` が成功すると、CLI は ignored な machine-local topology directory の外にある
+legacy fixed `role-pane-mapping.json` の compatibility read は削除済みです。この file が残り per-team record がない host では reader は安全側に停止し、`topology record --domain <domain> --team <team> ... --write` と `topology retire-legacy --domain <domain> --team <team> --evidence <evidence> --confirm-retire-legacy --write` を示します。reader は自動移行しません。`retire-legacy` が成功すると、CLI は ignored な machine-local topology directory の外にある
 `<host-repo>/.intent-cli/legacy-topology-retirements.jsonl` へ fleet-wide decision から引用可能な
 entry を 1 件追記します。定義済みの field は `timestamp_utc`、`host`、`domain`、`team`、
 `retired_path`、名前付きの `evidence` です。これにより、現在の legacy reader disposition を
--- a/src/IntentSystem.Cli/Commands/NotifyRoleTopology.cs
+++ b/src/IntentSystem.Cli/Commands/NotifyRoleTopology.cs
@@ -364,40 +364,22 @@ internal static class NotifyRoleTopologyStore
                 path = matches[0];
             }
         }
-        if (!File.Exists(path) && !string.IsNullOrWhiteSpace(domain) && File.Exists(ResolvePath(routingRoot)))
+        if (!File.Exists(path) && !string.IsNullOrWhiteSpace(domain))
         {
-            path = ResolvePath(routingRoot);
-        }
-
-        if (!string.IsNullOrWhiteSpace(domain)
-            && File.Exists(ResolvePath(routingRoot, domain, team))
-            && File.Exists(ResolvePath(routingRoot)))
-        {
-            var dualLocation = Resolve(routingRoot, domain, team);
-            if (!dualLocation.Resolved
-                && string.Equals(dualLocation.Cause, "topology-location-conflict", StringComparison.Ordinal))
+            var legacyPath = ResolvePath(routingRoot);
+            if (File.Exists(legacyPath))
             {
-                findings.Add(Finding("<topology>", "location", dualLocation.Cause!, dualLocation.Summary));
+                var compatibility = LegacyCompatibilityReadRemoved(legacyPath, domain, team);
+                findings.Add(Finding("<topology>", "file", compatibility.Cause!, compatibility.Summary));
                 return Validation(team, path, findings);
             }
         }
-
-        var warnings = !string.IsNullOrWhiteSpace(domain)
-            && string.Equals(path, ResolvePath(routingRoot), StringComparison.Ordinal)
-            && File.Exists(path)
-                ? new[]
-                {
-                    $"Deprecated topology compatibility read from '{path}'; run "
-                    + $"`intent-cli session-layer topology record --domain {domain} --team {team} ... --write` to re-record this "
-                    + "machine's topology at its per-team local path.",
-                }
-                : [];
 
         if (!File.Exists(path))
         {
             var resolution = Resolve(routingRoot, domain, team);
             findings.Add(Finding("<topology>", "file", resolution.Cause!, resolution.Summary));
-            return Validation(team, path, findings, warnings);
+            return Validation(team, path, findings);
         }
 
         if (!string.IsNullOrWhiteSpace(domain) && !string.Equals(path, ResolvePath(routingRoot), StringComparison.Ordinal))
@@ -414,7 +396,7 @@ internal static class NotifyRoleTopologyStore
                         $"Topology file '{path}' identifies domain '{recordedDomain ?? "missing"}' team "
                         + $"'{recordedTeam ?? "missing"}', but its path was requested for domain '{domain}' "
                         + $"team '{team}'. Refusing the copied or misplaced machine record."));
-                    return Validation(team, path, findings, warnings);
+                    return Validation(team, path, findings);
                 }
             }
             catch (Exception exception) when (exception is IOException or UnauthorizedAccessException or JsonException)
@@ -430,7 +412,7 @@ internal static class NotifyRoleTopologyStore
             {
                 findings.Add(Finding("<topology>", "team", "topology-team-missing",
                     $"Topology file '{path}' {teamError}"));
-                return Validation(team, path, findings, warnings);
+                return Validation(team, path, findings);
             }
 
             var rolesElement = teamElement.TryGetProperty("roles", out var nestedRoles)
@@ -440,7 +422,7 @@ internal static class NotifyRoleTopologyStore
             {
                 findings.Add(Finding("<topology>", "roles", "topology-invalid",
                     $"Team '{team}' has no object-valued roles field."));
-                return Validation(team, path, findings, warnings);
+                return Validation(team, path, findings);
             }
 
             var teamWorkspaceId = ReadString(teamElement, "workspace_id")
@@ -528,7 +510,7 @@ internal static class NotifyRoleTopologyStore
                 $"Topology file '{path}' is unreadable: {exception.Message}"));
         }
 
-        return Validation(team, path, findings, warnings);
+        return Validation(team, path, findings);
     }
 
     public static bool TryResolveReaderPath(
--- a/src/IntentSystem.Cli/Commands/NotifyRoleTopology.cs
+++ b/src/IntentSystem.Cli/Commands/NotifyRoleTopology.cs
@@ -95,6 +95,15 @@ internal static class NotifyRoleTopologyStore
         $"Run `intent-cli session-layer topology validate --domain <domain> --team {team} --format json`, then use "
         + "`session-layer topology record --domain <domain> ... --write` to record the operator-supplied correction.";
 
+    private static NotifyTopologyResolution LegacyCompatibilityReadRemoved(string legacyPath, string? domain, string team) =>
+        Failure(
+            "legacy-topology-compatibility-removed",
+            $"Legacy topology file '{legacyPath}' was found for team '{team}', but the compatibility read has been removed. "
+            + $"Run `intent-cli session-layer topology record --domain {domain ?? "<domain>"} --team {team} ... --write` "
+            + "to declare the current shape, or run "
+            + $"`intent-cli session-layer topology retire-legacy --domain {domain ?? "<domain>"} --team {team} "
+            + "--evidence <evidence> --confirm-retire-legacy --write` to retire the legacy file with evidence.");
+
     public static NotifyTopologyResolution Resolve(string routingRoot, string team) =>
         Resolve(routingRoot, domain: null, team);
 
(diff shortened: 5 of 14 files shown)


tomohisa:
## Re-review verdict: REQUEST UPDATE — G627 / PR #1358 (Issue #1357)

Head `d5932be`, base `main` @ `2bb20d3` (= current tip), two commits, 14 files.

**The CI fixture repair — the thing this re-review was dispatched to check — is clean.** I confirm the *acceptable* branch of the review-context fork, not the fake one. But constructing the legacy-only fixture that lens 1 requires turned up **one blocking implementation-finding**: two of the four readers named in the acceptance criterion still fall back to the legacy file.

---

## ✅ First: the fixture repair is genuine (commit `d5932be`)

Every change in the repair commit redirects a fixture writer from the legacy fixed path to the per-team path and adds the `domain` identity field:

```
- var path = NotifyRoleTopologyStore.ResolvePath(RootPath);
+ var path = NotifyRoleTopologyStore.ResolvePath(RootPath, "intent-cli", "intent-cli-dev");
  File.WriteAllText(path, JsonSerializer.Serialize(new
  {
+     domain = "intent-cli",
      team = "intent-cli-dev",
```

**No assertion was weakened, deleted, skipped, or flipped, and no fallback was re-introduced in product code to keep fixtures working.** Across the seven repaired files the only assertion touched is `NotifyRecordedRolesG588Tests:194`, where the expected value tracks the fixture move (`RelativePath` → `ResolvePath(root, Domain, Team)`); its subject — *the summary names the topology file* — is unchanged, and the new expectation is an absolute path, so it is marginally stronger. The two other edits (`File.Delete` in `SessionLayerMarkerG601Tests` and `NotifyCommandG578Tests.SetMode`) are setup, and both had to move or the tests would no longer construct their intended no-topology condition. Exit codes, delivery outcomes, preflight verdicts and marker findings are untouched. The formerly-failing groups pass **with their original assertions intact**: `SessionLayerMarkerG601Tests`, `NotifyRecordedRolesG588Tests`, `NotifyCommandG578Tests`, `SessionLayerTopologyG604Tests` → 68 passed, 0 failed.

Applying the packet's refinement rule to commit `129f431`, each removed assertion is compatibility-subject **and** matched by a replacement:

| Removed | Kind | Replaced by |
|---|---|---|
| `topology-location-conflict` in `Resolve` | compat behaviour | `Assert.True(current.Resolved)` |
| `compatibility.Resolved == true` + warning naming `topology record` | compat behaviour | `False(Resolved)`, cause `legacy-topology-compatibility-removed`, summary contains the legacy path, **`topology record` and `retire-legacy`** |
| validate exit 1, preflight `Ready == false`, doctor `Contains(location-conflict)` | compat behaviour | validate exit 0 + `valid: true`, `Ready == true`, `DoesNotContain(location-conflict)` |

The identity-mismatch half of the first test survives untouched, and the neither-present half of the second survives untouched. No collateral assertion changed value. I also confirm the fixture usage is not treated as evidence about host migration either way.

---

## 🔴 Blocking implementation-finding: `Validate()` still performs the compatibility read

**Contract clause violated**: Acceptance Criterion — *"Legacy-present-without-record ⇒ fail-closed error naming the file, the removal, and both recovery commands, **in every previously-falling-back reader**"*; In Scope — *"every reader that previously fell back — **topology show/validate, the session-layer preflight**, notify's recipient resolution"*; review-context lens 1 — *"confirm every previously-falling-back reader (topology show, **validate**, **the preflight**, notify recipient resolution) errors while naming the file, the removal, and both commands. A reader that merely stops resolving … leaves a lagging host stuck."*

The PR fixed `NotifyRoleTopologyStore.Resolve`. It did not touch **`NotifyRoleTopologyStore.Validate`** in the same file (~line 367), which keeps its own independent fallback:

```csharp
if (!File.Exists(path) && !string.IsNullOrWhiteSpace(domain) && File.Exists(ResolvePath(routingRoot)))
{
    path = ResolvePath(routingRoot);          // ← still reads the legacy file
}
…
var warnings = … ? new[]
{
    $"Deprecated topology compatibility read from '{path}'; run "
    + $"`intent-cli session-layer topology record --domain {domain} --team {team} ... --write` …"
}
```

That warning is the *same string the PR deleted from `Resolve`* — the removed vocabulary is still live here. Both `session-layer topology validate` (`SessionLayerTopologyCommand.cs:93,126`) and **the session-layer preflight** (`SessionLayerPreflight.cs:275`) route through `Validate`, so a single root cause takes out two of the four named readers.

### Constructed fixture: legacy file present, no per-team record

```
$ intent-cli session-layer topology show --domain intent-cli --team intent-cli-dev --format json
  cause: legacy-topology-compatibility-removed                                    ✅ exit 1
  "Legacy topology file '…/role-pane-mapping.json' was found … the compatibility read has
   been removed. Run `… topology record …`, or run `… topology retire-legacy …`"

$ intent-cli session-layer topology validate --domain intent-cli --team intent-cli-dev --format json
{
  "valid": true,                                                                  ❌ exit 0
  "record_path": ".intent-cli/topology/intent-cli/intent-cli-dev.json",
  "findings": [],
  "summary": "Recorded delivery topology for team 'intent-cli-dev' is valid. Deprecated topology
              compatibility read from '…/role-pane-mapping.json'; run `intent-cli session-layer
              topology record …` to re-record this machine's topology at its per-team local path."
}
```

Note `record_path` names the per-team file **that does not exist on this host**, while the summary admits it read the legacy one — and speaks the vocabulary this slice deleted.

### The inversion, side by side

| host state | `topology validate` | session-layer preflight |
|---|---|---|
| neither file present | `valid: false`, `topology-missing` | `configuration-incomplete`, findings `[marker-not-generated, topology-missing]` |
| **legacy present, no record** | **`valid: true`, no findings** | **`ready`**, findings `[marker-not-generated]` |

**Dropping a stale `role-pane-mapping.json` onto a misconfigured host turns both readers green.** The migration case that should be the loudest is the only one reported as healthy — and the same host simultaneously refuses `topology show` and `notify`, so one host now yields contradictory verdicts depending on which reader you ask. The fail-closed message tells a lagging operator to run `topology record`; the natural next thing they run is `topology validate`, which tells them everything is already valid.

The docs this PR adds now state that *"readers fail closed and name `topology record …` and `topology retire-legacy …`"* — true for two readers, false for the other two, so the guidance overstates the shipped behaviour.

**Requested fix**: give `Validate` the same fail-closed treatment as `Resolve` — when the per-team record is absent and the legacy file exists, emit the `legacy-topology-compatibility-removed` finding (invalid, both commands named) instead of adopting the legacy path, and delete the now-dead `Deprecated topology compatibility read` warning and the unreachable `topology-location-conflict` branch above it. Please add fixture coverage for **validate and the preflight on a legacy-only host**, mirroring `Resolve_LegacyOnlyFailsClosedAndNeitherRemainsConfigurationIncomplete_G627` — that test only exercises `Resolve`, which is why a green suite did not catch this.

---

## Non-blocking observations

- **The remedy presents the two commands as alternatives, but only one runs in that state.** On the legacy-only host the message says *"Run `topology record` …, **or** run `topology retire-legacy` …"*. Live, `retire-legacy` refuses there: `retired: false, conflict: true, "A valid per-team record is required before retiring legacy topology."` I verified the working path is sequential — `topology record` first, then `retire-legacy` succeeds (`"Retired '.intent-cli/role-pane-mapping.json' with named migration evidence."`, file removed). The host is therefore not stuck, and both commands *are* named as the AC requires, so this is not a blocker — but *"…, then run `retire-legacy` …"* would describe what actually works.
- `Equivalent(NotifyTeamTopology, NotifyTeamTopology)` at `NotifyRoleTopology.cs:699` is now dead — its only caller was the deleted location-conflict comparison.
- `Resolve_CopiedNewRecordFailsIdentityAndIgnoresLegacyFile_G627` ends at `Assert.True(current.Resolved)`. The fixture writes the legacy file with `workspace_id: "other"` against a record at `w1`, so it is perfectly set up to prove *which* record won — but as written the assertion would still pass if the fallback were reintroduced, despite the test's name. I confirmed live that the per-team record does win (record `wNEW` + stale legacy `wLEGACY` → resolves `wNEW`, no conflict), so this is a test-strength gap, not a product defect.

---

## Verified and passing

- **Lens 2 — neither-present untouched**: `topology-missing` / `configuration-incomplete` preserved verbatim, in both the live fixture and the retained second half of the G627 unit test.
- **Lens 3 — no automatic migration**: the only writers of a per-team record are `SessionLayerTopologyCommand` (`record` / `retire-legacy`); `NotifyRoleTopology.cs` contains no `File.WriteAllText`/`CreateDirectory`, and `NotifyTaskEnvelopeStore` only reads through `Resolve`. No reader writes on a host's behalf.
- **Lens 4 — ledger cites evidence**: the row moves `retain-through-1.x` → `retire-before-1.0` and cites *"`legacy-topology-retirements.jsonl` entry one (Zero4Racer); all four measured hosts are legacy-free"* — evidence, not a claim that the fleet migrated. JA row equivalent.
- **Lens 5 — `retire-legacy` retained** and demonstrated working end-to-end above.
- **Record-present resolution normal**: per-team record only → resolves `wNEW`, valid.
- ja/en parity under G613 in both `12-agent-message-orchestration.md` and the ledger; JA additionally states readers do not auto-migrate.
- No per-team format change, no delivery/settle change, no `.intent-cli/**` or `intents/**` mutation, no AI provider launched, no label knowledge, no `intent-target`/`intent-pr-created`.
- **Intent**: `means/08-agent-message-orchestration.md` — the removal is justified by measured absence, and node 08's writeback expectation is that *its removal fails closed with the recovery commands named*. That is exactly the clause the finding above blocks.

```
dotnet test IntentSystem.sln -c Release
  → exit 0; Cli.Tests 4496 passed, 1 pre-existing skip, 0 failed; all other assemblies pass
git diff --check 2bb20d3 d5932be → clean (exit 0)
```
CI passes; `Closes #1357` matches `publish.yaml` `created_issue_number: 1357`; base `main` per `direct-main`.

**Tests-pass is necessary but not sufficient**, and this PR is the clearest illustration of it in the series: the suite is fully green, CI is green, the fixture repair is honest — and two of the four readers the acceptance criterion enumerates still read the legacy file, which only a constructed legacy-only host reveals.

tomohisa:
## Re-review verdict: APPROVE — G627 / PR #1358 (Issue #1357)

Head `e9c0ffd`, base `main` @ `2bb20d3`, three commits. Follow-up commit `e9c0ffd` "fix: fail closed on legacy topology validation" touches exactly the two files named in the blocker — `NotifyRoleTopology.cs` and `SessionLayerTopologyG604Tests.cs` — and nothing else.

**The blocker is fixed at the root cause, not patched at the symptom.** `NotifyRoleTopologyStore.Validate` no longer carries its own compatibility read; it now calls the same `LegacyCompatibilityReadRemoved` helper that `Resolve` uses, so the two readers cannot drift apart again:

```csharp
-if (!File.Exists(path) && !string.IsNullOrWhiteSpace(domain) && File.Exists(ResolvePath(routingRoot)))
-{
-    path = ResolvePath(routingRoot);
+if (!File.Exists(path) && !string.IsNullOrWhiteSpace(domain))
+{
+    var legacyPath = ResolvePath(routingRoot);
+    if (File.Exists(legacyPath))
+    {
+        var compatibility = LegacyCompatibilityReadRemoved(legacyPath, domain, team);
+        findings.Add(Finding("<topology>", "file", compatibility.Cause!, compatibility.Summary));
+        return Validation(team, path, findings);
```

The surviving `Deprecated topology compatibility read` warning block and the unreachable `topology-location-conflict` branch are both deleted, and all six `Validation(..., warnings)` call sites were updated in step — a complete removal rather than a suppressed one. The string is now absent repo-wide.

### Live re-verification — all four readers, four constructed hosts

Rebuilt from scratch at `e9c0ffd`; per-team record `wNEW`, legacy file `wLEGACY`.

| host state | `topology show` | `topology validate` | session-layer preflight | notify recipient resolution |
|---|---|---|---|---|
| **legacy only, no record** | exit 1 `legacy-topology-compatibility-removed` | **exit 1, `valid: false`, `legacy-topology-compatibility-removed`** ⟵ was `valid: true` / exit 0 | **`configuration-incomplete` + `legacy-topology-compatibility-removed`** ⟵ was `ready` | exit 1, same cause |
| neither present | `topology-missing` | `topology-missing` | `configuration-incomplete` + `topology-missing` | `topology-missing` |
| record only | exit 0 | `valid: true`, no findings | `ready` | resolves |
| record + stale legacy | exit 0, resolves `wNEW` | `valid: true`, no findings, no conflict | `ready` | resolves |

The inversion I reported is gone: the legacy-only host is now the *loudest* state rather than the only green one, and the four readers agree with each other on the same host. Both new failure messages carry the full required content verbatim:

> Legacy topology file '…/role-pane-mapping.json' was found for team 'intent-cli-dev', but the compatibility read has been removed. Run `intent-cli session-layer topology record …` to declare the current shape, or run `intent-cli session-layer topology retire-legacy … --evidence <evidence> --confirm-retire-legacy --write` to retire the legacy file with evidence.

— the file it found, that the read was removed, and **both** commands, from `validate` and from the preflight alike. `warnings` is now `null` on every path.

The regression surface is clean: the neither-present report is byte-for-byte the prior text, and the record-present rows confirm the new guard (`!File.Exists(path) && !IsNullOrWhiteSpace(domain)`) only consults the legacy path when the per-team record is genuinely absent — with both files present the record wins (`workspace_id: wNEW`, `record_path` per-team, zero findings), so removing the location-conflict branch introduced no new false positive.

The new test `ValidateAndPreflight_LegacyOnlyFailClosed_G627` is precisely the coverage that was missing: it drives `SessionLayerTopologyCommand.ExecuteValidate` to **exit 1** and `SessionLayerPreflight.Analyze` to `ConfigurationIncomplete`, asserting the cause and **both** command names on each. That closes the gap that let a fully green suite hide the defect.

---

## Acceptance criteria

| Criterion | Change | Test |
|---|---|---|
| Compat read removed — legacy file no longer consulted when the record is absent | `Resolve` (`129f431`) and `Validate` (`e9c0ffd`); no remaining `ResolvePath(routingRoot)` fallback assignment | `Resolve_LegacyOnlyFailsClosed…_G627`, `ValidateAndPreflight_LegacyOnlyFailClosed_G627` |
| Fail closed naming file + removal + **both** commands, in **every** previously-falling-back reader | single `LegacyCompatibilityReadRemoved` helper shared by `Resolve` and `Validate`; `topology show`, `validate`, preflight (via `Validate`) and notify (via `Resolve`) all route through it | both tests above assert cause + `topology record` + `retire-legacy`; verified live on all four readers |
| Neither present ⇒ configuration-incomplete unchanged | fall-through preserved | second half of `Resolve_LegacyOnly…_G627`, retained unchanged; live fixture B |
| Ledger row off `retain-through-1.x` citing evidence | row → `retire-before-1.0` citing `legacy-topology-retirements.jsonl` entry one (Zero4Racer) + four measured hosts | `CompatibilityPromiseG612Tests` |
| ja/en guidance: what changed, why, recovery steps | both `12-agent-message-orchestration.md` state the removal and name both commands; JA additionally states readers do not auto-migrate | G613 parity tests |
| No automatic migration | `NotifyRoleTopology.cs` contains no `WriteAllText`/`CreateDirectory`/`Move`/`Copy`/`Delete`; only `SessionLayerTopologyCommand` (record / retire-legacy) writes records | verified live — after all four readers ran against the legacy-only host, the legacy file is unchanged at 225 bytes and no `topology/` directory was created |

**Out-of-scope boundaries held**: `retire-legacy` retained and demonstrated working end-to-end; per-team record format, delivery semantics, recipient identity and the settle contract untouched; no `.intent-cli/**` or `intents/**` mutation; no label or workflow-state changes.

**Related rule**: `means/08-agent-message-orchestration.md` — node 08's writeback expectation is that the compat read's removal *fails closed with the recovery commands named*. That is the clause the previous head missed in two readers and this head satisfies in all four.

---

## Non-blocking observations (not gating)

- **`topology-location-conflict` is now emitted nowhere in `src/`** (it appeared twice at base `2bb20d3`). That is a legitimate consequence of the authorized retirement — the dual-location state is unreachable by design. But ADR `0002-one-dot-zero-compatibility-promise.md` and both 1.0 ledgers still cite it in the present tense as *the* worked example of the coverage rule ("covered **because automation can receive it**"), and `CompatibilityPromiseG612Tests` pins those strings, so the stale example is held in place by a passing test. Worth a follow-up to switch the illustration to a still-emitted cause, or to note the value retired with the compat read. The rule itself is unaffected.
- I withdraw my earlier "or → then" suggestion as a PR finding: the packet's own Goal says a host *"is told to run `topology record` **or** `retire-legacy`"*, so the implementation matches its contract exactly. The sequencing point still holds operationally — `retire-legacy` refuses until a record exists, and the working order is record → retire-legacy — but that is contract wording for a future slice, not something this PR should have deviated from.
- `Equivalent(NotifyTeamTopology, NotifyTeamTopology)` (`NotifyRoleTopology.cs:681`) now has zero callers — dead private code, safe to drop whenever this file is next touched.
- `Resolve_CopiedNewRecordFailsIdentityAndIgnoresLegacyFile_G627` still ends at `Assert.True(current.Resolved)` without asserting *which* record won, despite its name. I confirmed live that the per-team record does win over a stale legacy file, so this is test strength, not behaviour.

---

## Verification

```
dotnet test IntentSystem.sln -c Release
  → exit 0; Cli.Tests 4497 passed (+1, the new test), 1 pre-existing skip, 0 failed
    ConceptIntake 36 · Clarify 47 · Projection 80 · Supervisor 66 · Review 35
    Drift 6 · Workflow 21 · WorkerAdapter 16 · DomainBinding 13 · DogfoodingBridge 10 — all pass
git diff --check 2bb20d3 e9c0ffd → clean
```
CI "Build and test (source contract)" is green at `e9c0ffd`; `Closes #1357` matches `publish.yaml` `created_issue_number: 1357`; base `main` per `direct-main`.

**Tests passing is necessary but not sufficient**, and this PR is the reason to keep saying so: at the previous head the suite was equally green while two of four readers still fell back. The approval rests on the live four-reader × four-host matrix above — reproduced against a freshly built binary on constructed hosts — with the green suite as corroboration rather than as the evidence.

Issue #1353, opened 2026-08-06
G625: Say what you actually know


tomohisa:
# G625: Say what you actually know

## Goal

Fix four places where the CLI's output is inconsistent or asserts more than it knows, all reproduced on this host from a field report: `automation issue-publish` accepts `--execution-unit` like every other step in the flow; `guide orchestrator-thread` without `--domain` says it cannot resolve instead of asserting a default; `--help` is accepted on topology subcommands; `queue-seed-from-packet`'s repair guidance names `packet draft`; and `closeout pr` proposes submodule steps only when the host actually has submodules.

## Why This Slice Exists Now

**Field report from the Zero4Racer team, verified item by item on this host rather than relayed.**

1. **The pointer changes shape at the last boundary.** `packet draft`, `queue-seed-from-packet` and `issue publish-flow` all take `--execution-unit`; only `automation issue-publish` refuses it and demands `--issue <n>`. Reproduced here: `Unknown argument '--execution-unit'`. `publish.yaml` already carries `created_issue_number`, so the unit is resolvable.

2. **The guide asserts a default it did not check.** Without `--domain`, `guide orchestrator-thread` prints *"Session layer: agmsg (PRIMARY) — no selection recorded, so the default is in force"* — while `session-layer show` on the **same host** reports `herdr-only`, `source: recorded`. **Reproduced on this team**, which is herdr-only, so it is not the reporter's misconfiguration. An unresolved lookup is being reported as a determined fact, which invites the reader to diagnose broken configuration that is not broken.

3. **`--help` is rejected as an unknown argument** on topology subcommands (`Unknown or incomplete argument '--help'`). The usage line still prints, so there is no functional loss — but `--help` being "unknown" is a poor first contact with a command.

4. **Repair guidance does not name the repair command.** When the four-file packet contract is incomplete, `queue-seed-from-packet` says to author the missing canonical files without naming `packet draft`, which is exactly the command that authors them.

5. **`closeout pr` proposes submodule sync unconditionally** — reported from a host with no `.gitmodules` and no submodule tree. **Not reproducible here** because this host does have submodules; it stands on the reporter's evidence, and the fix is to condition the guidance on the host's actual layout.

## Design ↔ orchestrator double-check outcome

One slice for these five, because they are consistency and honesty-of-output defects with no semantic risk. The settle-boundary defect from the same report is deliberately **not** here — it touches the G598 delivery contract and has its own slice.

## Current Observed State

[...]

PR #1367, opened 2026-08-06
G632: classify issue targets from declarations

### LINKED ISSUE #1366: G632: A check that reads prose instead of the declaration
# G632: A check that reads prose instead of the declaration

## Goal

Derive `worker issue-preflight`'s target classification from the issue's **declared target** — the packet's target paths and repository — instead of from **substring matches in free prose**. Prose may at most raise an advisory note; it may never make an actionable issue unactionable.

## Why This Slice Exists Now

**Two false blocks in one day, on two different tokens, both stopping a dispatch.**

| unit | issue | prose that triggered it | verdict |
|---|---|---|---|
| G625 | #1353 | "reported from a host with no `.gitmodules` and no `submodules/`" | `issue body references a submodules/ path but --workdir does not point inside a submodules/ tree` |
| G627 | #1357 | naming this machine's host directory as retirement evidence | `issue body references parent-host/…​ execution path which conflicts with child --repo` |

Both bodies carried an explicit `Target paths:` declaration. Both declarations were correct — orchestration said so in both blocked reports before asking design to reword. **The check never looked at either one.**

**What it cost, twice**: implementation stopped, an attention record was opened, design amended a canonical packet, the GitHub issue was re-synced, and a corrected packet was re-reported. Roughly one round trip through three roles for a sentence that was already accurate.

**The worse cost is what the workaround does to the packets.** The only way past the check is to make the prose less precise. G625 lost the literal path of a reported host layout; G627 lost the name of the host that was actually measured — in a slice whose entire argument is *"this was measured rather than assumed."* **A gate that is escaped by removing evidence is teaching the wrong lesson**, and it teaches it to every packet author from now on.

**This is the class G625 shipped a fix for, in the tool that guards G625.** The check asserts a determined conclusion — *this issue targets a submodule* — from an input it never consulted. The declaration was right there.

**Classify from the declaration.** If the declared target paths lie inside a submodule tree, the workdir must be inside that submodule — that is a real mismatch and stays blocking. If the declared target is the child repository, an issue is actionable **no matter what its prose mentions**.

**Prose may warn, never block.** If a mention is worth surfacing, surface it as an advisory note alongside `actionable: true`, following the G618 precedent that an advisory field warns without refusing. A caller that wants to look is told; a caller that is correct is not stopped.

**Do not fix this by excluding code spans.** Teaching the scanner to skip fenced blocks and inline code would have passed G627, whose trigger was plain prose. The basis of the judgment is wrong, not its punctuation handling.

## Current Observed State

- `issue-preflight` returns `classification: target-mismatch`, `actionable: false`, `recommended_action: switch-repo` on the two issues above, with reasons naming the prose token.
- The same command returns `ready-to-implement` after the sentence is reworded and nothing else changes — no target, no repository, no path.
- The declared `Target paths` block and the packet's `target_path` are not consulted in the decision.

- Blocking a genuine submodule/parent mismatch is correct and must be preserved.
- The other preflight classifications and their fail-closed behaviour are correct and unchanged.
- Packet bodies carry an explicit target declaration; that is the input this check should have been using.

## Target Repo / Path / Part

Target part: the issue-preflight target classification and its evidence basis.

## In Scope

- **Declaration-driven classification**: the target verdict is derived from the declared target paths and repository, compared against the worker's workdir and `--repo`.
- **The genuine mismatch stays blocking**: a declared target inside a submodule tree with a workdir outside it remains `target-mismatch`, `actionable: false`.
- **Prose becomes advisory at most**: a mention that does not agree with the declaration produces an advisory note in both human-readable and machine output, with `actionable` unaffected.
- **The reasons name their evidence**: every reason states which input it came from — the declaration or the advisory scan — so a reader can tell a fact from a guess.
- **Both live cases are regression tests**: the two sentences above, verbatim, must classify as actionable when the declaration is the child repository.
- **A missing declaration is not a licence to guess**: if no target declaration can be read, the command says the declaration is missing and names it, rather than falling back to scanning prose.

## Out Of Scope

- Changing any other preflight classification, the fail-closed posture, or `recommended_action` vocabulary for genuine mismatches.
- Changing packet structure or the target-declaration format.
- Excluding code spans or fenced blocks as a substitute for using the declaration.

The child PR derives the worker issue-preflight target classification from the issue's declared target paths and repository compared against the worker's working directory and repository argument, rather than from substring matches in the issue's prose. A declared target inside a submodule tree with a working directory outside that submodule remains a blocking target-mismatch with actionable false, while an issue whose declared target is the child repository is actionable regardless of what its prose mentions. A prose mention that disagrees with the declaration produces an advisory note in both human-readable and machine output without affecting actionability, and every reason states which input it was derived from so that a declaration-based fact is distinguishable from an advisory observation. When no target declaration can be read the command reports that the declaration is missing and names it instead of falling back to scanning prose. The two sentences that caused live false blocks — a report of a host having no gitmodules file and no submodule tree, and a packet naming this machine's host directory as measurement evidence — are added as regression tests asserting an actionable verdict when the declared target is the child repository, and excluding code spans or fenced blocks is not used as a substitute for reading the declaration. All other preflight classifications, the fail-closed posture, the recommended-action vocabulary for genuine mismatches, the packet structure and the target-declaration format are unchanged, EN and JA stay in parity under the G613 terminology policy, and the full suite is green.

## Acceptance Criteria

- Classification derived from the declared target paths and repository versus workdir and `--repo`.
- Genuine mismatch (declared target inside a submodule, workdir outside) still blocks with `actionable: false`.
- Prose disagreement yields an advisory note in human and machine output; `actionable` unchanged.
- Each reason names the input it came from.
- Both live sentences are regression tests returning actionable.
- Missing declaration reported by name; no prose fallback.
- Other classifications, fail-closed posture, packet structure, declaration format unchanged; parity under G613; full suite green.

## Verification

- Re-run against the two recorded bodies verbatim: expect `ready-to-implement` with at most an advisory note.
- Construct an issue whose declared target paths are inside a submodule with a workdir outside: expect the blocking mismatch, unchanged.
- Construct a body with no target declaration: expect a named missing-declaration report, not a prose-derived verdict.
- Diff check: no other classification's behaviour changed.

## Related Links

- Live false blocks, 2026-08-06: issue #1353 (G625) and issue #1357 (G627); both amended by rewording only, with no target, repository or path change.
- G625 — the same class of defect, fixed in the CLI's own output; G618 — the advisory-warning precedent that warns without refusing.

## Knowledge Maintenance

- Intent placement: node 03 primary (verdicts derived from declared state), node 02 supporting (command-surface correctness).
- ADR candidate: none.
- Diagram candidate: none.
- Docs update: ja/en preflight guidance (in scope).
- Closeout writeback expected: yes — node 03 records that a verdict must be derived from the declaration rather than from prose about it, and that a gate escaped by deleting evidence is training authors to write worse packets.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1366

Derive worker issue-preflight target mismatch from declared Repository/Target paths. Prose mismatches are advisory, missing declarations fail closed, and EN/JA recovery guidance plus regression tests are included.

Focused: dotnet test tests/IntentSystem.Cli.Tests/IntentSystem.Cli.Tests.csproj --no-restore --filter FullyQualifiedName~WorkerIssuePreflightCommandTests
Full: dotnet test IntentSystem.sln --no-restore -c Release

### DIFF
--- a/docs/en/07-recovery.md
+++ b/docs/en/07-recovery.md
@@ -123,6 +123,13 @@ intent-cli worker pr-comment-preflight --repo <owner>/<repo> --pr    <n> --forma
 intent-cli automation doctor --format json
 ```
 
+`worker issue-preflight` derives target-mismatch from the issue's declared
+`Repository:` and `Target paths:` fields. A declared child-repository target
+remains actionable even when explanatory prose mentions a host or `submodules/`;
+those mentions are advisory notes only. A declared submodule target still
+blocks when the workdir is outside that submodule, and a missing `Target paths:`
+declaration fails closed instead of guessing from prose.
+
 | Result | Action |
 |--------|--------|
 | `safe_repair_category: child-selector-label-gap` | Apply the one repair `intent-cli` marks safe; retry once |
--- a/docs/ja/07-recovery.md
+++ b/docs/ja/07-recovery.md
@@ -119,6 +119,8 @@ intent-cli worker pr-comment-preflight --repo <owner>/<repo> --pr    <n> --forma
 intent-cli automation doctor --format json
 ```
 
+`worker issue-preflight` の target-mismatch 判定は、本文の推測的な文字列検索ではなく、Issue に宣言された `Repository:` と `Target paths:` を根拠にします。子リポジトリを宣言した Issue は、説明文に host や `submodules/` が現れても actionable のままで、その言及は advisory note として出力されます。宣言された対象が submodule 内で workdir が外側なら従来どおりブロックし、`Target paths:` 宣言がない場合は本文から推測せず fail closed します。
+
 | 結果 | 対応 |
 |------|------|
 | `safe_repair_category: child-selector-label-gap` | `intent-cli` が安全と判断した修復を 1 回適用し、リトライ |
--- a/src/IntentSystem.Cli/Commands/WorkerIssuePreflightAnalyzer.cs
+++ b/src/IntentSystem.Cli/Commands/WorkerIssuePreflightAnalyzer.cs
@@ -138,8 +138,19 @@ internal static class WorkerIssuePreflightAnalyzer
                 WorkerIssuePreflightConstants.RecommendedActions.ReleaseFromTarget);
         }
 
+        var declaredPaths = HostOnlyPacketClassifier.ExtractTargetPaths(body);
+        if (declaredPaths.Count == 0)
+        {
+            return Build(
+                lookup, repo, issueNumber, title, state, labels,
+                WorkerIssuePreflightConstants.Classifications.MissingTargetDeclaration,
+                ["declaration-derived: missing `Target paths:` declaration; cannot classify the issue from prose"],
+                actionable: false,
+                WorkerIssuePreflightConstants.RecommendedActions.WaitForClarification);
+        }
+
         // Step 5: target-mismatch.
-        var mismatchReasons = DetectTargetMismatch(body, repo, workdir);
+        var mismatchReasons = DetectTargetMismatch(body, repo, workdir, declaredPaths);
         if (mismatchReasons.Count > 0)
         {
             return Build(
@@ -154,6 +165,8 @@ internal static class WorkerIssuePreflightAnalyzer
                 actionable: false,
                 WorkerIssuePreflightConstants.RecommendedActions.SwitchRepo);
         }
+
+        var advisories = DetectAdvisories(body, declaredPaths);
 
         // Step 6: contract-incomplete (reuse G183 validator).
         var sourcePath = $"github://{repo}/issues/{issueNumber}";
@@ -198,10 +211,11 @@ internal static class WorkerIssuePreflightAnalyzer
             WorkerIssuePreflightConstants.Classifications.ReadyToImplement,
             Array.Empty<string>(),
             actionable: true,
-            WorkerIssuePreflightConstants.RecommendedActions.Implement);
+            WorkerIssuePreflightConstants.RecommendedActions.Implement,
+            advisories);
     }
 
-    private static IReadOnlyList<string> DetectTargetMismatch(string body, string repo, string workdir)
+    private static IReadOnlyList<string> DetectTargetMismatch(string body, string repo, string workdir, IReadOnlyList<string> declaredPaths)
     {
         var reasons = new List<string>();
         if (string.IsNullOrEmpty(body))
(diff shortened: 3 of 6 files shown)


tomohisa:
## Review blocker — PR #1367 @ ad907a3 (G632)

The classification change itself is **correct**. I constructed the negative cases against a build of this head and every behavioural expectation held:

| constructed case | result |
|---|---|
| declared `Target paths: submodules/intent-system/src/Foo.cs`, `--workdir /tmp/host` | `target-mismatch`, `actionable: false`, reason `declaration-derived: declared target paths are inside submodules/ but --workdir is outside a submodules/ tree` — the guard is preserved |
| G625 sentence verbatim ("reported from a host with no `.gitmodules` and no `submodules/`") + declared child target | `ready-to-implement`, `actionable: true`, advisory only |
| G627 sentence (naming this machine's host directory / parent-host execution path) + declared child target | `ready-to-implement`, `actionable: true`, advisory only |
| no `Target paths:` line | `missing-target-declaration`, `actionable: false`, reason names `Target paths:` — no prose fallback |
| same body with and without advisory-triggering prose | `actionable` byte-for-byte identical |

Full suite green (`IntentSystem.Cli.Tests`: 4505 passed / 1 skipped; the one `IntentSystem.Drift.Tests` failure under `--no-restore` is a missing-`YamlDotNet` restore artifact and passes with restore). `git diff --check` clean. The declaration is now the basis, and code-span exclusion was correctly **not** used.

Three acceptance criteria are nevertheless not met on this head.

### 1. The two live sentences are not regression tests (AC: "Both live sentences are regression tests returning actionable")

Neither recorded sentence appears in the test suite. `Execute_GivenBodyReferencingSubmodulesPath_...` was flipped to assert the advisory, but its body is invented prose (`Additional note: edits land under submodules/intent-system/src/Foo.cs.`), not the G625 sentence. The G627 sentence has **no test at all** — `grep -n "parent-host\|MyIntentHost" tests/IntentSystem.Cli.Tests/WorkerIssuePreflightCommandTests.cs` returns nothing, so the entire parent-host branch (both the removed blocking heuristic and the new advisory) ships untested. review-context.md item 2 is explicit: verify against the sentences, not a paraphrase.

**Unblock**: add both recorded sentences verbatim as regression tests asserting `ready-to-implement` / `actionable: true` with the declaration set to the child repository.

### 2. Not every reason names its input (AC: "Each reason names the input it came from")

The new reasons carry `declaration-derived:`, but the pre-existing `Repository:` mismatch reason in `DetectTargetMismatch` heuristic 1 does not. Observed machine output:

```json
"classification": "target-mismatch",
"reasons": [
  "issue body declares Repository: SomeOther/Repo which does not match --repo J-Tech-Japan/intent-system"
],
"advisories": []
```

That reason *is* declaration-derived — it reads the declared `Repository:` — but a reader cannot tell it apart from a prose guess.

**Unblock**: prefix it (and any other surviving reason on this path) with its evidence basis.

### 3. Advisories are dropped on every non-ready classification (AC: "Prose disagreement yields an advisory note in human and machine output")

`DetectAdvisories` is called once and passed to `Build` only on the `ready-to-implement` return; every other return uses the `advisories = null` default. The body in the block above mentions both `submodules/` and parent-host and still reports `"advisories": []`.

Consequence: the evidence review-context.md asks for — "machine output showing a declaration-derived reason and an advisory note side by side" — cannot be produced by this head at all, because the two fields are mutually exclusive by construction.

**Unblock**: compute advisories once before the classification switch and attach them to every result, so a blocked verdict still surfaces the prose observation alongside its declaration-derived reason.

### What was NOT verified / must not be claimed

No real GitHub issue was re-run through `worker issue-preflight` (issues #1353 and #1357 have already been reworded, so their current bodies no longer carry the trigger sentences). All evidence above comes from constructed bodies executed against a local build of `ad907a3`; no runtime or operator verification beyond that was performed.

tomohisa:
Repair pushed at `6a438e2`. Addressed all three blocker findings: added verbatim G625/G627 sentence regression tests, labeled Repository mismatch reasons as declaration-derived, and preserved prose advisories on every classification (including blocked cases). Focused tests: 29 passed. Full Release suite: passed. Exact-head CI: run `31077107221` passed.

tomohisa:
## Re-review — PR #1367 @ 6a438e2: approved

All three findings from https://github.com/J-Tech-Japan/intent-system/pull/1367#issuecomment-5201106420 are repaired, verified by re-running the constructed negative cases against a local build of `6a438e2`.

### Finding 1 — both live sentences are now regression tests

`Execute_GivenG625SentenceWithChildDeclaration_IsActionableWithAdvisory` carries the G625 sentence verbatim ("reported from a host with no `.gitmodules` and no `submodules/`") and `Execute_GivenG627SentenceWithChildDeclaration_IsActionableWithAdvisory` carries the G627 sentence ("naming this machine's host directory as retirement evidence"). Both assert `ready-to-implement` / `actionable: true` with an advisory, and the parent-host branch is no longer untested. Independently reproduced: both sentences classify actionable with the declaration set to the child repository.

### Finding 2 — every reason now names its input

The `Repository:` mismatch reason is now `declaration-derived: declared Repository: SomeOther/Repo which does not match --repo J-Tech-Japan/intent-system`. Constructed cases confirm every reason on the target path starts with `declaration-derived:` and every advisory with `advisory-derived:`.

### Finding 3 — advisories survive every classification

`Build` now derives advisories from the issue body when the caller passes none, so a blocked verdict carries its prose observation. The evidence review-context.md asked for is now producible in both surfaces:

```
## Reasons
- declaration-derived: declared Repository: SomeOther/Repo which does not match --repo J-Tech-Japan/intent-system

## Advisories
- advisory-derived: prose mentions a submodules/ path, but the declared target is the child repository
- advisory-derived: prose mentions parent-host/MyIntentHost, but the declared target is the child repository
```

### Acceptance criteria at this head

| criterion | evidence |
|---|---|
| Classification derived from declared paths + repository vs workdir / `--repo` | `DetectTargetMismatch` consumes `HostOnlyPacketClassifier.ExtractTargetPaths`; the `body.Contains("submodules/")` and `parent-host`/`MyIntentHost` blocking heuristics are gone |
| Genuine mismatch still blocks | declared `submodules/intent-system/src/Foo.cs` + `--workdir /tmp/host` → `target-mismatch`, `actionable: false` |
| Prose disagreement is advisory in human and machine output; `actionable` unchanged | block above; identical `actionable` / `classification` / `recommended_action` with and without the prose |
| Each reason names its input | all reasons `declaration-derived:`, all advisories `advisory-derived:` |
| Both live sentences are regression tests | the two new tests, reproduced independently |
| Missing declaration named, no prose fallback | `missing-target-declaration`, reason names `` `Target paths:` ``, `recommended_action: wait-for-clarification` |
| Other classifications / fail-closed posture / vocabulary / packet structure / declaration format unchanged | no classification constant removed; `wait-for-clarification` already existed on `main`; no packet or declaration-format change; code-span exclusion correctly not used |
| EN/JA parity | `docs/en/07-recovery.md` and `docs/ja/07-recovery.md` both carry the same guidance under G613 terminology |

Also checked the real packet declaration shape (all globs inside one backtick span, as in G632's own body): `Target paths: `src/IntentSystem.Cli/**, docs/en/**, docs/ja/**, tests/**`` parses correctly and classifies actionable.

### Verification run

- `dotnet test IntentSystem.sln -c Release` — fully green: `IntentSystem.Cli.Tests` 4507 passed / 1 skipped, all other projects 0 failed. (The single `IntentSystem.Drift.Tests` failure I saw in the prior pass was a `--no-restore` `YamlDotNet` artifact; it passes here.)
- `git diff --check origin/main...6a438e2` — clean.
- Intent reference: `intents/intent-cli/intent-tree/means/03-state-and-audit-strategy.md` (node 03 — a verdict is derived from declared state, not from prose about it); consistent with the G618 advisory-warns-without-refusing precedent and the G625 same-class fix.
- `Closes #1366` matches the linked source issue (G311).

Tests passing is treated as necessary, not sufficient — the criteria above are each mapped to a change or a reproduced observation.

### What was NOT verified

No real GitHub issue was re-run through `worker issue-preflight`: #1353 and #1357 have already been reworded, so their current bodies no longer carry the trigger sentences. All evidence comes from constructed bodies executed against a local build of `6a438e2`. No runtime or operator verification beyond that was performed, and none is claimed.

PR #1368, opened 2026-08-06
G631: pin encoding for every child process stream

### LINKED ISSUE #1365: G631: The tenth spawn site
# G631: The tenth spawn site

## Goal

Decode **every** redirected child-process stream as UTF-8, not only the `gh` ones — fixing the nine sites that still inherit the console code page, starting with `NotifyTransport` where the defect stops the entire orchestration loop — and add a **guard that fails when a new spawn site omits it**, so the class cannot return a tenth time.

## Why This Slice Exists Now

**A reported total-loop failure on Japanese Windows, with the root cause identified in code and confirmed here.**

On Windows 11 (Japanese, console cp932) with herdr-only, `notify delegate` failed permanently — ten retries, identical error:

```
herdr agent list returned invalid JSON: 't' is invalid after a value.
LineNumber: 0 | BytePositionInLine: 493.
```

`notify report` and `escalate` failed at the same moment, so **every transport-carrying operation stopped and the loop halted** on a real slice. The raw `herdr agent list` output was valid UTF-8 JSON; byte position 493 was where a Japanese string began. .NET decoded the child's stdout with the console OEM code page because no encoding was declared.

**The trigger was ambient, not authored**: a pane titled 「レビュー GitHub プルリクエスト」 appeared in another workspace. Closing that pane restored service; reopening it broke service again. A Japanese username in a path had been mis-decoding harmlessly for a long time because those bytes corrupted characters without breaking JSON structure. **Onset is byte-sequence dependent, so this fails at an unpredictable moment rather than on first use.**

**Verified independently on this host against `main` at 06f1a71:**

- `src/IntentSystem.Cli/Commands/NotifyTransport.cs:18` builds `ProcessStartInfo` with `RedirectStandardOutput`/`RedirectStandardError` and **declares no encoding**.
- **Nine sites redirect output with no declared encoding**: `Commands/NotifyTransport.cs`, `GitCommandRunner.cs`, `IntentSystem.Review/GitCommandRunner.cs`, `Commands/AutomationDurableStatePreflightCommand.cs`, `Commands/AutomationHostSyncPreflightCommand.cs`, `Commands/AutomationInstalledCliSurfaceProbe.cs`, `Commands/AutomationSameRepoMetadataPreflightCommand.cs`, `Commands/AutomationWorkspaceGuardCommand.cs`, `Commands/IntentHostCheckCommand.cs`.
- **Fourteen sites do declare it** — and every one of them calls `gh`.

**That split is the whole story.** G484 (`ebeaa20`) fixed this class for `gh` and named the helper `GitHubCliProcessEncoding`. The fix was attached to a *callee* rather than to the *act of redirecting a stream*, so every spawn site added afterwards — `herdr` transport among them — started life with the bug. **Fixing nine sites by hand repeats the same mistake at a larger radius**: the tenth site will be written next month by someone who has never read this issue.

**Fix the sites and add the guard in one slice**, because the sites alone are a patch and the guard alone leaves the outage in place. **Rename the helper to a callee-neutral name**: the `GitHubCli` prefix is not cosmetic — it is the reason a developer wiring up `herdr` had no reason to think it applied to them.

**A macOS reviewer cannot reproduce this and must not claim to.** The failure needs Windows plus a non-UTF-8 console code page plus non-ASCII output; on macOS and Linux .NET already defaults to UTF-8, so the runtime failure is structurally unreachable. Verification is therefore by **construction** (the declared encoding on each site) and by **guard** (a new omitting site fails the suite) — stated plainly rather than dressed up as a reproduction.

## Current Observed State

- Nine redirected-stream sites decode with the ambient console code page; on a non-UTF-8 console any non-ASCII byte in child output can corrupt the read.
- `NotifyTransport` is the highest-severity of the nine: the transport is a single point of failure for the whole four-thread loop.
- Nothing prevents the next spawn site from repeating the omission.

- G484's fix and the UTF-8-no-BOM encoding value are correct; this slice generalises their reach, it does not revisit the decision.
- The reporter's root-cause analysis is correct and was re-verified here against `main`; the code facts above are measured, not quoted.
- herdr's own output is valid UTF-8; nothing about herdr needs to change.

## Target Repo / Path / Part

Target paths: `src/IntentSystem.Cli/**, src/IntentSystem.Review/**, docs/en/**, docs/ja/**, tests/**`

Target part: child-process stream decoding across all spawn sites, and the guard that keeps it universal.

## In Scope

- **All nine sites declare UTF-8 for both redirected streams**, `NotifyTransport` first among them.
- **The helper is renamed to a callee-neutral name** so it reads as "how this codebase decodes child output" rather than "how this codebase talks to `gh`"; the encoding value itself is unchanged.
- **A guard that fails the suite when a spawn site redirects a stream without declaring the encoding.** It must be demonstrated to fail: the PR shows the guard rejecting a deliberately omitting site, not merely passing on the fixed tree.
- **The severity recorded where an operator will meet it**: transport failure stops every role, and onset depends on ambient bytes — a pane title in an unrelated workspace — so an intermittent, environment-triggered total stall on Windows is the symptom to recognise.
- **ja/en note** for Windows users: no non-ASCII avoidance workaround is required after this fix.

## Out Of Scope

- Changing the encoding value, `gh` behaviour, transport semantics, delivery semantics, or anything herdr-side.
- Console output encoding of intent-cli's own process (a different concern from decoding a child's stream).
- Broad refactoring of the spawn sites beyond declaring the encoding and adopting the renamed helper.

The child PR declares UTF-8 encoding for both the redirected standard output and standard error of every child-process spawn site that currently omits it, namely Commands/NotifyTransport.cs, GitCommandRunner.cs in the CLI and in the Review project, and the AutomationDurableStatePreflightCommand, AutomationHostSyncPreflightCommand, AutomationInstalledCliSurfaceProbe, AutomationSameRepoMetadataPreflightCommand, AutomationWorkspaceGuardCommand and IntentHostCheckCommand commands, using the same UTF-8 without byte-order-mark value the gh sites already use. The helper carrying that value is renamed to a callee-neutral name so that it reads as the codebase's rule for decoding child output rather than as a GitHub CLI concern, with all existing call sites updated and the value unchanged. A guard is added that fails the test suite when any spawn site redirects a stream without declaring the encoding, and the pull request demonstrates that guard failing against a deliberately omitting site rather than only passing against the fixed tree. EN and JA guidance records that a transport decoding failure stops every role at once and that its onset depends on ambient non-ASCII bytes such as a pane title in an unrelated workspace, so that an operator can recognise an intermittent total stall on Windows, and states that no non-ASCII avoidance workaround is needed after this fix. The encoding value, gh behaviour, transport semantics and delivery semantics are unchanged, no herdr-side change is made, the process's own console output encoding is untouched, no spawn site is refactored beyond declaring the encoding and adopting the renamed helper, EN and JA stay in parity under the G613 terminology policy, and the full suite is green.

## Acceptance Criteria

- All nine named sites declare UTF-8 for stdout and stderr; no redirected-stream site remains without a declared encoding.
- Helper renamed to a callee-neutral name, value unchanged, all call sites updated.
- Guard fails the suite on an omitting site, **demonstrated failing** in the PR, and passes on the fixed tree.
- EN/JA record the operator-facing symptom (total, intermittent, environment-triggered stall on Windows) and withdraw the avoidance workaround.
- Encoding value, `gh` behaviour, transport/delivery semantics, herdr, and process console output unchanged; no unrelated refactoring.
- ja/en parity under G613; full suite green.

## Verification

- Enumerate every `ProcessStartInfo` with a redirected stream and assert each declares the encoding — the same enumeration that produced the nine, re-run to return zero.
- Add a temporary omitting site and assert the guard fails; remove it and assert the suite passes.
- Assert by diff that the encoding value and the `gh` call sites' behaviour are unchanged.
- **State explicitly that the Windows runtime failure was not reproduced** if the reviewing host is macOS or Linux, where it is structurally unreachable; verification there is by construction and by guard.

## Related Links

- Bug report, J-Tech Japan estivo, 2026-08-06: Japanese Windows 11, cp932 console, herdr-only; `notify delegate` permanently failing at byte position 493; pane title 「レビュー GitHub プルリクエスト」 as trigger; recovery by closing that pane.
- G484 (`ebeaa20`) — the same class fixed for `gh`, and the naming that limited its reach.
- Verification on this host against `main` 06f1a71: nine sites without a declared encoding, fourteen with — all fourteen calling `gh`.

## Knowledge Maintenance

- Intent placement: node 02 primary (tooling correctness across platforms), node 08 supporting (transport as a single point of failure).
- ADR candidate: none.
- Diagram candidate: none.
- Docs update: ja/en Windows note (in scope).
- Closeout writeback expected: yes — node 02 records that a fix attached to a callee does not travel to the next callee, that a guard is what makes a class fix a class fix, and that a defect unreachable on the reviewer's platform must be verified by construction and said so plainly.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1365

Pin UTF-8/no-BOM decoding on every redirected child-process stream, rename the helper to a callee-neutral ProcessOutputEncoding, and add a source guard that catches omissions. EN/JA recovery guidance records the intermittent total Windows stall and the no-workaround outcome.

Focused guard: dotnet test tests/IntentSystem.Cli.Tests/IntentSystem.Cli.Tests.csproj --no-restore --filter FullyQualifiedName~ChildProcessEncodingGuardTests
Full: dotnet test IntentSystem.sln --no-restore -c Release

The Windows runtime failure was not reproduced on this macOS host; verification is by construction and the guard.

### DIFF
--- a/docs/en/07-recovery.md
+++ b/docs/en/07-recovery.md
@@ -130,6 +130,14 @@ those mentions are advisory notes only. A declared submodule target still
 blocks when the workdir is outside that submodule, and a missing `Target paths:`
 declaration fails closed instead of guessing from prose.
 
+On Japanese Windows, a redirected child-process decoding failure can appear as
+an intermittent, environment-triggered total stall: ambient non-ASCII bytes
+(for example, a pane title in an unrelated workspace) can stop every transport
+operation at once. Child streams are now pinned to UTF-8, so no workaround that
+avoids non-ASCII text is required after upgrading. On macOS/Linux the Windows
+code-page failure is structurally unreachable; verification is by construction
+and by the source guard.
+
 | Result | Action |
 |--------|--------|
 | `safe_repair_category: child-selector-label-gap` | Apply the one repair `intent-cli` marks safe; retry once |
--- a/docs/ja/07-recovery.md
+++ b/docs/ja/07-recovery.md
@@ -121,6 +121,8 @@ intent-cli automation doctor --format json
 
 `worker issue-preflight` の target-mismatch 判定は、本文の推測的な文字列検索ではなく、Issue に宣言された `Repository:` と `Target paths:` を根拠にします。子リポジトリを宣言した Issue は、説明文に host や `submodules/` が現れても actionable のままで、その言及は advisory note として出力されます。宣言された対象が submodule 内で workdir が外側なら従来どおりブロックし、`Target paths:` 宣言がない場合は本文から推測せず fail closed します。
 
+日本語 Windows では、子プロセスのリダイレクト出力のデコード不良が、間欠的・環境依存の「全 transport が同時に止まる」ストールとして現れることがあります。別 workspace の pane title など、周囲の非 ASCII バイトがきっかけになります。子プロセスの stream は UTF-8 に固定されたため、更新後に非 ASCII を避ける回避策は不要です。macOS/Linux では Windows の code page 起因の失敗は構造上発生せず、構築結果と source guard で検証します。
+
 | 結果 | 対応 |
 |------|------|
 | `safe_repair_category: child-selector-label-gap` | `intent-cli` が安全と判断した修復を 1 回適用し、リトライ |
--- a/src/IntentSystem.Cli/Commands/AutomationDurableStatePreflightCommand.cs
+++ b/src/IntentSystem.Cli/Commands/AutomationDurableStatePreflightCommand.cs
@@ -675,6 +675,8 @@ internal static class AutomationDurableStatePreflightCommand
         process.StartInfo.WorkingDirectory = workingDirectory;
         process.StartInfo.RedirectStandardOutput = true;
         process.StartInfo.RedirectStandardError = true;
+        process.StartInfo.StandardOutputEncoding = ProcessOutputEncoding.Utf8NoBom;
+        process.StartInfo.StandardErrorEncoding = ProcessOutputEncoding.Utf8NoBom;
         process.StartInfo.UseShellExecute = false;
         process.StartInfo.CreateNoWindow = true;
         process.Start();
--- a/src/IntentSystem.Cli/Commands/AutomationHostSyncPreflightCommand.cs
+++ b/src/IntentSystem.Cli/Commands/AutomationHostSyncPreflightCommand.cs
@@ -332,6 +332,8 @@ internal static class AutomationHostSyncPreflightCommand
         process.StartInfo.WorkingDirectory = workingDirectory;
         process.StartInfo.RedirectStandardOutput = true;
         process.StartInfo.RedirectStandardError = true;
+        process.StartInfo.StandardOutputEncoding = ProcessOutputEncoding.Utf8NoBom;
+        process.StartInfo.StandardErrorEncoding = ProcessOutputEncoding.Utf8NoBom;
         process.StartInfo.UseShellExecute = false;
         process.StartInfo.CreateNoWindow = true;
         process.Start();
--- a/src/IntentSystem.Cli/Commands/AutomationInstalledCliSurfaceProbe.cs
+++ b/src/IntentSystem.Cli/Commands/AutomationInstalledCliSurfaceProbe.cs
@@ -154,6 +154,8 @@ internal static class AutomationInstalledCliSurfaceProbe
             UseShellExecute = false,
             RedirectStandardOutput = true,
             RedirectStandardError = true,
+            StandardOutputEncoding = ProcessOutputEncoding.Utf8NoBom,
+            StandardErrorEncoding = ProcessOutputEncoding.Utf8NoBom,
         };
         startInfo.Environment["INTENT_CLI_SURFACE_PROBE"] = "1";
 
--- a/src/IntentSystem.Cli/Commands/AutomationIssueRetireCommand.cs
+++ b/src/IntentSystem.Cli/Commands/AutomationIssueRetireCommand.cs
@@ -942,8 +942,8 @@ internal sealed class GhCliGitHubIssueRetirementMutator : IGitHubIssueRetirement
         var startInfo = new ProcessStartInfo
         {
             FileName = "gh",
-            StandardOutputEncoding = GitHubCliProcessEncoding.Utf8NoBom,
-            StandardErrorEncoding = GitHubCliProcessEncoding.Utf8NoBom,
+            StandardOutputEncoding = ProcessOutputEncoding.Utf8NoBom,
+            StandardErrorEncoding = ProcessOutputEncoding.Utf8NoBom,
             RedirectStandardOutput = true,
             RedirectStandardError = true,
             UseShellExecute = false,
--- a/src/IntentSystem.Cli/Commands/AutomationPublishLifecycleRepairCommand.cs
+++ b/src/IntentSystem.Cli/Commands/AutomationPublishLifecycleRepairCommand.cs
@@ -406,8 +406,8 @@ internal sealed class GhCliGitHubAutomationIssueLookup : IGitHubAutomationIssueL
             FileName = "gh",
             // G484: decode gh stdout/stderr as UTF-8 regardless of the ambient
             // console code page (Windows cp932) so Japanese payloads stay valid.
-            StandardOutputEncoding = GitHubCliProcessEncoding.Utf8NoBom,
-            StandardErrorEncoding = GitHubCliProcessEncoding.Utf8NoBom,
+            StandardOutputEncoding = ProcessOutputEncoding.Utf8NoBom,
+            StandardErrorEncoding = ProcessOutputEncoding.Utf8NoBom,
             RedirectStandardOutput = true,
             RedirectStandardError = true,
             UseShellExecute = false,
--- a/src/IntentSystem.Cli/Commands/AutomationSameRepoMetadataPreflightCommand.cs
+++ b/src/IntentSystem.Cli/Commands/AutomationSameRepoMetadataPreflightCommand.cs
@@ -308,6 +308,8 @@ internal static class AutomationSameRepoMetadataPreflightCommand
         process.StartInfo.WorkingDirectory = repoRoot;
         process.StartInfo.RedirectStandardOutput = true;
         process.StartInfo.RedirectStandardError = true;
+        process.StartInfo.StandardOutputEncoding = ProcessOutputEncoding.Utf8NoBom;
+        process.StartInfo.StandardErrorEncoding = ProcessOutputEncoding.Utf8NoBom;
         process.StartInfo.UseShellExecute = false;
         process.StartInfo.CreateNoWindow = true;
         try
--- a/src/IntentSystem.Cli/Commands/AutomationWorkspaceGuardCommand.cs
+++ b/src/IntentSystem.Cli/Commands/AutomationWorkspaceGuardCommand.cs
@@ -843,6 +843,8 @@ internal sealed class ShellGitRunner : IGitRunner
         process.StartInfo.WorkingDirectory = workingDirectory;
         process.StartInfo.RedirectStandardOutput = true;
         process.StartInfo.RedirectStandardError = true;
+        process.StartInfo.StandardOutputEncoding = ProcessOutputEncoding.Utf8NoBom;
+        process.StartInfo.StandardErrorEncoding = ProcessOutputEncoding.Utf8NoBom;
         process.StartInfo.UseShellExecute = false;
         process.StartInfo.CreateNoWindow = true;
         process.Start();
--- a/src/IntentSystem.Cli/Commands/GhIssueCreator.cs
+++ b/src/IntentSystem.Cli/Commands/GhIssueCreator.cs
@@ -24,8 +24,8 @@ internal sealed class GhIssueCreator : IGhIssueCreator
             UseShellExecute = false,
             // G484: decode gh stdout/stderr as UTF-8 regardless of the ambient
             // console code page (Windows cp932) so Japanese payloads stay valid.
-            StandardOutputEncoding = GitHubCliProcessEncoding.Utf8NoBom,
-            StandardErrorEncoding = GitHubCliProcessEncoding.Utf8NoBom
+            StandardOutputEncoding = ProcessOutputEncoding.Utf8NoBom,
+            StandardErrorEncoding = ProcessOutputEncoding.Utf8NoBom
         };
 
         startInfo.ArgumentList.Add("issue");
--- a/src/IntentSystem.Cli/Commands/IGitHubAutomationCandidateLister.cs
+++ b/src/IntentSystem.Cli/Commands/IGitHubAutomationCandidateLister.cs
@@ -312,8 +312,8 @@ internal sealed class GhCliGitHubAutomationCandidateLister : IGitHubAutomationCa
             FileName = "gh",
             // G484: decode gh stdout/stderr as UTF-8 regardless of the ambient
             // console code page (Windows cp932) so Japanese payloads stay valid.
-            StandardOutputEncoding = GitHubCliProcessEncoding.Utf8NoBom,
-            StandardErrorEncoding = GitHubCliProcessEncoding.Utf8NoBom,
+            StandardOutputEncoding = ProcessOutputEncoding.Utf8NoBom,
+            StandardErrorEncoding = ProcessOutputEncoding.Utf8NoBom,
             RedirectStandardOutput = true,
             RedirectStandardError = true,
             UseShellExecute = false,
--- a/src/IntentSystem.Cli/Commands/IGitHubIssueClosingPrLookup.cs
+++ b/src/IntentSystem.Cli/Commands/IGitHubIssueClosingPrLookup.cs
@@ -57,8 +57,8 @@ internal sealed class GhCliGitHubIssueClosingPrLookup : IGitHubIssueClosingPrLoo
             FileName = "gh",
             // G484: decode gh stdout/stderr as UTF-8 regardless of the ambient
             // console code page (Windows cp932) so Japanese payloads stay valid.
-            StandardOutputEncoding = GitHubCliProcessEncoding.Utf8NoBom,
-            StandardErrorEncoding = GitHubCliProcessEncoding.Utf8NoBom,
+            StandardOutputEncoding = ProcessOutputEncoding.Utf8NoBom,
+            StandardErrorEncoding = ProcessOutputEncoding.Utf8NoBom,
             RedirectStandardOutput = true,
             RedirectStandardError = true,
             UseShellExecute = false,
--- a/src/IntentSystem.Cli/Commands/IGitHubIssueLookup.cs
+++ b/src/IntentSystem.Cli/Commands/IGitHubIssueLookup.cs
@@ -43,8 +43,8 @@ internal sealed class GhCliGitHubIssueLookup : IGitHubIssueLookup
             FileName = "gh",
             // G484: decode gh stdout/stderr as UTF-8 regardless of the ambient
             // console code page (Windows cp932) so Japanese payloads stay valid.
-            StandardOutputEncoding = GitHubCliProcessEncoding.Utf8NoBom,
-            StandardErrorEncoding = GitHubCliProcessEncoding.Utf8NoBom,
+            StandardOutputEncoding = ProcessOutputEncoding.Utf8NoBom,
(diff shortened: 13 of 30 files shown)


tomohisa:
Implemented G631 at `cabeb3e`. All redirected child streams in CLI and Review now declare UTF-8/no-BOM; helper renamed to `ProcessOutputEncoding`; source guard catches omissions and was demonstrated failing (exit 1) against a temporary omission before restoration. Focused guard passed; full Release suite passed. Exact-head CI run `31077837101` passed. Windows cp932 runtime failure was not reproduced on macOS; verification is by construction and guard.

tomohisa:
## Review blocker — PR #1368 @ cabeb3e (G631)

**Platform statement first**: this review ran on **macOS** (Darwin, .NET 11 preview SDK, net10.0 targets). The Windows cp932 runtime failure is structurally unreachable here and **was not reproduced, and is not claimed to have been**. Everything below is verified by construction (source enumeration and diff) and by exercising the guard.

### What holds

**Enumeration returns zero.** Re-ran the enumeration that produced the nine — every `ProcessStartInfo` with a redirected stream anywhere under `src/` — and every one of the 25 files now declares both `StandardOutputEncoding` and `StandardErrorEncoding`. No redirected-stream site remains without a declared encoding, including `src/IntentSystem.Review/GhReviewCommandRunner.cs`, which was in neither the nine nor the fourteen.

**Encoding value and `gh` behaviour unchanged.** `git diff origin/main...cabeb3e` on the fourteen `gh` sites is a pure identifier substitution `GitHubCliProcessEncoding.Utf8NoBom` → `ProcessOutputEncoding.Utf8NoBom`; the `UTF8Encoding(encoderShouldEmitUTF8Identifier: false)` value is byte-identical and the rename is recorded by git as a 91%-similarity rename. No spawn site is refactored beyond declaring the encoding and adopting the helper. Transport/delivery semantics, herdr, and the process's own console encoding are untouched.

**The guard does fail on an omitting site.** I did not take the comment's word for it — I added one and ran it:

```
Failed IntentSystem.Cli.Tests.ChildProcessEncodingGuardTests.EveryRedirectedChildStreamDeclaresUtf8Encoding
  Assert.Empty() Failure: Collection was not empty
  Collection: ["src/IntentSystem.Cli/ZzTempOmittingSpawnSite.cs:12", "src/IntentSystem.Cli/ZzTempOmittingSpawnSite.cs:13"]
```

Removed, and the guard plus the full suite pass: `dotnet test IntentSystem.sln -c Release` green across all eleven test projects (`IntentSystem.Cli.Tests` 4508 passed / 1 skipped). `git diff --check` clean. EN and JA both record the symptom as intermittent, environment-triggered, and total, cite the unrelated-workspace pane title, and withdraw the non-ASCII avoidance workaround, in parity under G613. `Closes #1365` matches the linked source issue.

### The blocker: the guard is not universal, and I have it green with two live omissions

The packet's thesis is that G484 failed **because the fix was attached to a subset** — a callee — so it did not travel to the next spawn site. The guard as written repeats that shape at directory granularity, and the proximity heuristic adds a second hole. Both are demonstrated, not reasoned about.

#### 1. Nine of the eleven `src/` projects are unscanned

The guard filters to `src/IntentSystem.Cli/` and `src/IntentSystem.Review/`. `src/IntentSystem.Supervisor`, `Projection`, `Clarify`, `ConceptIntake`, `Drift`, `Workflow`, `WorkerAdapter`, `DogfoodingBridge`, `DomainBinding` are invisible to it. Today none of them spawn a process — which is exactly the condition under which the tenth site gets written there.

Demonstration: I placed a `ProcessStartInfo` with both streams redirected and no declared encoding at `src/IntentSystem.Supervisor/ZzTempOmittingSpawnSite.cs`. The guard passed green.

**Unblock**: scan every `*.cs` under `src/` (or the whole repo minus `bin`/`obj`), not an allow-list of two projects. The set of projects that may spawn a child is not knowable in advance — that is the premise of the slice.

#### 2. An omitting site within ±12 lines of a compliant one is masked

The guard accepts a redirect line if the ±12-line window contains the two encoding identifiers *anywhere* — it does not associate them with the same `ProcessStartInfo`. Two initializers in one file therefore cover for each other.

Demonstration: a single file with a compliant `ProcessStartInfo` followed immediately by an omitting one — the guard passed green with the omission present. This is not hypothetical: `IGitHubLabelMutator.cs` (lines 641/721) and `IssuePublishFlowCommand.cs` (lines 1638/1738) each already hold two spawn sites, and a third added between them would be masked.

Both probes were live at once and the guard still reported success — a guard that is green while two omitting spawn sites exist in the tree does not close the class.

**Unblock**: bind the check to the initializer rather than to a line window — e.g. parse each `new ProcessStartInfo { ... }` block (or use Roslyn on the compilation) and require that the *same* initializer that sets a `RedirectStandard*` also sets both encodings. Optionally assert the value is the shared helper rather than any expression, so the guard also catches a site that pins the wrong encoding.

### Included with the above

#### 3. The renamed helper still documents itself as a `gh` concern — and the rename broke the sentence

review-context.md item 3: the naming is the mechanism by which the class recurred. The type name is neutral now, but its XML doc is not, and only the first line was edited:

```
/// G631: shared decoding contract for redirected child-process output.
/// always emits UTF-8 (JSON and error text), but when a redirected
```

"GitHub CLI" was deleted from the head of that sentence, leaving `/// always emits UTF-8` with no subject. The body still reads "Every `gh` invocation pins both stream encodings…" and the field doc still says "for every `gh` subprocess". A developer wiring up the next transport reads that and concludes it does not apply to them — the precise failure this slice exists to end.

**Unblock**: rewrite the doc to state the codebase rule (every redirected child stream is decoded as UTF-8/no-BOM) with `gh` and cp932 as the motivating example, not the scope.

#### 4. `GhReviewCommandRunner` still carries its own private copy instead of adopting the helper

`src/IntentSystem.Review/GhReviewCommandRunner.cs:10` keeps `private static readonly Encoding Utf8NoBom = …` while the new `src/IntentSystem.Review/ProcessOutputEncoding.cs` defines the identical value beside it. The AC is "value unchanged, **all call sites updated**"; this call site was not, so the Review project now has two definitions of the same constant and one of them is still attached to a `gh` callee. The new Review helper also carries none of the explanatory doc, so the rule is documented in one project only.

**Unblock**: point `GhReviewCommandRunner` at `ProcessOutputEncoding.Utf8NoBom` and delete the private field.

### Note on evidence, not a blocker

The PR records the guard demonstration as a prose claim ("demonstrated failing (exit 1) against a temporary omission before restoration") with no output. review-context.md asks the PR to *show* it. I reproduced it myself and the claim is accurate, so I am not raising it as a separate finding — but pasting the failing output into the PR would make the demonstration self-evidencing.

### What was NOT verified and must not be claimed

The Windows/cp932/non-ASCII runtime failure was **not** reproduced and no Windows host was exercised. No `notify delegate` / `herdr agent list` round trip was run against a non-UTF-8 console. Verification is by source enumeration, by diff, and by the guard, on macOS only.

tomohisa:
Repair pushed at `8380c9d`. The guard now scans all `src/**/*.cs` projects and binds checks to each ProcessStartInfo initializer (plus assignment-style StartInfo redirects), so same-file omissions are detected. Rewrote ProcessOutputEncoding docs as a universal child-stream rule and updated GhReviewCommandRunner to use the shared helper, removing its duplicate encoding. Temporary same-file omission failed the guard with exit 1; restored guard passes. Full Release suite passed. Exact-head CI run `31078434250` passed.

tomohisa:
## Re-review — PR #1368 @ 8380c9d (G631)

**Platform**: macOS again. The Windows cp932 runtime failure was **not reproduced** and is not claimed. Verification is by source enumeration, by diff, and by exercising the guard.

### Both prior blockers are fixed — verified by re-running the same probes

**Blocker 1 (nine of eleven projects unscanned) — closed.** The filter is gone; the guard now enumerates `src/**/*.cs`. I re-placed the omitting spawn site at `src/IntentSystem.Supervisor/ZzTempOmittingSpawnSite.cs` — the same file that passed green last round — and it is now reported as an offender.

**Blocker 2 (±12-line window masking a neighbour) — closed.** The initializer-bound pass brace-matches each `new ProcessStartInfo { … }` and requires the encodings inside the *same* initializer. I re-placed the compliant-then-omitting pair in one file; the omitting one is now flagged. Both probes were live together and the guard failed on both:

```
Assert.Empty() Failure: Collection was not empty
Collection: ["src/IntentSystem.Cli/ZzTempNearbyOmittingSite.cs:17",
             "src/IntentSystem.Supervisor/ZzTempOmittingSpawnSite.cs:9"]
```

**Finding 3 (gh-scoped, broken doc) — fixed.** The subjectless sentence is repaired and the doc now states the codebase rule with `gh` as the motivating example: "Every redirected child stream pins both encodings … The `gh` path motivated this contract, but it applies equally to git, herdr, and every other child process." The field doc says "every redirected child subprocess".

**Finding 4 (private duplicate) — fixed.** `GhReviewCommandRunner` now uses `ProcessOutputEncoding.Utf8NoBom` and its private field and `using System.Text;` are gone. The Review project has a single definition.

Also still holding at this head: enumeration of every redirected-stream site under `src/` returns zero omissions (24 files, all declaring both encodings); the encoding value is untouched; `gh` behaviour, transport/delivery semantics, herdr and the process's own console encoding are unchanged; EN/JA parity intact; full suite green (`dotnet test IntentSystem.sln -c Release` — `IntentSystem.Cli.Tests` 4508 passed / 1 skipped, all eleven projects clean); `git diff --check` clean; `Closes #1365` correct.

### One structural gap remains: the guard is bound to two syntactic forms, not to the act of redirecting

The guard now recognises exactly two shapes — the literal `process.StartInfo.RedirectStandard…` assignment (±3 lines), and a literal `new ProcessStartInfo` followed by a brace initializer. Two ordinary C# spellings of the same spawn site fall outside both. I put them in one file and ran the guard:

```csharp
// A: target-typed new() — `new ProcessStartInfo` never appears
private static readonly ProcessStartInfo TargetTyped = new()
{
    FileName = "echo",
    RedirectStandardOutput = true,
    RedirectStandardError = true,
    UseShellExecute = false
};

// B: constructor + property assignment on a local not named `process.StartInfo`
var startInfo = new ProcessStartInfo("echo");
startInfo.RedirectStandardOutput = true;
startInfo.RedirectStandardError = true;
```

```
Passed!  - Failed: 0, Passed: 1 - IntentSystem.Cli.Tests.dll
```

Green, with two undecoded spawn sites in the tree. Neither idiom exists in the codebase today — so no acceptance criterion about actual coverage is violated, and there is no live defect — but the guard's stated job is the site that does not exist yet. Form B in particular (`var startInfo = new ProcessStartInfo(...)` then property assignments) is the most common spelling in .NET samples, so it is a likely shape for the tenth site. A guard keyed to the two spellings that happen to be in the tree today is the same shape of mistake as a helper keyed to the one callee that happened to exist in G484.

**Unblock — one form-independent assertion, no whack-a-mole.** Keep the two passes and add a per-file coverage check: for every file under `src/`, the count of `RedirectStandardOutput\s*=` assignments must be matched by at least as many `StandardOutputEncoding\s*=` assignments, and likewise for the error pair. That is blind to construction syntax and catches both probes above. I verified it is already satisfied by the tree at this head — all 24 files that redirect are exactly balanced (`IGitHubLabelMutator.cs` 2/2, `IssuePublishFlowCommand.cs` 2/2, every other file 1/1) — so it can be added without touching any production file. A Roslyn-based check over the compilation would be stronger still, but this is sufficient and cheap.

Please also paste the failing-guard output into the PR when you re-run it, so the demonstration review-context.md asks for is self-evidencing rather than asserted.

Everything else in this PR is approved as it stands; this is the only remaining item.

### What was NOT verified and must not be claimed

No Windows host was exercised; the cp932 decoding failure was not reproduced and no `notify delegate` / `herdr agent list` round trip was run against a non-UTF-8 console. All probe files were removed and the tree restored before the full suite run reported above.

tomohisa:
Re-repair pushed at `1773688`. The guard now also enforces per-file balance between every RedirectStandardOutput/Error assignment and its matching StandardOutput/ErrorEncoding assignment, covering target-typed `new()` and constructor-plus-assignment sites while retaining initializer-bound checks. Both omission forms failed the guard (exit 1) in temporary probes; probes removed. Focused guard and full Release suite passed. Exact-head CI run `31078998706` passed.

tomohisa:
## Re-review — PR #1368 @ 1773688: approved

**Platform**: macOS. The Windows cp932 runtime failure is structurally unreachable here, was **not reproduced**, and is not claimed. Verification is by source enumeration, by diff, and by exercising the guard against deliberately omitting sites.

### The residual gap is closed

The per-file count-parity assertion is in: every file under `src/` must match its `RedirectStandardOutput =` / `RedirectStandardError =` assignments with an equal number of `StandardOutputEncoding =` / `StandardErrorEncoding =` assignments. It is blind to construction syntax, so it backstops both form-specific passes. Re-ran the exact two idioms that passed green last round:

```
Assert.Empty() Failure: Collection was not empty
Collection: ["src/IntentSystem.Cli/ZzTempAssignmentOmitting.cs:redirect/encoding counts 1/0, 1/0",
             "src/IntentSystem.Cli/ZzTempTargetTypedOmitting.cs:redirect/encoding counts 1/0, 1/0"]
```

- target-typed `new()` with an object initializer — **now caught**
- `new ProcessStartInfo("echo")` plus property assignments on a local — **now caught**

**No false positive on the compliant modern idiom.** I placed a target-typed `new()` site that *does* declare both encodings alongside an omitting object-initializer site in `src/IntentSystem.Supervisor`; only the omitting one was reported, by both the count check and the initializer check. The strict-equality form is also sound in the converse direction: setting an encoding without the matching redirect throws at `Process.Start` in .NET anyway, so flagging that combination is correct rather than a nuisance.

**Guard green on the unmodified tree**, with no offender reported across all eleven `src/` projects.

### All earlier findings remain fixed at this head

| finding | state |
|---|---|
| Guard scanned only 2 of 11 `src/` projects | closed — enumerates `src/**`; the Supervisor probe is reported |
| ±12-line window masked a neighbouring omission | closed — initializer-bound brace matching, plus the new count backstop |
| Helper doc still gh-scoped, rename left a subjectless sentence | closed — states the codebase rule, `gh` as motivation only |
| `GhReviewCommandRunner` kept a private `Utf8NoBom` duplicate | closed — uses `ProcessOutputEncoding.Utf8NoBom`; private field and `using System.Text;` removed |

### Acceptance criteria

| criterion | evidence |
|---|---|
| All nine sites declare UTF-8 for stdout and stderr; no redirected-stream site remains without one | enumeration of every `ProcessStartInfo` with a redirected stream under `src/` returns zero omissions across 24 files, including `GhReviewCommandRunner` which was in neither the nine nor the fourteen |
| Helper renamed to a callee-neutral name, value unchanged, all call sites updated | git records a rename of `GitHubCliProcessEncoding.cs` → `ProcessOutputEncoding.cs`; the value is byte-identical `UTF8Encoding(encoderShouldEmitUTF8Identifier: false)`; the fourteen `gh` sites changed by pure identifier substitution; no private duplicate remains |
| Guard fails on an omitting site, demonstrated failing, passes on the fixed tree | four omission probes across three syntactic forms and two projects, each reported with file and line; clean tree green |
| EN/JA record the total, intermittent, environment-triggered Windows stall and withdraw the workaround | `docs/en/07-recovery.md` and `docs/ja/07-recovery.md` both name the unrelated-workspace pane title as trigger, state every transport operation stops at once, and state no non-ASCII avoidance workaround is needed |
| Encoding value, `gh` behaviour, transport/delivery semantics, herdr, process console output unchanged; no unrelated refactoring | diff touches only the encoding declarations, the rename, the doc, and the guard test; nothing herdr-side; `Console.OutputEncoding` untouched |
| ja/en parity under G613; full suite green | parity holds; `dotnet test IntentSystem.sln -c Release` green across all eleven projects (`IntentSystem.Cli.Tests` 4508 passed / 1 skipped) |

`git diff --check` clean. Intent reference: `intents/intent-cli/intent-tree/means/02-tooling-strategy.md` (node 02 — tooling correctness across platforms; a fix attached to a callee does not travel to the next callee, which is what the callee-neutral helper plus a syntax-independent guard now prevent). `Closes #1365` matches the linked source issue (G311). Tests passing is treated as necessary, not sufficient — each criterion above is mapped to a change or a reproduced observation.

### What was NOT verified and must not be claimed

No Windows host was exercised. The cp932 decoding failure was not reproduced, and no `notify delegate` / `herdr agent list` round trip was run against a non-UTF-8 console. All probe files were removed and the tree restored before the suite run reported above.

PR #1374, opened 2026-08-07
G629: track pending delegations and expose status

### LINKED ISSUE #1373: G629: A delegation that is never answered leaves no trace
# G629: A delegation that is never answered leaves no trace

## Goal

Give a dispatched task a **durable pending record** and a **query that answers whether it is still alive**: `notify status --task-id <id>` reports when the task was handed out, whether the recipient is running **by the same judgment `delegate` already makes**, whether the expected report arrived, and a verdict of `live`, `settled`, or `lost`. Detection only — no supervision loop and no automatic re-dispatch in this slice.

## Why This Slice Exists Now

**A recipient that dies mid-task is invisible to durable state. Measured on this host, 2026-08-06:**

- The team's event log holds **170 records — 117 completion, 21 question, 18 escalation, 14 blocked**. Every one is a report *from* a recipient. **Not one records a dispatch.**
- `runs.jsonl` carries **zero** notify or delegate events.
- `notify` has exactly three subcommands: `delegate`, `escalate`, `report`. **There is no way to ask what is outstanding.**

So an unanswered delegation is not a state the system can be in — it is the absence of a state. Nothing can time it out, nothing can list it, and nothing can tell a slow recipient from a dead one.

**The failure is not hypothetical, and not one team's.** The Zero4Racer team reported the exact sequence on `PR #1044` / `Z4R-G525`: `delivered: true` with an observed working transition, the recipient process exited mid-review, and the PR stayed at `intent-pr-reviewing` with queue-state at `review` and **no notification to anyone**. Unwatched, it stops there permanently. This team has hit the same class repeatedly on its own review seat.

**Two teams independently built the missing piece outside the tool.** This host runs external monitors (state-change wake, blocked-poll); the Zero4Racer team wrote a two-minute supervisor. Both reconstruct, from the outside, information the tool holds on the inside. That is the signal that the surface belongs in the tool.

**The detection substrate already exists too.** When a prompt never starts a turn, `notify` already reports `delivered: false` with `cause: receiver-transition-unobserved` and the detail `agent_prompt_stalled; agent prompt produced no observed state change within 5000 ms`, plus the correct `resend_permitted: true`. The reporting team singled this out as working well and observed that a status query can be built on it rather than beside it: the hard part — deciding that nothing started — is already implemented and simply is not retained anywhere after the call returns.

**The liveness judgment already exists — it is simply not askable.** `notify delegate` refuses with `cause: agent-not-running` / `observed: <role>@<ws>/<pane>[idle; running=false]`. The reporter's supervisor read `agent_status` instead and **mis-classified a dead process as healthy**, because `herdr agent get` can report `idle` for a process that no longer exists. Exposing the running flag as a query removes that trap for every caller.

**Detection before automation.** This slice adds the pending record, the query, and the exposed liveness flag; the supervision loop and the opt-in re-dispatch are a separate slice, so a reviewer cannot approve automatic re-dispatch on the strength of the query's evidence.

**`lost` must be earned, not assumed.** Liveness alone cannot distinguish *crashed* from *finished and exited* — the reporter says so explicitly and it is why they declined to build auto-redispatch. The verdict therefore requires **both** facts: recipient not running **and** no matching report. That combination is only knowable to intent-cli, which holds the expected artifacts.

**Preview surface.** This lands after the v0.12.0 freeze, so under G628's rule it ships as **preview through the 1.0 line** and is formalised in 2.0 or later. The docs must say so.

## Current Observed State

- No durable record is written when a task is handed out; the only artifact is the file-backed task file from G619, which exists solely for recipients that declare it and is not a queryable pending set.
- `notify` offers no query subcommand; outstanding work is invisible.
- The `running=false` liveness judgment is internal to `delegate`'s refusal path.

- G598's delivery semantics are correct and unchanged: `delivered` comes from the observed unattended working transition, `resend_permitted` is reported separately.
- G626's outcome vocabulary is correct; this slice reads state, it does not rename outcomes.
- G619's file-backed task file is correct and stays; the pending record is a different artifact with a different job.
- Reports already carry `--task-id`, so resolution needs no new identifier.

## Target Repo / Path / Part

Target part: the durable pending-delegation record, the `notify status` query, and the exposed liveness flag.

## In Scope

- **A pending record written at dispatch**, carrying the task id, the recipient identity, the expected artifact, and the dispatch timestamp, in the team's durable store. **Writing it fails closed before anything reaches the pane**, following G619's precedent for the task file.
- **Resolution by the arriving report**: a report whose task id matches an open record resolves it. A report whose task id matches nothing **does not silently create or resolve a record** — it is refused with both values named, per the disagreement rule G625 shipped.
- **`notify status --task-id <id>`** reporting the dispatch time, the recipient's **current** liveness, whether the expected report arrived, and a verdict:
  - `live` — recipient running, no report yet. **Duration is not evidence of death**; a recipient running for an hour is `live`.
  - `settled` — the matching report arrived, whatever the recipient is doing now.
  - `lost` — recipient not running **and** no matching report.
- **Liveness is the running flag, not the status string** — the same judgment `delegate` uses in its `agent-not-running` refusal, named in output so a caller cannot repeat the reporter's mistake of trusting `idle`.
- **Preview labelling** per G628: the surface is marked preview in EN and JA guidance, with the promise's preview lane cited.

## Out Of Scope

- Any supervision loop, polling command, or automatic re-dispatch — the following slice.
- Changing delivery semantics, recipient identity, the settle contract, or G626's outcome vocabulary.
- Restarting, killing, or otherwise acting on a recipient process.
- Timing out a task or declaring a task failed on elapsed time alone.

The child PR writes a durable pending record when a task is dispatched, carrying the task id, the recipient identity, the expected artifact and the dispatch timestamp in the team's durable store, and fails closed before anything reaches the recipient pane if that record cannot be written. A report whose task id matches an open pending record resolves it, while a report whose task id matches no open record is refused with both the supplied and the known values named rather than silently creating or resolving anything. It adds notify status --task-id which reports the dispatch time, the recipient's current liveness, whether the expected report has arrived, and a verdict of live when the recipient is running and no report has arrived, settled when the matching report has arrived regardless of the recipient's current state, and lost only when the recipient is not running and no matching report exists, with elapsed time never being sufficient evidence for lost. Liveness is determined by the same running flag that notify delegate uses in its agent-not-running refusal rather than by the agent status string, and the output names which flag it used. The surface is documented in EN and JA as a preview surface under the compatibility promise's preview lane, delivery semantics, recipient identity, the settle contract and the outcome vocabulary are unchanged, no supervision loop or automatic re-dispatch is introduced and no recipient process is acted upon, tests cover the fail-closed write, the matching and non-matching report cases, all three verdicts including a long-running live recipient and a not-running recipient whose report did arrive, and the liveness flag being read rather than the status string, ja and en stay in parity under the G613 terminology policy, and the full suite is green.

## Acceptance Criteria

- Pending record written at dispatch with task id, recipient, expected artifact, timestamp; unwritable record fails closed **before** the pane receives anything.
- Matching report resolves the record; non-matching task id refused naming both values; nothing silently created or resolved.
- `notify status --task-id` reports dispatch time, current liveness, report arrival, and verdict.
- `live` for a running recipient regardless of elapsed time; `settled` on matching report; `lost` only when not running **and** no report.
- Liveness read from the running flag, not the status string, and the output names it.
- Preview labelling in EN and JA citing the promise's preview lane.
- No supervision loop, no re-dispatch, no process action; delivery/settle/identity/vocabulary unchanged; parity under G613; full suite green.

## Verification

- Dispatch, then query before any report: expect `live` with the recipient running.
- Keep a recipient running well past the internal settle bound and query: still `live` — assert elapsed time alone never yields `lost`.
- Kill the recipient process with no report sent, then query: expect `lost`.
- Kill the recipient **after** its report arrived, then query: expect `settled`, not `lost`.
- Construct a recipient whose status string reads `idle` while the process is gone, and assert the verdict follows the running flag.
- Make the pending record unwritable: assert nothing reached the pane.
- Report with an unknown task id: assert refusal naming both values.

## Related Links

- Zero4Racer delegation-supervision proposal, 2026-08-06 — the `PR #1044` / `Z4R-G525` loss, the `agent_status` mis-classification, and the argument that only intent-cli can separate crashed from finished.
- Measurement, 2026-08-06: 170 team events, none a dispatch; zero notify events in `runs.jsonl`; `notify` has three subcommands.
- G598 — the delivery semantics kept intact; G626 — the outcome vocabulary this slice reads but does not rename; G619 — the fail-closed-before-send precedent; G625 — the name-both-values refusal rule; G628 — the preview lane this surface enters.

## Knowledge Maintenance

- Intent placement: node 08 primary (delegation lifecycle), node 03 supporting (durable state and audit).
- ADR candidate: none.
- Diagram candidate: none.
- Closeout writeback expected: yes — node 08 records that a dispatch is a state and not merely an act, that liveness alone cannot separate crashed from finished, and that elapsed time is not evidence of death.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1373

- Persist team-scoped pending delegation dispatch/report snapshots before delivery.
- Refuse unknown or settled reports and expose `notify status` with named liveness/verdict output.
- Document the preview-through-1.x compatibility boundary in EN/JA and add lifecycle/liveness coverage.

Local verification: focused notify/session/docs suite 116 passed; full CLI test suite passed with one pre-existing skipped test.


tomohisa:
## Review verdict: APPROVE — head `140b5205`

Applied canonically with `intent-cli automation pr-transition --transition approved --write` (labels: `intent-target`, `intent-pr-approved`; `intent-pr-reviewing` removed). PR was draft — `automation host-review-diagnostics --pr-draft true --draft-review-ready` returned `draft-ready-to-promote` (G376), so `gh pr ready` ran first; the draft flag was never approved through. Head verified `140b5205` before and after. Base `main` matches `direct-main`. `review closeout-plan` → `ready`, no gaps; `guide review` → `ready: true`, no gaps.

Review ran read-only in a managed worktree at `.intent-cli/worktrees/G629-review` (never `/tmp`), removed after the pass. Per review-context, the dead-recipient states were **constructed, not reasoned about**: I added a throwaway reviewer-only probe, captured the real command output below, then deleted it — it is not part of the PR and the worktree was clean afterwards.

### The four verdict cases, with command output

**CASE 1 — dispatched, recipient running → `live`**
```
- dispatched at: 2026-08-06T12:00:00.0000000+00:00
- recipient running: true
- liveness source: herdr.agent_running
- report arrived: false
- verdict: live
```

**CASE 2 — same recipient still running, clock advanced 45 days → still `live`** (review-context #1)
```
- dispatched at: 2026-08-06T12:00:00.0000000+00:00
- recipient running: true
- report arrived: false
- verdict: live
```
Structural confirmation as well as empirical: `NotifyPendingLiveness` and the verdict expression contain **no** `TimeSpan`/`Elapsed`/`Duration` term at all — the verdict is `ReportArrived ? settled : Running ? live : lost`. Elapsed time cannot reach it.

**CASE 3 — `agent_status` reads `idle` while the process is gone, no report → `lost` via the running flag** (review-context #3)
```
- recipient running: false
- liveness source: herdr.agent_running
- verdict: lost
… has no agent with running=true (status strings are ignored).
```
This is exactly the state that mis-classified a dead process as healthy in the Zero4Racer supervisor. `ProbeHerdr` filters on `agent.AgentRunning` and the output names `herdr.agent_running` as the source, so a caller cannot repeat the mistake.

**CASE 4 — recipient exits *after* its report arrived → `settled`, not `lost`** (review-context #2)
```
- recipient running: false
- report arrived: true
- verdict: settled
… Matching report status 'completed' arrived.
```
`settled` is decided by the recorded report, not by liveness — the distinction the reporting team said liveness alone cannot make.

### Fail-closed write and unknown task id
- **Fail closed before the pane** (review-context #4 / AC1): with the pending write denied, `notify delegate --write` exits `1` with `cause: pending-record-write-failed`, `delivered: false`, **zero** transport runner calls, and no `pending.jsonl` created. Ordering in `ExecuteDelivery` confirms it: pending record write → G619 task-envelope file write → `transport.Deliver`. Nothing reaches the pane first.
- **Unknown task id refused naming both values** (review-context #5 / AC2): `cause: unknown-task-id`, summary `Report task id 'G629-not-a-task' does not match an open pending delegation. Known open task ids: G629-demo.` — supplied and known both named, no record created or resolved, and zero transport calls.

### Remaining acceptance criteria
- **AC3 `notify status` output** — reports dispatch time, current liveness, the named liveness source, report arrival and verdict, in both markdown and JSON. Registered in `CommandRouter` and surfaced in `guide commands list`; read-only (no `--write`).
- **AC5 liveness named** — `liveness_source: herdr.agent_running` (agmsg path names `agmsg.team_roster`).
- **AC6 preview labelling** — `preview-through-1.x` rows for both `notify status` and the pending-delegation record in EN **and** JA `1.0-compatibility-ledger.md`, plus "Preview through 1.x (G629)" blockquotes in both `12-agent-message-orchestration.md` pages citing the promise's preview lane. EN/JA in parity under G613.
- **AC7 boundaries** (review-context #6) — grep over every added `src/` line for `while(`/`for(`/`Thread.Sleep`/`Task.Delay`/`Timer`/`Kill(`/`Restart`/`Respawn`/`re-dispatch`/`Retry`/`Poll` returns only prose ("without supervision or re-dispatch", "without prompting, waiting, restarting, or otherwise acting on a process") — no loop, no re-dispatch, no process action. Delivery semantics, settle contract and G626's vocabulary are untouched: `workingTransition`, `settleOutcome` and `resendPermitted` are forwarded unchanged, and no outcome kind is renamed.

### Design alignment
`intents/intent-cli/intent-tree/means/08-agent-message-orchestration.md` (primary) — a dispatch becomes durable, inspectable state rather than an unrecorded act, matching the node's rule that holds must be recorded and detectable rather than invisible. `…/03-state-and-audit-strategy.md` (supporting) — the append-only `pending.jsonl` keeps dispatch and resolution auditable without turning the six-field event channel into a second, incompatible state store.

### Test evidence (necessary, not sufficient — paired with the packet evidence above)
- Full suite at this head: **4520 passed, 1 skipped, 0 failed** in `IntentSystem.Cli.Tests`, all other projects green.
- `git diff --check` clean; exact-head CI on `140b5205` concluded **success**.

### Non-blocking observations (not merge blockers)
1. `GuideCommandsListCommand.cs` line 161 loses one space of indentation (`           Purpose =` instead of 12 spaces). Cosmetic only; no formatting gate fails.
2. A duplicate report against an already-settled record is refused with `cause: unknown-task-id`, which is slightly imprecise — the id *is* known, it is already resolved. The refusal is correct and safe (nothing is silently re-resolved) and the summary names the open ids, but a distinct cause such as `task-already-settled` would read better. Worth a follow-up rather than blocking.
3. The PR's own tests cover all three verdicts but none advances the clock, so "a long-running live recipient" from the standalone contract is not directly asserted in-repo. I demonstrated it above (45 days) and the verdict has no time input at all, so conformance holds; adding a clock-advanced case would make the guarantee regression-proof.

PR #1376, opened 2026-08-07
G630: add notify supervise recovery loop

### LINKED ISSUE #1375: G630: Recovery is not finished until the caller knows what was lost
# G630: Recovery is not finished until the caller knows what was lost

## Goal

Add `notify supervise --domain <d> --team <t> [--interval <s>]`: a loop that watches recipients holding open delegations, **stays silent while they are alive**, and when one is gone **restarts it, proves readiness against the response line, and notifies the delegating role that the in-flight task was lost and must be re-dispatched**. Automatic re-dispatch is a separate opt-in flag, **off by default**.

## Why This Slice Exists Now

**G629 makes a lost delegation visible; nothing yet acts on it.** A verdict nobody queries is the same as no verdict, and both teams running herdr-only currently close that gap with hand-written scripts outside the tool.

**The reporting team's supervisor worked, and its report names the load-bearing part.** Their two-minute loop restarted the dead recipient and — critically — **told the caller the in-flight task was lost**. Their words: recovery alone is not enough; the work continues only once the loss is reported. `PR #1044` completed after exactly one such cycle. A restart that silently resumes an empty seat leaves the delegating role believing work is in flight.

**Three hazards this slice must be built around, all measured rather than assumed:**

1. **A restart is not a readiness claim.** This team recorded the rule after reporting a seat restarted when the registering prompt had never been sent: start, register, prove, then report.
2. **A readiness gate must match the response line, never the prompt echo.** The reporting team's first gate matched the nonce anywhere in the pane and passed on a prompt **still sitting unsent in the composer** — the evidence was a pane holding three concatenated prompts and a session that had consumed **zero tokens**. A gate that has not been shown to fail on the broken state has not been tested.
3. **Starting an agent does not necessarily replace a broken one.** Both teams observed that recovery required killing the process by pid first; a start issued over a wedged seat left the broken state in place. Supervision must verify the process it intended to replace is actually gone.

**Why the loop belongs in the tool rather than in each team's scripts.** Deciding that a recipient is lost needs the expected artifacts — PR label, queue-state, the report inbox — and only intent-cli holds them. The reporting team declined to build automatic re-dispatch for precisely this reason: from outside, *crashed* and *finished and exited* are indistinguishable. Inside, they are not.

**Silence is the normal output.** A supervisor that reports on healthy seats trains its reader to ignore it, which is the failure G626 removed from dispatch vocabulary; the same rule applies here.

**Re-dispatch stays opt-in and off by default.** Detection and recovery are safe to automate; deciding that a task should be sent again is a judgment about work, not about processes, and defaults to the human.

**Preview surface** under G628: added after the v0.12.0 freeze, so it ships preview through the 1.0 line and is formalised in 2.0 or later.

## Current Observed State

- Nothing watches recipients; a dead seat is discovered when a human notices.
- Both teams run external monitors that reconstruct, imperfectly, what the tool knows.
- Recovery is entirely manual: kill by pid, start, send the registering prompt, prove readiness, re-dispatch.

- G629's pending record, `notify status` verdicts, and the running-flag liveness judgment are shipped and correct; this slice consumes them and does not restate them.
- G617's unattended-launch recipes carry the per-agent start command; supervision starts a seat the way its recipe says to.
- Delivery semantics, recipient identity and the settle contract are correct and unchanged.

## Target Repo / Path / Part

Target part: the supervision loop, its recovery sequence, and its loss notification.

## In Scope

- **The loop**: at a bounded interval, for each recipient holding an open pending record, evaluate the G629 verdict. `live` ⇒ **no action and no output**. `settled` ⇒ the record closes; no action.
- **The recovery sequence on `lost`, in order and each step verified before the next**: confirm the recipient is genuinely gone rather than mid-exit; ensure the old process is actually no longer present before starting a replacement; start the seat by its recipe; send the registering prompt; **prove readiness by matching a nonce in a response line, not anywhere in the pane**; only then report.
- **The loss notification**: the delegating role is told, in one message, that the recipient was recovered **and that the task it was holding was lost and must be re-dispatched**, naming the task id. Recovery without this message is not a completed recovery.
- **`--auto-redispatch`, default off**: when enabled, the supervisor re-sends the lost task through the normal dispatch path after readiness is proven, and says it did. When disabled, it only reports.
- **Fail-closed supervision**: if any recovery step cannot be verified — the old process still present, readiness unproven, the recipe unknown — the supervisor stops for that recipient, reports what it could not verify, and takes no further action. It never reports a readiness it did not observe.
- **Preview labelling** in EN and JA per G628.

## Out Of Scope

- Changing G629's verdict rules, delivery semantics, recipient identity, or the settle contract.
- Supervising anything other than recipients with an open pending record.
- Repairing the underlying agent-runtime defects; this slice works around them and says so.
- Making re-dispatch the default.

The child PR adds notify supervise for a domain and team with a bounded interval which, for each recipient holding an open pending delegation record, evaluates the verdict defined by the preceding slice and takes no action and produces no output while that verdict is live, closes the record and takes no action when it is settled, and performs a recovery sequence only when it is lost. The recovery sequence confirms the recipient is genuinely gone rather than mid-exit, verifies that the previous process is no longer present before a replacement is started, starts the seat by its recorded unattended-launch recipe, sends the registering prompt, and proves readiness by matching a nonce in a response line rather than anywhere in the pane so that a prompt still sitting unsent in the composer can never satisfy the gate. After readiness is proven the supervisor notifies the delegating role in one message that the recipient was recovered and that the task it held was lost and must be re-dispatched, naming the task id, and a recovery that omits this message is not complete. An auto-redispatch flag defaulting to off causes the lost task to be re-sent through the normal dispatch path after readiness is proven and reported as such, while with the flag absent the supervisor only reports. If any recovery step cannot be verified, including the previous process still being present, readiness being unproven, or the recipe being unknown, the supervisor stops for that recipient, reports precisely what it could not verify, takes no further action on it, and never reports a readiness it did not observe. Tests cover the silent live path, the settled path, the full lost-and-recovered path, the readiness gate rejecting an unsent prompt echo, each fail-closed stop, and both states of the re-dispatch flag; the verdict rules, delivery semantics, recipient identity and the settle contract are unchanged; the surface is documented in EN and JA as preview under the compatibility promise's preview lane; ja and en stay in parity under the G613 terminology policy; and the full suite is green.

## Acceptance Criteria

- `live` produces no action and no output; `settled` closes the record with no action.
- On `lost`, the recovery sequence runs in order with each step verified before the next, including that the previous process is gone before a replacement starts.
- Readiness is proven by a nonce in a **response line**; a nonce present only as an unsent prompt echo **fails** the gate, and there is a test that demonstrates the failure.
- The delegating role receives one message naming the task id, stating recovery and stating that the task was lost and must be re-dispatched.
- `--auto-redispatch` defaults to off; enabled, it re-sends after proven readiness and says so.
- Any unverifiable step stops supervision for that recipient with a report naming what could not be verified; no readiness is ever reported unobserved.
- Preview labelling in EN and JA; verdict rules, delivery/identity/settle unchanged; parity under G613; full suite green.

## Verification

- A live recipient across several intervals: assert zero output and zero actions.
- A recipient killed mid-task: assert the ordered sequence, the proven readiness, and the loss message naming the task id.
- Feed the gate a pane whose nonce appears only in an unsent composer line: assert the gate **fails**.
- Leave the old process present and assert the supervisor stops rather than starting a second seat.
- Run with and without `--auto-redispatch`: assert re-send happens only when enabled and only after readiness.

## Related Links

- Zero4Racer delegation-supervision proposal, 2026-08-06 — the working supervisor, the `PR #1044` cycle, the prompt-echo gate failure with a zero-token session, and the argument for placing the decision inside intent-cli.
- G629 — the pending record and verdicts this slice consumes; G617 — the recipes it starts seats from; G626 — the rule that a signal firing on the normal path is ignored; G628 — the preview lane.
- This team's recorded rule after a premature readiness report: start, register, prove, then report.

## Knowledge Maintenance

- Intent placement: node 08 primary (recipient lifecycle and recovery), node 03 supporting (verified state transitions).
- ADR candidate: none.
- Diagram candidate: none.
- Closeout writeback expected: yes — node 08 records that a recovery is incomplete until the caller learns what was lost, that a gate untested against the broken state is not a gate, and that starting a seat does not prove the old one is gone.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1375

## Summary
- add bounded `notify supervise` wakes with silent live/settled paths
- recover lost herdr recipients only after process attribution, replacement, registration, and response-line readiness proof
- notify the delegating role of task loss; keep auto-redispatch opt-in
- document the EN/JA preview lane and add focused G630 coverage

## Verification
- `dotnet test IntentSystem.sln --no-restore`
- `git diff --check`


tomohisa:
## Review verdict: APPROVE — PR #1376 @ `8ee090dbc326a325c73d5015d9d845980c17486b`

Reviewed against the committed G630 packet and review context. Every negative case below was **constructed and executed** in a managed worktree with a reviewer-only probe (deleted afterwards); none of the evidence is reasoned about.

### Packet contract clause restated

> "Recovery without this message is not a completed recovery" — the delegating role must be told, in one message, that the recipient was recovered **and** that the task it was holding was lost and must be re-dispatched, naming the task id.

Verified: `BuildLossNotification` emits exactly one payload carrying `task_id`, `recovered`, `lost`, `must_redispatch`. A delivery failure after proven readiness does **not** report success — it returns `outcome=recovered-but-not-reported` with a cause (`NotifySupervisor.cs`).

### Intent reference touched by the diff

`intents/intent-cli/intent-tree/means/08-agent-message-orchestration.md` — "Receiver readiness boundary": readiness is a handshake and only an acknowledged reply proves it; "a silent completion is lost work from the orchestrator's perspective". The response-line gate and the mandatory loss notification are the direct mechanisation of both rules. Supporting node `03-state-and-audit-strategy.md` covers the verified-transition ordering.

### AC-by-AC

**AC1 — `live` produces no action/output; `settled` closes with no action.** ✅
Five consecutive wakes against a running recipient: `silent=true actions=0 nonProbeCommands=0`; zero `kill`/`agent start`/`agent prompt` calls. A settled record (report written) produced `silent=True actions=0 commands=0` — the supervisor issued **no herdr command at all**, so `settled` cannot trigger recovery.

**AC2 — ordered recovery, each step verified before the next.** ✅ Observed command sequence on the lost path:
```
agent list                                  # verdict probe (G629)
agent list                                  # confirm gone, not mid-exit
pane process-info --pane wG630:p2           # attribute foreground processes
kill -TERM 7001                             # pid-targeted, only after attribution
pane process-info --pane wG630:p2           # re-verify the old process is gone
agent list                                  # re-verify no recipient reappeared
agent start implementation --kind codex --pane wG630:p2 --timeout 10000 -- --sandbox
agent list                                  # verify exactly one replacement + cwd match
agent prompt wG630:p2 "$agmsg actas implementation. Register this role, then reply with exactly READY <nonce> on a separate response line." --wait --until working
agent read wG630:p2 --source recent-unwrapped --lines 200   # prove readiness
```
Nothing is started before the previous process is proven absent, and readiness is read only after registration.

**AC3 — readiness proven by a *response line*; a prompt echo FAILS.** ✅ This is the load-bearing check, so I fed the gate five distinct broken panes. All five stopped fail-closed with `cause=readiness-unproven`, `recovered=false`, and **zero** loss notifications:

| constructed pane | result |
|---|---|
| A. the supervisor's own registering prompt sitting unsent in the composer (verbatim) | `stopped-fail-closed / readiness-unproven` |
| B. nonce embedded mid-line (`composer (unsent): please reply READY <nonce> now`) | `stopped-fail-closed / readiness-unproven` |
| C. wrong prefix (`READYX <nonce>`) | `stopped-fail-closed / readiness-unproven` |
| D. bare nonce, no `READY` prefix | `stopped-fail-closed / readiness-unproven` |
| E. **three concatenated unsent prompts** — the exact shape of the reported field failure | `stopped-fail-closed / readiness-unproven` |
| F. positive control: `READY <nonce>` alone on its own line amid chatter | `recovered-and-reported` |

```
No response line exactly matching 'READY revnonce' was observed. A nonce in an unsent prompt echo is not readiness.
```
`HasResponseLine` requires whole-line ordinal equality after trim, and reads `--source recent-unwrapped` so soft wrapping cannot manufacture a standalone line. Case E is the specific failure the reporting team hit; it is rejected. The PR ships its own `ReadinessGate_RejectsAnUnsentPromptEcho_G630` test, so the failure path is covered in CI, not just in this review.

**AC4 — one message naming the task id, stating recovery and loss.** ✅ Quoted verbatim from the constructed run:
```json
{"notification":"supervise-recovery","task_id":"G630-demo","recipient_role":"implementation",
 "recovered":true,"lost":true,"must_redispatch":true,"auto_redispatched":false,
 "readiness_nonce":"revnonce",
 "summary":"Recipient recovered; the in-flight task was lost and must be re-dispatched.",
 "redispatch_detail":null}
```

**AC5 — `--auto-redispatch` defaults off; enabled it re-sends after proven readiness and says so.** ✅
- Off (default, flag absent): `outcome=recovered-and-reported`, `auto_redispatched=false`, redispatch invocations `0`.
- On: `outcome=recovered-and-redispatched`, order `[redispatch, notify]` with the readiness read at call index 9 — strictly after the gate; the notification flips to `must_redispatch=false` / `auto_redispatched=true`.
- **On, but readiness unproven:** `cause=readiness-unproven`, `redispatchCalls=0`. Re-dispatch cannot fire ahead of the gate.
- `--dry-run` performs no `kill` / `agent start` / `agent prompt` / delegate / append.

**AC6 — any unverifiable step stops that recipient, naming what could not be verified; readiness is never reported unobserved.** ✅ Constructed:
- Old process still present after `kill -TERM`: `cause=old-process-still-present`, "no replacement was started" — and **no `agent start` was issued**, so a second seat cannot stack on a wedged one.
- Foreign cwd: `cause=process-attribution-unverified` — "Process pid 7001 reports cwd '/tmp/another-team', not recorded role cwd '/tmp/g630-role'; refusing a cross-team kill." **No `kill` call at all.**
- Additional guarded stops read in code: `recovery-contract-incomplete` (missing recipe/roles, checked *before* any mutation), `recipient-not-gone`, `multiple-agents-at-pane`, `replacement-not-running`, `replacement-cwd-mismatch`, `launch-recipe-unknown`, `registration-failed`, `agent-list-invalid`, `process-info-invalid` (an absent `foreground_processes` array is refused rather than read as "gone").

**AC7 — preview labelling EN+JA; G629 contracts unchanged; G613 parity; full suite green.** ✅
- `docs/en|ja/1.0-compatibility-ledger.md`: new `notify supervise` row marked `preview-through-1.x` in both; the pending-record row updated to `G629/G630` in both.
- `docs/en|ja/12-agent-message-orchestration.md`: matching "Bounded recipient supervision (G630)" sections with identical command synopsis, ordering prose, and a preview-through-1.x blockquote in each — EN/JA in parity.
- `NotifyPendingLiveness.cs` and `ExecuteStatus` are **untouched** by the diff; the `live|settled|lost` rules, delivery semantics, recipient identity and the settle contract are unchanged. Store/topology/transport changes are strictly additive optional fields (`delegating_role`, `report_to_role`, `cwd`, `kind`, `launch_args`, `HerdrAgentState.Cwd` with a default), so existing records deserialise unchanged.
- Base branch `main` matches policy `direct-main`.
- Full suite in a clean worktree at this head: **4528 passed / 1 skipped / 0 failed** (plus all satellite test projects green). `git diff --check` clean. CI at `8ee090d`: **success**.

### Non-blocking observations (no change requested)

1. `DefaultRedispatch` temporarily swaps the process-wide static `NotifyCommand.ProcessRunnerFactory` and restores it in a `finally`. Correct for the single-threaded CLI loop, but it is global mutable state reachable from a loop; worth a note if supervision ever becomes concurrent.
2. On the non-redispatch path the original pending record stays open after recovery. That is the right default (the caller must decide), but a long-running supervisor will re-evaluate that record each wake; it stays silent only because the replacement seat is live. Worth confirming intent if a seat is recovered and then left idle.
3. `Register` selects the `/agmsg` vs `$agmsg` prefix from `record.Kind`; an unrecognised kind silently takes the `$agmsg` branch. It fails closed later at the readiness gate, so behaviour is safe, but an explicit unknown-frontend stop would name the cause earlier.

Passing tests alone would not have been sufficient here; the approval rests on the constructed broken states above, in particular the concatenated-unsent-prompt pane (case E) that the gate rejects.

PR #1380, opened 2026-08-07
G638: make CI waits durable and report sleeping seats

### LINKED ISSUE #1379: G638: CI finishing is the one event that sends no message
# G638: CI finishing is the one event that sends no message

## Goal

Close the message-driven hole where **a green CI wakes nobody**: let an orchestrator record that it is waiting on a pull request's checks, and teach `automation stalled-work` to classify **checks green with the next transition not taken** as actionable. Also warn when a report is addressed to a role whose seat is not running, so a completion cannot vanish into a sleeping recipient.

## Why This Slice Exists Now

**Reported by the Sekiban team on 2026-08-06, and independently true on this host.** Their `PR #1108` sat complete with 4/4 green checks while every seat was idle. Nobody was at fault:

1. implementation opened the PR and reported; the orchestrator woke in the same cycle.
2. **At that instant the checks were still queued or running**, so the orchestrator did what the guidance says — treat CI as something to re-check at a later wake — and ended its wake.
3. The checks then went green. **No message is produced when checks finish**, so nothing woke anyone again.
4. `stalled-work` returned nothing actionable, because no class describes this state, so the watchdog did not catch it either.

**The more careful the orchestrator, the more reliably it stops.** Guidance presents timers as a fallback, so an orchestrator that follows it arms nothing and waits for a message that will never exist.

**This host proves the same gap by what it had to build.** The design thread runs an external CI-terminal watcher whose entire job is to notice a finished run and wake the orchestrator — a script that exists only because the tool cannot express "wake me when these checks settle". Two teams, two hand-written watchers, one missing surface. Measured here today: `automation stalled-work` reports only `operator-attention-pending` and `knowledge-writeback-pending`, and `automation` has no CI-related subcommand at all.

**A second, quieter failure in the same report**: implementation reported with `--report-to orchestrator` while the orchestrator was asleep, so the design thread's inbox held **zero** entries for that task and design did not learn the work was finished until it read a pane directly. The report was delivered correctly and still reached nobody awake.

**Record the wait; do not poll from intent-cli.** The orchestrator declares that it is waiting on a PR's checks, and that declaration is durable state a supervisor or a later wake can act on. intent-cli does not spawn a watcher, and this slice does not make it a daemon.

**`stalled-work` is where detection belongs**, because it is already the surface a watchdog reads and it already holds the repository facts needed to judge the state.

**The report-to warning is advisory, never a refusal** — the G618 precedent. A team may legitimately report to a role that is momentarily down.

## Current Observed State

- No way to record "waiting on checks for PR N"; the intention lives only in an orchestrator's head between wakes.
- `stalled-work` classes cover attention and write-back debt; nothing covers a PR whose checks finished while the next transition was never taken.
- `notify report` succeeds identically whether the addressed role's seat is running or not.

- G629's pending-delegation record and `notify status` are shipped; this slice adds a different waiting kind and does not restate them.
- G630's supervision loop is shipped and is the natural consumer of a recorded CI wait.
- Delivery semantics, seat identity and the settle contract are unchanged.

## Target Repo / Path / Part

Target part: the recorded CI wait, the `stalled-work` classification, and the report-recipient warning.

## In Scope

- **A recorded CI wait**: an orchestrator can register that a named pull request's required checks are being awaited, carrying the PR, the head it observed, and what transition is owed once they settle. It is durable, it is cleared when the transition is taken, and **registering it does not start any polling**.
- **A `stalled-work` class for the state**: a PR with the created-work label whose required checks are all green while the owed transition has not been taken is reported as actionable, naming the PR and the transition. **A PR whose checks are still running is not stalled** — the class must not fire on work that is simply in progress.
- **A stale-head guard**: if the head has moved since the wait was recorded, the item says so rather than asserting green checks for a head nobody is waiting on.
- **A report-recipient warning**: when `notify report` addresses a role whose seat is not running, the output warns, names the role and the observed liveness, and **delivery still proceeds**. It never refuses.
- **Guidance stating the shape**: in a message-driven transport, every transition is carried by a message except one — checks finishing — so a wait on checks must be recorded rather than remembered, and a report addressed to a sleeping role is delivered and unread.

## Out Of Scope

- Polling, watching, or any background process started by intent-cli.
- Automatically taking the owed transition; this slice makes the state visible, it does not act on it.
- Changing delivery semantics, seat identity, the settle contract, or G629's verdicts.
- Any new label or change to the existing label vocabulary.

The child PR lets an orchestrator record a durable wait on a named pull request's required checks, carrying the pull request, the head observed at registration and the transition owed once the checks settle, cleared when that transition is taken, with registration starting no polling of any kind. It adds a stalled-work class reporting as actionable a pull request that carries the created-work label and whose required checks are all green while the owed transition has not been taken, naming the pull request and the transition, while a pull request whose checks are still running is not reported as stalled, and it guards against a moved head by saying so rather than asserting green checks for a head no one is waiting on. It adds an advisory warning when notify report addresses a role whose seat is not running, naming the role and the observed liveness while delivery still proceeds and is never refused. English and Japanese guidance states that in a message-driven transport every transition is carried by a message except checks finishing, so a wait on checks must be recorded rather than remembered, and that a report addressed to a sleeping role is delivered and unread. Intent-cli starts no background process and takes no owed transition automatically, delivery semantics, seat identity, the settle contract and the preceding slice's verdicts are unchanged, no label is added or altered, tests cover the recorded wait and its clearing, the green-and-untaken case, the still-running case not firing, the moved-head case, and the warning firing and not firing with delivery proceeding in both, the surface is documented as preview under the compatibility promise's preview lane, EN and JA stay in parity under the G613 terminology policy, and the full suite is green.

## Acceptance Criteria

- A CI wait can be recorded with PR, observed head and owed transition; it is durable, cleared on transition, and starts no polling.
- `stalled-work` reports green-checks-with-untaken-transition as actionable, naming PR and transition.
- Still-running checks do **not** produce a stalled item.
- A moved head is reported as such rather than as green checks.
- `notify report` warns when the addressed role's seat is not running, naming role and liveness; delivery proceeds; never refuses.
- Guidance states the message-driven exception and the sleeping-recipient consequence.
- No background process, no automatic transition, no label change; delivery/identity/settle/verdicts unchanged; preview labelling; parity under G613; full suite green.

## Verification

- Record a wait, leave checks running, run `stalled-work`: assert no stalled item for it.
- Let the checks go green without taking the transition: assert the actionable item naming PR and transition.
- Take the transition: assert the item and the wait clear.
- Push a new head after recording: assert the moved-head report rather than a green assertion.
- Report to a role whose seat is stopped: assert the warning and that delivery still reports normally.
- grep the diff for any spawned process or timer: assert none.

## Related Links

- Sekiban team report, 2026-08-06: `PR #1108` green at 4/4 with every seat idle; the orchestrator deferred while checks were running and was never woken again; `stalled-work` found nothing actionable; the completion report went to a sleeping orchestrator and design's inbox held zero entries.
- Measured on this host, 2026-08-06: `stalled-work` classes are `operator-attention-pending` and `knowledge-writeback-pending` only; `automation` has no CI subcommand; the design thread runs an external CI-terminal watcher whose only purpose is to wake the orchestrator when a run finishes.
- G629 — the pending-delegation record this sits beside; G630 — the supervision loop that can consume a recorded wait; G618 — the warn-never-refuse precedent; G628 — the preview lane.

## Knowledge Maintenance

- Intent placement: node 08 primary (message-driven transitions), node 03 supporting (durable waits and stall classification).
- ADR candidate: none.
- Diagram candidate: none.
- Docs update: ja/en orchestration guidance (in scope).
- Closeout writeback expected: yes — node 08 records that checks finishing is the one transition no message carries, and that a report addressed to a sleeping role is delivered and unread.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1379\n\nAdds the durable exact-head CI wait record/clear/show lifecycle, stale-head stalled-work evidence, and an advisory report warning for stopped recorded seats while preserving delivery. Includes EN/JA guidance and preview-through-1.x ledger entries.\n\nFocused and full CLI tests plus diff check run locally.


tomohisa:
## Review verdict: request-update — PR #1380 at exact head `2bd744a382956e768c6f4ab7aae915e54b2f4dac`

The design is sound and the slice does what G638 asked for. Build is clean (0 warnings / 0 errors), the full suite is green (4535 Cli + 330 across 10 other projects, 0 failures), CI is green at the exact head with no hash drift, and I verified **every** acceptance criterion by constructing the cases the review-context asks for — they all behave correctly today. Two blocking findings below.

### Verified by constructed cases (all passing)

| AC | Observed |
|---|---|
| Durable wait: PR + observed head + owed transition, cleared on transition | append-only `ci-waits.jsonl` record→read→clear round-trip; `automation pr-transition` clears on applied write |
| No polling | diff grep for `Task.Run`/`Task.Delay`/`Thread.Sleep`/`Timer`/`Process.Start`/`BackgroundService`/`while (true)`/`CancellationTokenSource` → **zero matches** |
| Green + untaken → actionable, naming PR and transition | `ci-all-green-not-transitioned`, `is_informational=false`, `owed_transition` honoured — a wait recording `merge-start` correctly emits `--transition merge-start`, so the transition is *recorded*, not guessed from labels |
| Still-running checks do not fire | `ci-pending`, `is_informational=true`, `stalled=false` |
| Moved head reported as such | `ci-head-moved`, `ci_wait_state=stale-head`, `observed_head_sha` + `current_head_sha`, and **no** `ci_outcome` — it refuses to assert green for a head nobody is waiting on |
| Wait clears; cleared wait produces no further items | after clear, the same fixture stops producing `ci-head-moved` |
| Report to sleeping seat warns, delivers, never refuses | `delivered=true`, `recipient_warning{role, observed_liveness:"not-running"}`, `settle_outcome=unread`, `agent prompt` actually invoked; running seat → no warning; correctly scoped to `report`, so `notify delegate` still fails closed |
| Preview labelling, EN/JA parity, no label vocabulary change | all present |

---

### Blocker 1 — `ci-head-moved` recommends a command its own store rejects

`ci-head-moved` emits:

```
intent-cli automation ci-wait record --domain <d> --repo <r> --pr <n> --head <NEW-HEAD> --transition <t> --write
```

But `CiWaitStore.Record` refuses to overwrite an existing open wait for the same PR when the head differs:

> `PR #N already has an open CI wait for head '<OLD>' and transition '<t>'; refusing to overwrite it with head '<NEW>' / '<t>'.`

Running the emitted command verbatim against the store that produced the item exits **1** with `applied:false`.

**Consequence:** the moved-head item has no working remedy path. Clearing it requires a `ci-wait clear` that the recommendation never names, so every subsequent wake re-reports the same `ci-head-moved` item indefinitely. That is the exact failure mode review-context §6 guards against — *"a detector that keeps reporting a handled state trains its reader to ignore it"* — and it undercuts the AC *"it is durable, it is cleared when the transition is taken."*

**Contract clauses:** In Scope, *"A stale-head guard: if the head has moved since the wait was recorded, the item says so"*; review-context §6, *"The wait must clear."*

**Suggested fix (implementer's choice):** either let `record` idempotently re-point an existing wait when only the head changed — a moved head is the expected case, not a conflict — or emit the `clear`-then-`record` pair in `recommended_action`. Pin whichever you choose with a test.

---

### Blocker 2 — the constructed test cases the packet requires are not in the PR

The Standalone Child Issue Contract requires:

> tests cover the recorded wait and its clearing, **the green-and-untaken case, the still-running case not firing, the moved-head case, and the warning firing and not firing with delivery proceeding in both**

`tests/IntentSystem.Cli.Tests/CiWaitG638Tests.cs` ships 3 tests covering only the store round-trip, the `automation ci-wait` record/clear JSON, and doc-string presence. `AutomationStalledWorkCommandTests.cs` (93 facts) contains zero references to `ci-head-moved`, `CiWaitRecord`, `ci_wait_state`, or `owed_transition`, and `NotifyCommandG578Tests.cs` has no sleeping-recipient warning test.

So the four behaviours that are the entire point of the slice are unpinned. I confirmed by construction that they work today — the gap is regression protection, not correctness. Please add them; the existing `StalledWorkWorkspace` / `FakeLister` / `BuildPr` harness and the `HerdrRoster(...running:false)` helper cover everything needed.

---

### Non-blocking

3. **`ci-head-moved` fires actionable while the new head's checks are still running.** The stale-head check precedes the CI projection, so a moved head with `IN_PROGRESS` checks yields `is_informational=false`. Defensible — the recorded wait is stale regardless — but it is the one path where an actionable class fires on work in progress, which review-context §1 explicitly warns about. Please pin the intended behaviour with a test either way.

4. **Ledger ordering.** The `automation ci-wait` row in `docs/{en,ja}/1.0-compatibility-ledger.md` is inserted after `automation heartbeat`, breaking the table's alphabetical order; it belongs between `automation check` and `automation clarification-stop`.

---

Out-of-scope boundaries were not crossed: no background process, no automatic transition, no label added or altered, delivery semantics / seat identity / settle contract / G629 verdicts unchanged. `Closes #1379` is present and correct (G311). Consistent with `intents/intent-cli/intent-tree/means/08-agent-message-orchestration.md`.

Tests-pass is treated as necessary but not sufficient — hence Blocker 2.

tomohisa:
## Re-review verdict: approve — PR #1380 at exact head `1d0f81ea20010c7a6a2d1ef3a86ef65fa1b2f55d`

Re-reviewed against [the prior blocker comment](https://github.com/J-Tech-Japan/intent-system/pull/1380#issuecomment-5211871024) and the G638 packet. The repair is a single clean commit (`1d0f81ea`) on top of the previously reviewed head — no rewrite of earlier history. **All four findings are resolved, and I verified each by experiment rather than by reading the diff.**

### Blocker 1 — `ci-head-moved` remedy command rejected by its own store → **RESOLVED**

`CiWaitStore.Record` now re-points an open wait when the **same** owed transition observes a new head, keeping the store append-only. I re-ran the exact experiment that previously failed, driving the command surface verbatim:

- Record wait at head `…aaa12` / `review-start`, then run the literal `recommended_action` (`ci-wait record --head …bbb99 --transition review-start --write`) → **exit 0**, wait re-pointed to the new head. Previously this exited 1 with *"refusing to overwrite"*.
- A **different** owed transition with a third head → still **exit 1** and the store is unchanged. The fix does not become a silent overwrite of an unrelated lifecycle obligation, which is the right boundary.
- `--dry-run` re-point → exit 0 with **no mutation** (`observed_head` still the old head).
- `ClearForTransition` after a re-point → clears cleanly, `ReadOpen` empty. The append-only log stays coherent across record → re-point → clear.
- End-to-end loop closes: `Execute_DurableWait_MovedHeadIsActionableUntilTheWaitIsRepointed` shows `ci-head-moved` (actionable) → apply the remedy → `ci-pending` (informational, `stalled=false`). The detector no longer re-reports a handled state, satisfying review-context §6.

Same-head/same-transition still short-circuits to `already_converged`, so the re-point branch did not make the idempotent case lossy.

### Blocker 2 — packet-required test coverage → **RESOLVED**

All four enumerated cases are now pinned, and all pass:

| Required case | Test |
|---|---|
| green-and-untaken naming the recorded owed transition | `Execute_DurableWait_GreenAndUntaken_UsesTheOwedTransition` — asserts `owed_transition=repair-pushed` and `--transition repair-pushed`, i.e. a **non-default** transition, so it genuinely pins "recorded, not inferred" |
| still-running not firing | second half of `Execute_DurableWait_MovedHeadIsActionableUntilTheWaitIsRepointed` — `ci-pending`, `stalled=false` |
| moved head | first half of the same test — `ci-head-moved`, `stale-head`, both SHAs, **no** `ci_outcome`, remedy names the new head |
| warning firing / not firing, delivery proceeding in both | `Report_ProceedsWithAdvisoryWarning_WhenRecordedRecipientSeatIsStopped_G638` (`delivered=true`, `seat-not-running`, `unread`, `agent prompt wH:p1` invoked) + `Assert.False(TryGetProperty("recipient_warning"))` added to the running-seat case |
| store re-point / conflict boundary | `Record_RepointsAnOpenWaitWhenTheSameTransitionSeesANewHead` |

Using `repair-pushed` rather than `review-start` in the green-and-untaken fixture is a good choice — it would fail against a hard-coded default.

### Non-blocking 3 — moved head while checks still run → **RESOLVED (behaviour decided and pinned)**

The moved-head test deliberately uses `CheckRun("IN_PROGRESS")` and asserts the item is actionable `ci-head-moved`. The intended semantics are now explicit: a stale wait is a stale wait regardless of the new head's check state, and it has a working remedy. That is a defensible reading of review-context §1 and it is no longer implicit.

### Non-blocking 4 — ledger ordering → **RESOLVED**

`automation ci-wait` now sits between `automation check` and `automation clarification-stop` in **both** `docs/en/` and `docs/ja/1.0-compatibility-ledger.md`.

### Verification at this head

- `dotnet build IntentSystem.sln -c Release` → **0 warnings, 0 errors**.
- `dotnet test IntentSystem.sln -c Release --no-build` → **green: 4539 passed / 1 pre-existing skip in Cli, plus 330 across 10 other projects. 0 failures.** (+4 tests vs the previous head.)
- Exact-head CI: `Build and test (source contract)` → success with `head_sha == 1d0f81ea…`. **No hash drift** — the green signal is bound to the reviewed commit.
- `git diff --check` → clean.
- Diff grep over the repair commit for `Task.Run`/`Task.Delay`/`Thread.Sleep`/`Timer`/`Process.Start`/`BackgroundService`/`while (true)`/`CancellationTokenSource` → **zero matches**. The repair added no poller; the new branch is a single `File.AppendAllText`.
- Guidance remains accurate: the EN/JA docs describe the moved-head remedy as `ci-wait record … --head <new>`, which is now exactly what works.

### Packet conformance

Execution unit **G638**, packet `.intent-cli/issues/G638/`. Every acceptance criterion maps to a concrete change **and** a test: durable record/clear without polling, green-and-untaken actionable naming PR and transition, still-running not firing, moved-head guard, advisory sleeping-recipient warning that never refuses, EN/JA guidance parity under G613, and preview-through-1.x labelling per G628.

Out-of-scope boundaries were **not** crossed: no background process, no automatic transition, no label added or altered, and delivery semantics / seat identity / settle contract / G629 verdicts are unchanged (`notify delegate` still fails closed with `agent-not-running`; the stopped-seat allowance is scoped to `report` only). Consistent with `intents/intent-cli/intent-tree/means/08-agent-message-orchestration.md` — it records the one transition no message carries instead of adding a new transport. `Closes #1379` present and correct (G311). Base branch `main` matches the packet's `direct-main` policy.

Tests-pass is treated as necessary but not sufficient; the approval rests on the packet/intent conformance evidence above, with the two blockers independently re-verified by experiment.

**Approved.**

PR #1388, opened 2026-08-07
G641: measure bounded supervision recovery

### LINKED ISSUE #1387: G641: How long a loop may stay stopped
# G641: How long a loop may stay stopped

## Goal

Make **time-to-recovery a bounded, measured property of the tool** rather than a function of whether a human happens to look. One supervision entry point covers every stall class the team knows how to detect, declares the interval within which a stall will be noticed, wakes the owning role through the recorded transport, and **records how long each stall actually lasted** so the bound is checked against reality instead of assumed.

## Why This Slice Exists Now

**On 2026-08-06 this team's loop stopped silently three times, each with a different cause, and a human noticed every time:**

| cause | what was true | who noticed |
|---|---|---|
| recipient died mid-task | delivered, working observed, process gone | operator asked why nothing moved |
| checks went green | orchestrator deferred while CI ran; no message follows a finished run | operator asked again |
| report reached a sleeping seat | delivery succeeded, nobody read it | design read a pane directly |

**The longest measured gap was about two hours fifteen minutes**, between the last orchestration event and the operator asking. Nothing in the system knew that time had passed.

**The pieces exist and do not add up to a mechanism.** `notify supervise` watches recipients, `automation ci-wait` plus the `ci-all-green-not-transitioned` class covers finished checks, `stalled-work` covers queue and write-back debt, and a seat can go absent entirely. Four detectors, no single loop, **no declared bound**, and **no record of how long anything was actually stopped**.

**Three teams independently wrote their own watchdogs** — this host, the Sekiban team, and the Zero4Racer team — which is the same signal that produced G629 and G638: when every team builds the same missing part outside the tool, the part belongs inside it. This host's own watchdog script existed but **was not running**, and nothing noticed that either, which is the failure one layer up.

**This matters most for the agent kinds the project is opening up to.** Copilot seats on this host wedged repeatedly, and the Sekiban team hit an unresponsive session after heavy use. Supporting Copilot, opencode, Cursor and the rest is only real if a wedged seat costs minutes without a human, rather than however long until someone looks. **Broad agent support and unbounded recovery time are incompatible.**

**One loop, every known class.** Supervision that covers only recipients leaves teams writing scripts for the rest, which is the state this slice ends. It extends `notify supervise` rather than adding a fifth surface.

**A declared bound, and evidence against it.** The team declares how long a stall may go unnoticed; the command reports whether the loop is meeting that bound and records the measured duration of each detected stall. **An unmeasured promise is the thing being replaced, so this slice must not add another one.**

**Wake, do not decide.** On detection the owning role is woken through the transport its topology records. Taking the transition stays a judgment for the role that owns it — the same line G630 drew for re-dispatch.

**A fourth cause, measured while this packet was being written**: `notify escalate` appends an event to the durable log but **does not wake the recipient's pane** — its own output says `delivered: false`. An escalation therefore records a fact nobody is told. On 2026-08-07 design sent a release approval by escalate, because the report path was blocked, and orchestration sat idle for thirty-five minutes with the approval already recorded. **A message that is durable and unheard is the same stall in a new place**, so the loop must treat an escalation whose delivery was false as a condition to surface.

**Not running is a stall too.** The loop reports its own liveness, because a supervisor that silently stops is indistinguishable from a healthy system, and that is exactly what happened here.

## Current Observed State

- `notify supervise` covers recipient liveness only.
- `ci-all-green-not-transitioned`, queue debt and seat absence are detectable but only by separate invocations a team must arrange itself.
- No bound is declared anywhere, and no duration is recorded for any stall.
- This host's watchdog script existed unrun for an unknown period; nothing reported its absence.

- G630's supervision loop, its silence on healthy seats, its verified recovery sequence and its opt-in re-dispatch are correct and stay.
- G638's `ci-all-green-not-transitioned` class and `ci-wait` records are correct and are consumed here, not redefined.
- `stalled-work`'s existing classes are correct and are consumed as they are.
- Seat identity, delivery semantics and the settle contract are unchanged.

## Target Repo / Path / Part

Target part: the supervision loop's coverage, its declared bound, and its recovery-duration record.

## In Scope

- **One loop covering every known class**: recipient lost, checks settled with the owed transition untaken, queue and write-back debt, and a seat absent from the workspace entirely. Healthy state stays silent.
- **A declared detection bound** the team records, with the loop reporting whether its actual interval satisfies it. A bound that cannot be met is reported as unmet rather than quietly missed.
- **A recovery-duration record**: for each detected stall, when it became detectable, when it was surfaced, and when it cleared, so time-to-recovery is a measured number the team can read back. **The record must show its own gaps** — a stall detected only after the loop restarted is recorded with an unknown start rather than a flattering one.
- **Waking through the recorded transport**: the owning role is woken by the transport its topology declares; the loop never invents a transport nor hard-codes one agent kind's mechanism.
- **An undelivered escalation is a detectable condition**: an escalation recorded with delivery false means the fact is durable and nobody was told, and the loop surfaces it like any other stall.
- **Self-liveness**: the loop reports when it is running and, on start, whether it was absent since the last recorded cycle. A supervisor that is not running is reported as a finding, not as silence.
- **Agent-kind neutral throughout**: nothing in the loop assumes a particular agent; per-kind facts come from the recipe as G636 established.

## Out Of Scope

- Taking any owed transition automatically; the loop wakes and records.
- Inventing stall classes without measured evidence — this slice consumes the classes that exist.
- Changing G630's recovery sequence, its re-dispatch default, delivery semantics, seat identity or the settle contract.
- Replacing a team's existing external watchdog by force; adoption is the team's choice.

The child PR extends the supervision loop so a single invocation covers every stall class the tool can already detect, namely a lost recipient, checks that settled while the owed transition was not taken, queue and write-back debt, and a seat absent from the workspace, while remaining silent when nothing is wrong. The team records a detection bound and the loop reports whether its actual interval satisfies that bound, reporting an unmet bound explicitly rather than missing it quietly. For each detected stall the loop records when the condition became detectable, when it was surfaced and when it cleared, so that time to recovery is a measured value, and where the loop cannot know when a condition began it records an unknown start rather than an optimistic one. On detection the owning role is woken through the transport recorded in that team's topology, without inventing a transport or hard-coding any one agent kind's mechanism, and no owed transition is taken automatically. The loop reports its own liveness including, at start, whether it was absent since the last recorded cycle, so that a supervisor which is not running is surfaced as a finding rather than as silence. Nothing in the loop assumes a particular agent kind and per-kind facts continue to come from the launch recipe. G630's recovery sequence and its opt-in re-dispatch default, the existing stall classes, delivery semantics, seat identity and the settle contract are unchanged, no stall class is invented without measured evidence, the surface is documented as preview under the compatibility promise's preview lane, EN and JA stay in parity under the G613 terminology policy, and the full suite is green.

## Acceptance Criteria

- One invocation covers all four classes; silent when healthy.
- Declared bound recorded; loop reports whether its interval satisfies it; an unmet bound is reported.
- Per-stall record of detectable-at, surfaced-at and cleared-at; unknown start recorded as unknown.
- Wake goes through the topology's recorded transport; no transport invented, no agent kind hard-coded.
- Loop reports its own liveness and its absence since the last recorded cycle.
- No automatic transition; G630's sequence and defaults unchanged; existing classes consumed, none invented.
- Preview labelling; ja/en parity under G613; full suite green.

## Verification

- Construct each of the four conditions and assert exactly one loop invocation surfaces each, and that a healthy team produces no output.
- Set a bound the loop cannot meet and assert it reports the bound as unmet.
- Clear a stall and assert the record carries all three timestamps; construct a stall that predates the loop and assert an unknown start.
- Stop the loop, restart it, and assert the absence since the last cycle is reported.
- Change a team's recorded transport and assert the wake follows it.

## Related Links

- Measured on this host, 2026-08-06: three silent stalls with three distinct causes, a human noticing each time, and a longest observed gap of about two hours fifteen minutes between the last orchestration event and the operator asking.
- This host's `design-watchdog.sh` existing but not running, with nothing reporting its absence.
- Watchdogs written independently by three teams; the Sekiban report of an unresponsive session after heavy use; repeated Copilot seat wedges on this host.
- G630 — the loop this extends; G638 — the CI class it consumes; G636 — per-kind facts come from the recipe; G628 — the preview lane.

## Knowledge Maintenance

- Intent placement: node 08 primary (supervision and recovery), node 03 supporting (measured durations as durable evidence).
- ADR candidate: none.
- Diagram candidate: none.
- Docs update: ja/en supervision guidance (in scope).
- Closeout writeback expected: yes — node 08 records that time-to-recovery must be bounded and measured rather than promised, that a supervisor which is not running is a stall of its own, and that broad agent-kind support depends on recovery not requiring a human.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1387

## Summary
- extend `notify supervise` with one measured loop over recipient, terminal-CI, queue/write-back, undelivered-escalation, and recorded-seat conditions
- record declared detection bounds, cycle gaps, self-liveness, and per-stall detectable/surfaced/cleared recovery evidence
- wake the owning logical role through the recorded transport without applying owed transitions; preserve G630 recovery semantics
- document the preview-through-1.x contract in English and Japanese and add constructed recovery-bound tests

## Verification
- `dotnet build src/IntentSystem.Cli/IntentSystem.Cli.csproj --no-restore`
- G630 focused tests: 8 passed
- G641 focused tests: 3 passed
- G613 terminology focused tests: 6 passed
- full CLI suite: 4543 passed, 1 skipped, 0 failed
- `git diff --check`


tomohisa:
## Review verdict — G641 / PR #1388 @ `3330ec0` — **REQUEST-UPDATE**

Reviewed at the exact head named in the review task: `3330ec0123e8d2bdb0ae4058897e55f48843eccf`.

Most of this slice is in good shape, and I verified it **by construction** rather than by reading — I wrote a throwaway probe suite against the real `NotifyMeasuredSupervisor`, captured its output, and then deleted it (the tree is unmodified). **Two acceptance criteria do not hold**, and both are the specific failure mode this packet exists to prevent: the loop staying quiet when it should speak, and the loop making a claim it has not measured.

The full suite is green (**4543 passed, 1 skipped, 0 failed**) and `git diff --check` is clean. Per the review contract, a green suite is necessary but not sufficient — both findings below are invisible to the shipped tests.

---

## 🔴 Finding 1 — self-absence is inert unless `--bound` is declared (implementation-finding)

**Failed AC:** *"Loop reports its own liveness and its absence since the last recorded cycle."*
**Failed In-Scope clause:** *"A supervisor that is not running is reported as a finding, not as silence."*

`--bound` is optional (`NotifyCommand.cs` validates it only `is not null`) and no default is ever applied. `NotifyMeasuredSupervisor.DefaultDetectionBoundSeconds = 300` is declared but **never referenced anywhere in the repository** — the intended default appears to have been left unwired.

Because `NotifyMeasuredSupervisor.cs:106` defines absence purely as a bound breach:

```csharp
var absentSinceLastCycle = boundMet is false;
```

…and `boundMet` is `null` when no bound exists, a supervisor that stops for hours reports nothing. Constructed evidence — a team with **no declared bound**, supervisor absent for **3 hours**:

```
PROBE2 gapSeconds=10800  absentSinceLastCycle=False
PROBE2 boundRecorded=False  boundSeconds=(null)  boundStatus=unrecorded  boundMet=(null)
PROBE2 supervisorNotRunningFinding=False
PROBE2 silent=True
```

The loop **measured the 10800s gap correctly** and then declared itself healthy. This is precisely the motivating incident in the packet: *"This host's `design-watchdog.sh` existing but not running, with nothing reporting its absence."* As written, the fix cannot detect its own absence in the default configuration.

**It also states something it has not measured.** In the same run, `Liveness.Summary` reads:

```
Supervision is running; the measured 10800s cycle gap is within the declared bound.
```

There is no declared bound (`boundStatus=unrecorded`). The `:` branch at `NotifyMeasuredSupervisor.cs:117` interpolates that sentence unconditionally. The packet is explicit that *"An unmeasured promise is the thing being replaced, so this slice must not add another one."*

**Required unblock action:** report absence against a threshold that always exists — the loop's own `intervalSeconds` is already available and is already persisted as `IntervalSeconds` on the cycle record — or apply `DefaultDetectionBoundSeconds` when nothing is declared. Either way, `absent_since_last_cycle` and the `supervisor-not-running` finding must fire without a team having opted into `--bound`, and the liveness sentence must not claim a bound that was never declared.

---

## 🔴 Finding 2 — an `undelivered-escalation` finding never clears, so the loop never returns to silence (implementation-finding)

**Failed AC:** *"One invocation covers all four classes; **silent when healthy**."*
**Failed review lens #3:** *"a supervisor that chatters is one nobody reads, which is how the original failure happened."*

`ReadUndeliveredEscalations` derives its observation key from each `kind: escalation` line in `events.jsonl`. That log is an **append-only durable event channel** — escalation lines are never removed in normal operation. A stall only clears when its key disappears from the observation set, so the key persists forever. Once the wake is delivered, `RunOnce` short-circuits and re-emits the record as a finding on every subsequent cycle:

```csharp
if (existing is not null && existing.WakeDelivered)
{
    records.Add(existing with { Summary = observation.Summary });
    findings.Add(ToFinding(records[^1]));
    continue;
}
```

Constructed evidence — one escalation, successfully woken and delivered on cycle 1, nothing else wrong, three consecutive cycles:

```
PROBE6 cycle1 silent=False findings=1 delivered=True
PROBE6 cycle2 silent=False findings=1
PROBE6 cycle3 silent=False findings=1
PROBE6   cycle3 kind=undelivered-escalation wakeDelivered=True
```

After a single `notify escalate`, this team's supervision is **permanently non-silent**, re-waking the owning role about the same already-acknowledged escalation on every cycle forever. That is the chatter failure lens #3 warns about, and it also means the healthy-silence AC holds only until a team's first escalation.

Note the shipped test clears this condition with `File.Delete(ResolveEventPath())` — deleting the entire durable event log. That is not a mechanism available in real operation, and it is what hides this behaviour from the suite.

**Required unblock action:** give the escalation class a real clearing/acknowledgement condition — e.g. treat an escalation as cleared once its wake was delivered (or once a later event acknowledges/answers that unit) — so that a delivered, acknowledged escalation stops being re-reported and the loop returns to silence.

---

## ✅ What I verified and found correct

| AC / lens | Verdict | Constructed evidence |
|---|---|---|
| Healthy team is silent (fresh team) | ✅ | `PROBE1 silent=True findings=0 actions=0` — recorded topology, both seats running |
| One invocation covers the classes | ✅ | Constructed `ci-all-green-not-transitioned` through the loop with an injected candidate lister: `PROBE5 kind=ci-all-green-not-transitioned source=automation-stalled-work`. `seat-absent` covered by the shipped test; `recipient-lost` is consumed unchanged from G630's supervisor; queue/write-back debt flows through the identical `automation-stalled-work` path (only `ci-pending` is filtered) |
| Unmet bound is reported | ✅ | Shipped test: `BoundMet=false`, `GapSeconds=301`, `supervisor-not-running` — correct **when a bound is declared** (see Finding 1) |
| Three timestamps + measured duration | ✅ | `PROBE3 detectableAt=12:10 surfacedAt=12:15 clearedAt=12:25 durationSeconds=900 unknown=False` — this path was **not** covered by the shipped tests, which only exercise the unknown-start variant |
| Unknown start recorded as unknown | ✅ | Shipped test asserts `DetectableAtUnknown` with `DurationSeconds=null`; `PROBE5` confirms a first-cycle detection records `detectableAt=(null)` rather than a flattering start |
| Wake follows the **recorded** transport | ✅ | Same stall, two recorded transports: agmsg → `bash send.sh intent-cli-dev supervisor orchestration {...}`; herdr-only → `fake-herdr agent prompt wG641:p1 {...} --wait --until working`. No `send.sh` call in the herdr run. Transport is resolved via `SessionLayerModeStore.Resolve`, not invented |
| No automatic transition | ✅ | Grep over `NotifyMeasuredSupervisor.cs` + `NotifySupervisionStore.cs` for `pr-transition`/label/merge/queue-state mutation → none. Wake payload carries `must_transition: false` |
| No agent kind hard-coded | ✅ | Grep for `claude|codex|copilot|cursor|opencode|gemini` in the new loop/store → none |
| Preview labelling; EN/JA parity | ✅ | Both docs carry `G641 — preview-through-1.x`, the preview block, and the compatibility-ledger link; EN/JA are section-by-section equivalent with English technical terms retained per G613 |
| Suite green; `git diff --check` | ✅ | 4543 passed / 1 skipped / 0 failed; `git diff --check` exit 0 |

## Supporting note on test coverage

Both findings are invisible to the suite because of gaps the packet's own Verification section asks for:

- Every shipped test constructs the supervisor with `repo: null`, so the **CI and queue/write-back classes are never exercised** — I had to inject a candidate lister to reach them.
- `HealthyRecordedSeatsRemainSilentAndAbsentSeatsAreConstructedFromTopology_G641` is named for healthy silence but supplies an **empty agent roster**, so every recorded seat is absent; it asserts `seat-absent` findings and never asserts silence. **No shipped test asserts the healthy-silence AC.**
- The only "clear" path deletes the whole event log, which masked Finding 2.

Adding tests for the two fixes — absence with no declared bound, and steady-state silence after a delivered escalation — would close these.

---

Nothing here is a host-metadata concern; both findings are code gaps fixable on this branch. Happy to re-review on the updated head.

tomohisa:
# Re-review: REQUEST-UPDATE — PR #1388 @ `de0a1429156a92525a45df7fdb5c37ae34c71994`

Both previously reported blockers are **fixed and independently re-verified**. The fix for blocker 1, however, **introduces a new regression in the same acceptance criterion it was meant to serve**, so this cannot be approved yet.

All findings below are **implementation-findings** (code gaps fixable on this branch). No host-metadata findings.

Verified by construction with a throwaway probe suite written against the real `NotifyMeasuredSupervisor`, run, and deleted — the working tree was left byte-identical to the PR head (`git status --porcelain` empty, `git diff HEAD` empty).

---

## ✅ Prior blocker 1 — self-absence without `--bound` — FIXED

No declared bound, 3-hour gap:

```
gap=10800  absent=True  silent=False  boundMet=null  recorded=False
threshold=300  kind=configured-interval
liveness: "Supervision restarted after a 10800s gap, exceeding the configured 300s
           supervision interval; no detection bound was declared."
finding:  supervisor-not-running — "Supervision was absent for 10800s, beyond the
           configured 300s supervision interval; no detection bound was declared."
```

The absence now fires, `bound_met` stays `null`, and no output claims a bound that was never declared. `absence_threshold_kind` correctly distinguishes `configured-interval` from `declared-bound`. The dead `DefaultDetectionBoundSeconds` constant is gone. This is the right shape.

## ✅ Prior blocker 2 — `undelivered-escalation` never cleared — FIXED

Re-verified **without** deleting the append-only event log (the mechanism the earlier test used to fake a clear):

```
cycle1 silent=False escalationFindings=1
cycle2 silent=True  escalationFindings=0 eventLogExists=True
cycle3 silent=True  escalationFindings=0 eventLogExists=True
cycle4 silent=True  escalationFindings=0 eventLogExists=True
cycle5 silent=True  escalationFindings=0 eventLogExists=True
```

Acknowledgement is derived from `stalls.jsonl` (`WakeDelivered`), which is uncapped and append-only, so it is durable across restarts. I also confirmed the filter does **not** over-suppress: the key is `escalation:{unit}:{timestamp-ticks}`, and a genuinely new escalation appended after an acknowledged one still surfaces exactly once.

---

## 🔴 NEW BLOCKER — the no-bound fallback makes a healthy loop chatter on every cycle

### Failed acceptance criterion

> "One invocation covers all four classes; **silent when healthy**."

and review-context lens 3:

> "Run against a healthy team across several cycles and assert no output. G626 and G630 both turn on this; **a supervisor that chatters is one nobody reads, which is how the original failure happened.**"

### Evidence

The fallback threshold is the configured interval **exactly**, compared with a strict `>` and no grace:

```csharp
var absenceThresholdSeconds = bound.BoundSeconds
    ?? (previousCycle?.IntervalSeconds is > 0 ? previousCycle.IntervalSeconds : intervalSeconds);
var absentSinceLastCycle = gapSeconds is { } gap && gap > absenceThresholdSeconds;
```

But the measured gap is **structurally always larger than the interval**:

- `NotifySupervisionCycle.CompletedAt = now` is stamped at cycle **start** (`StartedAt` and `CompletedAt` are both assigned the same `now` captured on entry to `RunOnce`), so the whole cycle's work lands *after* the timestamp it is measured from.
- `RunLoop` calls `NotifySupervisor.Delay(TimeSpan.FromSeconds(intervalSeconds))` **after** the pass and after `EmitSupervision`, and `Thread.Sleep` guarantees *at least* the requested duration.

So `gap = intervalSeconds + cycle-work + sleep-overshoot`, while the threshold is `intervalSeconds`. **The entire per-cycle compute budget before a false alarm is under one second.**

Boundary pinned exactly (no bound, interval 300, fresh store per case):

| gap | absent | silent | exit |
|---|---|---|---|
| 299 | False | True | 0 |
| 300 | False | True | 0 |
| **301** | **True** | **False** | **1** |

Faithful simulation of `RunLoop` against a **completely healthy team** — both topology seats present, `interactive_ready: true`, no stalls, no escalations, no queue debt — with a modelled 1.2s `herdr agent list` subprocess per cycle:

```
PROBEI cycle1 at=12:00:01 gap=null absent=False silent=True  exit=0 findings=[]
PROBEI cycle2 at=12:05:06 gap=301  absent=True  silent=False exit=1 findings=[supervisor-not-running]
PROBEI cycle3 at=12:10:07 gap=304  absent=True  silent=False exit=1 findings=[supervisor-not-running]
PROBEI cycle4 at=12:15:08 gap=301  absent=True  silent=False exit=1 findings=[supervisor-not-running]
```

Every cycle after the first emits `supervisor-not-running` — a **false claim that supervision was absent, emitted by a supervisor that never stopped running**. The loop is never silent again.

### Why this matters beyond the AC

1. **This is a regression, not a pre-existing gap.** At the previous head the no-bound path was inert (`absentSinceLastCycle = boundMet is false`, and `boundMet` was `null`), so healthy teams were silent. The fix moved the behaviour from *never fires* to *always fires*. Both are wrong.
2. **It is the default path.** `--bound` remains optional, and the packet's own Current Observed State says "No bound is declared anywhere" — so today every team lands on this fallback.
3. **The exit code compounds it.** This PR added `&& !Findings.Any(f => f.Kind == "supervisor-not-running")` to `NotifySupervisorPass.ExitCode`, so a healthy `notify supervise --once` now exits **1** every run. Any wrapper or watchdog treating non-zero as failure will read a healthy loop as broken — the exact "watchdog nobody trusts" failure this slice exists to end.
4. **It re-creates the alarm-fatigue failure the packet cites** as the reason the original stalls went unnoticed.

A generously declared bound is unaffected (`--bound 900`, interval 300, gap 301 → `absent=False`, `silent=True`), which confirms the defect is specific to the fallback threshold rather than the absence logic in general.

### Required unblock action

The absence threshold must be strictly greater than the cadence the loop actually runs at. Any of these would satisfy the AC:

- give the fallback headroom over the interval (e.g. a multiple such as `2 × interval`, or `interval + grace`) rather than the bare interval; and/or
- stamp `CompletedAt` at genuine cycle completion so `gap_seconds` measures the true inter-cycle idle gap instead of silently including one cycle's own duration; and/or
- reinstate an explicit default detection bound and take `max(default, headroom × interval)`.

Please also add a regression test that runs a **healthy** team across at least three cycles at the realistic cadence (gap ≥ interval, no `--bound`) and asserts `Silent` on every cycle. The two tests added in this commit both avoid this path: `RestartGapReportsSelfAbsenceWithoutClaimingUndeclaredBound_G641` only exercises a 3-hour gap, and `DeliveredEscalationClearsAndReturnsToSilence_G641` advances the clock by 1s and 2s, well under the interval. No shipped test asserts healthy silence across a normal cadence, which is why a green suite did not catch this.

### Secondary accuracy note (same root cause)

`gap_seconds` is reported as the gap since the last cycle "completed", but it is measured from the previous cycle's **start**, so every reported gap is inflated by one full cycle's duration. For a slice whose deliverable is a *measured* property, this number should measure what its name claims.

---

## Everything else re-verified at this head

| Check | Result |
|---|---|
| Head `de0a1429` | ✅ matches re-review envelope |
| Full suite | ✅ 4545 passed / 1 skipped / 0 failed |
| `git diff --check` | ✅ clean |
| CI at exact head | ✅ `Build and test (source contract)` success, `head_sha` verified |
| `Closes #1387` ↔ linked issue | ✅ matches (G311) |
| Multiple classes in one invocation | ✅ `[seat-absent, undelivered-escalation]` co-surfaced in a single pass |
| Unknown start recorded as unknown | ✅ `detectableUnknown=True` on records predating the loop |
| Three timestamps + measured duration | ✅ `detectable_at` / `surfaced_at` / `cleared_at` recorded; duration measured only for known starts |
| Unmet **declared** bound reported | ✅ `boundMet=False`, `actual=600`, `kind=declared-bound` |
| Wake follows recorded transport | ✅ agmsg → `bash send.sh …`; herdr topology → `fake-herdr agent …`; neither transport leaked into the other |
| No automatic transition | ✅ grep clean; wake payload carries `"must_transition": false` |
| No agent kind hard-coded | ✅ grep clean |
| Preview labelling + EN/JA parity | ✅ both documents updated in parity under G613, preview-through-1.x retained |

## False-claim boundaries

State plainly what was **not** verified and must not be claimed:

- The healthy-cadence behaviour above was reproduced with an **injected clock and a modelled 1.2s subprocess cost**, not against a live long-running `notify supervise` process on real hardware. The structural argument (threshold == interval, `Delay` after the cycle, `CompletedAt` at cycle start) is verified from source and the boundary is pinned at gap 300 vs 301; the specific per-cycle wall-clock cost of a production cycle was **not** measured.
- Real GitHub-backed `ci-all-green-not-transitioned` and queue-debt classes were **not** re-constructed at this head; that path is untouched by this commit's diff and was constructed and verified at the previous head.
- No claim is made that any real supervision loop has run in production under this build.

tomohisa:
# Re-review (pass 3): NO FINDINGS — PR #1388 @ `5cc8a2e27f41438237cf2caaa2ac5d79da3e7bba`

**All three previously reported blockers are resolved and independently re-verified by construction.** I found no implementation findings in this pass.

Approval is withheld for one reason only: **the PR is still a draft**, and promotion is host-scoped. See "Draft routing" at the end.

Verified with a throwaway probe suite written against the real `NotifyMeasuredSupervisor`, run, and deleted — the working tree was left byte-identical to the PR head (`git status --porcelain` empty, `git diff HEAD` empty).

---

## ✅ Blocker 3 (the cadence regression) — FIXED

`FallbackAbsenceThresholdSeconds(cadence) = max(2 × cadence, cadence + 60)` gives the fallback real headroom, and `CompletedAt` is now stamped at genuine completion rather than at cycle entry.

**Healthy no-bound supervision, five realistic cycles, monotonic clock, 1.5s of herdr subprocess work per cycle, `Delay(interval)` applied after each pass — exactly the scenario that failed at `de0a142`:**

```
P1 cycle1 start=12:00:01 gap=null cadence=300 threshold=600 kind=configured-interval absent=False silent=True exit=0 findings=[]
P1 cycle2 start=12:05:03 gap=300  cadence=300 threshold=600 kind=configured-interval absent=False silent=True exit=0 findings=[]
P1 cycle3 start=12:10:04 gap=300  cadence=300 threshold=600 kind=configured-interval absent=False silent=True exit=0 findings=[]
P1 cycle4 start=12:15:06 gap=300  cadence=300 threshold=600 kind=configured-interval absent=False silent=True exit=0 findings=[]
P1 cycle5 start=12:20:07 gap=300  cadence=300 threshold=600 kind=configured-interval absent=False silent=True exit=0 findings=[]
```

**Completion-to-start gaps at or above the configured interval, four cycles each, no `--bound`:**

| extra post-completion work | observed gap | all cycles silent |
|---|---|---|
| +0s | 300 | ✅ |
| +5s | 305 | ✅ |
| +50s | 350 | ✅ |
| +250s | 550 | ✅ |
| +299s | 599 | ✅ |

## ✅ Detection is not disabled — a genuine long gap still triggers

The headroom did not buy silence by turning self-absence off. Boundary pinned exactly:

| gap | absent | silent | exit | finding |
|---|---|---|---|---|
| 599 | False | True | 0 | — |
| 600 | False | True | 0 | — |
| **601** | **True** | **False** | **1** | `supervisor-not-running` |
| **1800** | **True** | **False** | **1** | `supervisor-not-running` |
| **10800** | **True** | **False** | **1** | `supervisor-not-running` |

```
Supervision was absent for 10800s, beyond the configured 300s cadence's 600s
self-absence threshold; no detection bound was declared.
```

A supervisor that misses two full beats is caught; one that is merely doing its work is not. `--interval` is capped at 3600, so `2 × cadence` cannot overflow.

## ✅ Blocker 1 (false declared-bound claim) — still fixed

At both a healthy gap and a 3-hour gap, with no `--bound`: `bound_seconds=null`, `bound_met=null`, `absence_threshold_kind=configured-interval`, and every summary carries `no detection bound was declared`. Neither `within the declared` nor `exceeding the declared` appears. A declared bound still governs when present (`--bound 60`, gap 600 → `bound_met=False`, `threshold=60`, `kind=declared-bound`), so the fallback never shadows an explicit bound.

## ✅ Blocker 2 (escalation never cleared) — still fixed

Verified **without** deleting the append-only event log, and checked for over-suppression:

```
P5 cycle1 silent=False escalations=1  (surfaced + wake delivered)
P5 cycle2 silent=True  escalations=0 eventLogLines=1
P5 cycle3 silent=True  escalations=0 eventLogLines=1
P5 cycle4 silent=True  escalations=0 eventLogLines=1
P5 cycle5 silent=True  escalations=0 eventLogLines=1
P5 newEscalation fired=1 silent=False eventLogLines=2
```

The log is never truncated, the acknowledged key stays suppressed, and a genuinely new escalation still surfaces exactly once.

---

## Acceptance criteria — packet `G641` (`packet.yaml` target_part; ACs from `github-body.md`)

| AC | Evidence |
|---|---|
| One invocation covers all four classes; **silent when healthy** | `[seat-absent, undelivered-escalation]` co-surfaced in a single pass; healthy team silent across 5 cycles (above). CI/queue classes flow through the same generic observation pipeline, untouched by this commit and constructed at an earlier head. |
| Declared bound recorded; loop reports whether its interval satisfies it; unmet bound reported | `--bound 60`, gap 600 → `bound_met=False`, `actual=600`, `kind=declared-bound` |
| Per-stall `detectable_at` / `surfaced_at` / `cleared_at`; unknown start recorded as unknown | Known start: `detectableAt=12:03:20 unknown=False surfacedAt=12:05:00 clearedAt=12:20:00 duration=1000`. Unknown start: `detectableUnknown=True`, cleared with `duration=null` — no flattering duration. |
| Wake through the topology's recorded transport; none invented; no agent kind hard-coded | agmsg topology → `bash send.sh intent-cli-dev supervisor orchestration {…}`; herdr topology → `fake-herdr agent …`. Neither transport leaked into the other. Grep for `"claude"/"codex"/"copilot"/"cursor"/"opencode"/"gemini"` in the supervision sources: clean. |
| Loop reports own liveness and absence since last cycle | `running`, `gap_seconds`, `cadence_interval_seconds`, `absence_threshold_seconds`, `absence_threshold_kind`, `absent_since_last_cycle` all emitted every cycle |
| No automatic transition; G630 sequence and defaults unchanged; existing classes consumed | Grep for `pr-transition` / `gh pr merge` / label mutation in the supervision sources: clean. Wake payload carries `"must_transition": false`. G630's recovery runs unchanged via the legacy `NotifySupervisor` pass. |
| Preview labelling; ja/en parity under G613; full suite green | Both `docs/{en,ja}/12-agent-message-orchestration.md` and both compatibility ledgers updated in parity, `preview-through-1.x` retained, headroom formula documented in both languages |

**Verification steps from the issue body** are all runnable and pass, including the two named greps and `git diff --check` (clean).

**Out-of-scope boundaries were not crossed.** The diff touches only `src/IntentSystem.Cli/**`, `docs/en/**`, `docs/ja/**`, `tests/**` — exactly the packet `target_path`. No owed transition is taken automatically, no stall class is invented, and G630's recovery sequence, re-dispatch default, delivery semantics, seat identity and settle contract are untouched.

**Intent reference.** Consistent with `intents/intent-cli/intent-tree/means/08-agent-message-orchestration.md` (supervision and recovery) and `…/03-state-and-audit-strategy.md` (measured durations as durable evidence): the loop records measured evidence and wakes, and never decides.

**Tests.** `dotnet test tests/IntentSystem.Cli.Tests/IntentSystem.Cli.Tests.csproj` → **4546 passed, 1 skipped, 0 failed**. CI `Build and test (source contract)` is **success at the exact head** `5cc8a2e2` (run `31172078982`). Tests-pass is treated as **necessary but not sufficient** — every AC above is paired with constructed evidence, and each of the three prior blockers was reproduced independently rather than accepted from the fix commit. Notably, the previous head was also fully green while carrying a defect that made a healthy loop chatter forever.

**Closing reference.** `Closes #1387` matches the linked source issue (G311). `review closeout-plan` → `ready: true, gaps: []`.

---

## Draft routing (not a finding)

`automation host-review-diagnostics --pr-draft true --draft-review-ready` classifies this PR as **`draft-ready-to-promote`**:

> PR #1388 is draft but otherwise review-ready with no findings and not operator-intended (G376). Mark it ready for review and continue approval/merge/closeout.

Recommended next command:

```
gh pr ready 1388 --repo J-Tech-Japan/intent-system && \
intent-cli automation host-review-preflight --repo J-Tech-Japan/intent-system --format json
```

That chain is host-scoped, and review is read-only, so I have not run it and have not approved — approving or merging while the draft flag is set is forbidden. Handing promotion to the host path.

## False-claim boundaries

- The cadence results above were produced with an **injected clock and a modelled 1.5s subprocess cost per cycle**, not by running a live long-lived `notify supervise` process on real hardware. The threshold arithmetic, the `CompletedAt`-at-completion change and the 600/601 boundary are verified from source and by construction; **no production wall-clock measurement was taken**.
- Real GitHub-backed `ci-all-green-not-transitioned` and queue-debt classes were **not** re-constructed at this head; that code path is unchanged by the last two commits and was constructed at an earlier head.
- No claim is made that any real supervision loop has run in production under this build.

tomohisa:
## Approval — PR #1388 @ `5cc8a2e27f41438237cf2caaa2ac5d79da3e7bba`

The draft flag has been cleared by the host path since my verdict comment above, and the head is unchanged. Re-verified at the moment of approval:

- head `5cc8a2e27f41438237cf2caaa2ac5d79da3e7bba` (unchanged throughout the review)
- `isDraft: false`, base `main` (matches the packet's `direct-main` base policy), `MERGEABLE` / `CLEAN`
- CI `Build and test (source contract)` **success**, `head_sha` confirmed equal to the reviewed head
- `intent-cli review closeout-plan` → `ready: true`, `gaps: []`
- `Closes #1387` matches the linked source issue (G311)

All acceptance criteria, the evidence for each, the out-of-scope check, the intent references, and the false-claim boundaries are recorded in the verdict comment immediately above. All three previously reported blockers are resolved and were reproduced independently rather than accepted from the fix commits.

Applying `intent-cli automation pr-transition --transition approved`. **Not merging** — merge and closeout belong to the host path.

PR #1392, opened 2026-08-07
G644: surface supervision setup in guides

### LINKED ISSUE #1391: G644: A surface no guide names does not exist
# G644: A surface no guide names does not exist

## Goal

Make supervision **reachable from the guidance a role actually reads**: the workflow guides name the supervision setup as a step — who runs it, one standing loop per team, and how its absence is noticed — and `guide next` recommends setting it up when the team has **no recorded supervision cycle**.

## Why This Slice Exists Now

**The operator's test, verbatim**: telling the design thread *"use intent-cli's latest supervise so the loop does not stop"* should be enough, and if it is not, that is a design fault. It is not enough today.

**Measured on installed 0.14.0, 2026-08-07** — every guide a role is told to read:

| guide | mentions of `supervise` |
|---|---|
| `guide workflow task review-next-slice-loop` | **0** |
| `guide workflow task implementation-loop` | **0** |
| `guide workflow task init-host` | **0** |
| `guide next` | **0** |

The only description lives inside the two-thousand-line orchestration reference. **A design thread reaches it only if it already knows the surface exists** — which, on this host, was true solely because design authored the slice.

**The part most easily lost is the deployment, not the command.** The reference correctly says intent-cli neither starts nor manages the background process; it does not say **who does**, **how many run per team**, or **what happens if none is running**. Those were settled here by measurement rather than by reading: the command loops when `--once` is omitted, a supervisor placed inside a seat cannot report its own death, and two supervisors on one team wake the same stall twice.

**This host proves the cost.** `design-watchdog.sh` existed for weeks and **was not running**, and nothing said so. A supervision surface that the supervising role cannot find is the same failure with a newer name: **the mechanism for noticing that work stopped is itself unnoticed**.

**And it undoes the point of the previous slice.** G641 made time-to-recovery bounded and measured; a bound nobody sets up is a number that never gets taken.

**Route from the guides people read, do not enlarge the reference.** The fix is discoverability, so it belongs where a role is already looking rather than deeper inside a page it reaches only when it knows what to search for.

**Make the missing setup detectable, not just documented.** `guide next` recommending supervision when no cycle is recorded is a check a reviewer can run, whereas prose can only be read. Documentation alone would repeat this defect one layer up.

**Record what was measured, not what seems sensible**: the loop semantics, one per team, outside the seats. Anything not measured is stated as guidance rather than as fact.

## Current Observed State

- No workflow guide and no `guide next` output mentions supervision.
- The reference documents the command and its semantics, but not who deploys it, how many, or the consequence of none.
- A team with no supervision running is indistinguishable, from every guide's point of view, from a supervised one.

- G641's supervision loop, its bound, its records and its self-liveness reporting are correct and unchanged.
- The reference page's content is correct; this slice routes to it and adds the deployment facts it lacks.
- intent-cli does not start or manage the background process, and this slice does not change that.

## Target Repo / Path / Part

Target part: supervision discoverability in the workflow guides and the next-action advisor.

## In Scope

- **The workflow guides name supervision as a setup step** — at minimum host initialisation and the design-side loop guide — stating that one standing loop runs per team, that it runs outside the seats because a supervisor inside a seat cannot report its own death, and that intent-cli neither starts nor manages it.
- **`guide next` recommends setting up supervision when the team has no recorded supervision cycle**, and stops recommending it once a cycle exists. This is the check that makes the gap detectable rather than merely documented.
- **The deployment facts recorded with their basis**: the command loops when `--once` is omitted; two supervisors on one team wake the same stall twice; a supervisor inside a seat dies with the seat. State which of these were measured on this host and which are guidance.
- **A pointer, not a copy**: the guides link the reference rather than restating its semantics, so the two cannot drift.

## Out Of Scope

- Starting, supervising or managing the background process from intent-cli.
- Changing G641's loop, bound, records or liveness behaviour.
- Prescribing a specific interval or bound value for any team.
- Adding supervision guidance to guides whose role has no part in it.

The child PR makes supervision reachable from the guidance roles read: the host initialisation guide and the design-side loop guide name supervision setup as a step, stating that exactly one standing loop runs per team, that it runs outside the seats because a supervisor placed inside a seat cannot report its own death, and that intent-cli neither starts nor manages that process, while linking the orchestration reference for the semantics rather than restating them so the two cannot drift. The next-action advisor recommends setting up supervision when the team has no recorded supervision cycle and stops recommending it once a cycle exists, so that a missing supervisor is detectable by running a command rather than only by reading prose. The deployment facts are recorded with their basis, namely that the command loops when once is omitted, that two supervisors on one team wake the same stall twice, and that a supervisor inside a seat dies with the seat, with each stated as measured or as guidance rather than presented uniformly as fact. Intent-cli still does not start, supervise or manage the background process, G641's loop, bound, recovery records and liveness behaviour are unchanged, no interval or bound value is prescribed for any team, no supervision guidance is added to guides whose role has no part in it, the surface is documented as preview under the compatibility promise's preview lane, tests cover the advisor recommending with no recorded cycle and staying silent once one exists and the guides naming the setup step, EN and JA stay in parity under the G613 terminology policy, and the full suite is green.

## Acceptance Criteria

- Host-init and design-side loop guides name supervision setup, including one-per-team, outside-the-seats, and the not-managed-by-intent-cli fact.
- `guide next` recommends supervision setup when no cycle is recorded, and does not once a cycle exists.
- Deployment facts recorded with their basis, distinguishing measured from guidance.
- Guides link the reference rather than restating semantics.
- No process management added; G641 behaviour unchanged; no interval or bound prescribed; no irrelevant guide touched.
- Preview labelling; ja/en parity under G613; full suite green.

## Verification

- Run each named guide and assert supervision setup appears; run an unrelated guide and assert it does not.
- With no recorded supervision cycle, run `guide next` and assert the recommendation appears; record a cycle and assert it disappears.
- grep the diff for any process start or management: assert none.
- Diff the guides against the reference: assert semantics are linked, not duplicated.

## Related Links

- Measured on installed 0.14.0, 2026-08-07: zero mentions of `supervise` in `review-next-slice-loop`, `implementation-loop`, `init-host` and `guide next`.
- This host's `design-watchdog.sh` existing unrun with nothing reporting it; the four silent stalls of 2026-08-06 and 07.
- G641 — the supervision this makes reachable; G628 — the preview lane.

## Knowledge Maintenance

- Intent placement: node 02 primary (guidance reachability), node 08 supporting (supervision deployment).
- ADR candidate: none.
- Diagram candidate: none.
- Docs update: ja/en guide surfaces (in scope).
- Closeout writeback expected: yes — node 02 records that a surface no guide names is unreachable by the process meant to adopt it, and that discoverability gaps must be made detectable by a command rather than closed with prose alone.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1391\n\nSurface the measured supervision deployment setup in host-init and the design-side review/next-slice guide, and make guide next detect a missing recorded cycle without starting or managing a process. Update EN/JA command reference pointers and add focused regression coverage. Focused and full CLI tests pass; git diff --check is clean.


tomohisa:
G644 review blocker — measured-vs-guidance basis (review-context lens 3 / AC deployment facts with their basis). `SupervisionGuideText.DeploymentBasis` currently places all three claims under “measured on this host”. The canonical review context states that only “omitting `--once` leaves the command looping” was measured here; “one per team” and “outside the seats / dies with the seat” are reasoned guidance. Please reword the rendered guide output so those bases are distinguishable in substance, not only by having separate `Basis:` and `Guidance:` labels, and update the G644 regression assertions to pin the corrected distinction. Everything else in the review passed; no G641 behavior change is requested.

PR #1414, opened 2026-08-10
feat(notify): persist report outbox

### LINKED ISSUE #1413: G653: A report must survive its own delivery
# G653: A report must survive its own delivery

## Goal

Persist every canonical report to a **local outbox before transport**, so a delivery failure never destroys the result; surface undelivered outbox entries through supervision; and support **collection**: when a recipient's terminal holds a completed `ORCH_RESULT` whose canonical report never arrived, the recorded remedy is to re-send that same report once for the same task id — collection, not re-delegation.

## Why This Slice Exists Now

**Reported by the remote-herdr team with transcript, 2026-08-08.** Their review seat finished its work and executed the canonical blocked report — and the report **failed at the herdr socket for a permission reason, and the result was simply gone**. `report_arrived=false` persisted while the design thread read `live` as working for hours. The work had been done, the reporting command had been run, and **nothing durable existed anywhere** to say so.

This host shipped the same principle for inbound tasks in G619 (the task file is written before the pointer) and for reports' delivery refusal in G640 (a message is not bookkeeping). The gap is the symmetric one: **an outbound result has no durable form independent of its transport.** The completion-marker contract already makes recipients print `ORCH_RESULT <nonce> <status> <artifact>` in the terminal — a de facto outbox nobody reads.

**Write first, send second** — the G619 shape. The outbox entry carries the task id, nonce, status, artifact and summary, and delivery failure marks it undelivered rather than erasing it.

**Collection is not re-delegation.** Re-sending a persisted report for the same task id resolves the same pending record; it never re-dispatches the task. One collection per task id, and the pending-record disagreement rules (G625/G629) are unchanged.

**Supervision surfaces, humans or roles collect.** The undelivered-outbox finding names the entry and the collection command; nothing is auto-sent.

## Current Observed State

- `notify report` builds the payload and attempts transport; on failure nothing durable remains on the sender's side.
- The `ORCH_RESULT` terminal marker exists by contract but no surface consumes it.
- Supervision has no notion of an undelivered result.

- G640's deliver-with-advisory semantics and G629's pending-record resolution are correct and unchanged.
- G619's write-before-send precedent is the shape to copy.
- The completion-marker contract (prefix, nonce, status, artifact) is correct and unchanged.

## Target Repo / Path / Part

Target part: the report outbox, the undelivered finding, and the collection path.

## In Scope

- Every `notify report` writes an outbox entry (task id, nonce, status, artifact, summary, timestamps) before transport; write failure fails closed before any transport attempt; delivery failure marks the entry undelivered and the command says where it is.
- Supervision surfaces undelivered outbox entries once per entry per cycle, naming the collection command.
- A collection path re-sends a persisted entry for its task id, resolving the matching pending record exactly as a first-time report would, refusing a second collection for an already-delivered task id, and never re-dispatching anything.
- ja/en guidance: the terminal `ORCH_RESULT` is the recipient-side record and the outbox is the sender-side one; when a result is visible in a terminal but no report arrived, collect — do not make the recipient redo the work.

## Out Of Scope

- Parsing terminals from intent-cli; collection input is the persisted outbox (or explicit operator-supplied values), not screen-scraping.
- Auto-sending anything from supervision.
- Changing delivery semantics, pending-record rules, or the completion-marker contract.

The child PR makes every notify report persist an outbox entry carrying the task id, result nonce, status, artifact, summary and timestamps before any transport is attempted, failing closed before transport when the entry cannot be written, marking the entry undelivered when transport fails so the result survives, and telling the caller where the entry lives. Supervision surfaces undelivered outbox entries once per entry per cycle naming the collection command and auto-sends nothing. A collection path re-sends a persisted entry for its task id and resolves the matching pending delegation exactly as a first-time report would, refuses collecting a task id whose report already arrived, and never re-dispatches any task. Terminals are not parsed by intent-cli, delivery semantics, the pending-record rules and the completion-marker contract are unchanged, EN and JA guidance explains the sender-side outbox and recipient-side marker and states that a visible result with no arrived report is collected rather than redone, the surfaces are documented as preview under the compatibility promise's preview lane, tests cover write-before-send, fail-closed on unwritable outbox, undelivered marking, the finding appearing once, successful collection resolving the record, and the already-delivered refusal, and EN and JA stay in parity under the G613 terminology policy with the full suite green.

## Acceptance Criteria

- Outbox written before transport; unwritable outbox fails closed pre-transport; failed delivery leaves the entry marked undelivered with its location reported.
- Supervision surfaces undelivered entries once per entry per cycle with the collection command; nothing auto-sent.
- Collection resolves the pending record like a first-time report; second collection for a delivered id refused; nothing re-dispatched.
- No terminal parsing; delivery/pending/marker contracts unchanged; preview labelling; parity under G613; full suite green.

## Verification

- Break transport, send a report: assert the entry survives marked undelivered and the location is reported.
- Run supervision: assert one finding naming the collection command.
- Collect: assert the pending record resolves; collect again: assert refusal.
- Make the outbox unwritable: assert fail-closed with no transport attempt.

## Related Links

- remote-herdr field report, 2026-08-08: the blocked report lost to a herdr-socket permission failure; recorded in node 08.
- G619 — write-before-send; G640 — a report is a message; G629 — pending-record resolution; the completion-marker contract in the task envelope.

## Knowledge Maintenance

- Intent placement: node 08 primary (report durability), node 03 supporting (durable state before transport).
- ADR candidate: none.
- Diagram candidate: none.
- Docs update: ja/en reporting guidance (in scope).
- Closeout writeback expected: yes — node 08 records that a result must exist durably before its delivery is attempted, and that collection is not re-delegation.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
## Summary

- persist report outbox entries before transport and retain failed deliveries
- add one-shot manual collection that resolves the original pending report without redispatch
- surface undelivered outbox entries through supervision without auto-send or terminal parsing

Closes #1413

## Validation

- `dotnet test tests/IntentSystem.Cli.Tests/IntentSystem.Cli.Tests.csproj --no-restore --filter "FullyQualifiedName~NotifyPendingDelegationG629Tests|FullyQualifiedName~NotifySupervisionG641Tests|FullyQualifiedName~JapaneseTerminologyGuardG613Tests"` (25 passed)
- `dotnet test IntentSystem.sln --no-restore`
- `git diff --check`


tomohisa:
CI repair for c5e63997: the G588 documented-schema test seeded and reported every supported status under one task ID. G653 correctly preserves the first delivered outbox entry and refuses the later synthetic collisions. The fixture now assigns each schema case a unique task ID; production durable-outbox and double-collection refusal are unchanged.\n\nValidated in Release: exact failing test (1/1), focused G588/G629/G641 (30/30), and full solution suite.  also passed.

tomohisa:
## G653 semantic review — REQUEST_UPDATE (head `bd0789ba94b1c25c26a976ce913c966af37c68b0`)

Reviewed in a managed scratch worktree detached at this exact head; head SHA confirmed `bd0789ba…` before and after, worktree removed afterwards. Full suite green (`IntentSystem.Cli.Tests` 4600 passed / 1 pre-existing skip, all other assemblies green), `git diff --check` clean, exact-head CI `Build and test (source contract)` success. The core mechanism works — every positive requirement in the review context reproduced correctly under my own probes. One finding blocks approval.

---

### Finding — implementation-finding — an outbox entry permanently blocks any later report for that task id

**Failed criterion.** Out Of Scope: *"Changing **delivery semantics**, pending-record rules, or the completion-marker contract."* The Standalone Child Issue Contract asks for exactly one refusal — *"**refuses collecting** a task id whose report already arrived"* — and this change also refuses **reporting**.

**Root cause.** `NotifyReportOutboxStore.WriteNew` keys on the task id alone, across every delivery state:

```csharp
if (current.ContainsKey(entry.TaskId))
{
    return new NotifyReportOutboxWriteResult(false, path, $"An outbox entry already exists for task '{entry.TaskId}'.");
}
```

and the report path treats that as fail-closed-before-transport. So once *any* outbox entry exists for a task id — delivered or undelivered — every subsequent `notify report` for that id is refused with no transport attempted.

**Verified by execution** (probes added to the existing fixtures at this head, run, then reverted):

**1. A re-delegated task id becomes permanently unreportable.** `notify delegate` accepts a task id that has already been used — it returns exit 0 and opens a fresh pending record — after which neither path can carry the result:

```
delegate#1            exit=0
report#1              exit=0   (delivered; pending resolved; outbox marked delivered)
delegate#2 (same id)  exit=0   ← the host accepts re-delegating the same task id
report#2              exit=1  cause=report-outbox-write-failed  transportAttempted=0
  "Could not persist report task 'PROBE-reuse' before transport:
   An outbox entry already exists for task 'PROBE-reuse'. No transport was attempted."
collect fallback      exit=1  cause=already-collected
```

The recipient has done the work and has **no way to report it at all**. `notify report` refuses before transport, and `notify collect` refuses because the *old* entry is delivered. Nothing durable records the second result — not even an undelivered outbox entry, because the write is what failed. That is the precise failure this slice exists to eliminate ("the work had been done, the reporting command had been run, and nothing durable existed anywhere to say so"), reintroduced by the slice itself, and strictly worse than the state it replaces because the pre-transport refusal leaves no artifact behind.

**2. Re-running the canonical report command after a transport failure is refused without naming the remedy.** This is the most likely real sequence — a role whose report failed re-runs the command its task envelope tells it to run:

```
report#1 (transport down)  exit=1  cause=transport-failure          (entry retained, undelivered — correct)
report#2 (transport up)    exit=1  cause=report-outbox-write-failed  transportAttempted=0
  "Could not persist report task 'PROBE-retry' before transport:
   An outbox entry already exists for task 'PROBE-retry'. No transport was attempted."
```

`notify collect` *is* the right action here and it works — but the refusal message says only that an entry exists. It does not name the collection command, so the caller is stopped without being routed to the remedy that the supervision finding does name.

**3. A second unmatched report is silenced.** For a task id with no pending record, the first report is delivered with the advisory; the second is refused pre-transport:

```
report#1  exit=0  delivered=True
report#2  exit=1  cause=report-outbox-write-failed  transportAttempted=0
```

This contradicts the paragraph that sits **directly below the new text in the same file**, `docs/en/12-agent-message-orchestration.md`: *"A report is a message rather than a bookkeeping entry: fail-closed protection belongs on pending-state mutation, not on carrying the message. **Refusing an unrecognised identifier would silence unsolicited reports and answers to escalations**."* An answer to an escalation, or a `question` report followed later by a `completed` one under the same id, is now silenced.

**The regression signal was removed rather than resolved.** The only test that exercised repeated reports under a single task id — `NotifyRecordedRolesG588Tests.EveryEventKindNotifyCanEmit_BelongsToTheDocumentedSchema_G588`, which called `SeedPending()` for `G588-demo` before each supported status — began failing on `c5e63997` and was changed in `bd0789ba` to mint a unique task id per status. The PR comment states "production durable-outbox and double-collection refusal are unchanged", which is accurate as far as it goes, but the test was not failing because of collection: it was failing because **reporting** under a re-seeded task id changed behavior. Rewriting the fixture removed the coverage of that path; no test in the changeset now exercises a reused task id.

**Required unblock action.** Make the outbox key match the lifecycle it is tracking rather than the task id alone, so a new report attempt for a re-opened pending record is a new entry (e.g. key by task id + dispatch/pending generation or result nonce, or supersede a delivered entry when the pending record has been re-opened). At minimum, a delivered entry must not block a report for a subsequently re-delegated task id, and an unmatched report must not be silenced on its second occurrence. If the collision refusal is genuinely intended for the undelivered case, its message should name `notify collect` as the remedy. Please also restore coverage for a reused task id — the G588 fixture was the only test holding that line.

---

### Verified and correct — reproduced independently, not taken from the changeset's tests

- **Write-before-send is ordered, not concurrent (G619 shape).** With the outbox write denied, the command failed closed and **zero** process calls were made — not merely no prompt, but no `agent list` either: `transportCallsAfterUnwritableOutbox=0 (any=)`.
- **A failed transport leaves the result durable and says where it is.** `cause=transport-failure`; outbox `state=undelivered error=transport-failure fileExists=True`; the summary carries the retained path, and `outbox_entry_path` is emitted.
- **Collection re-sends only the report and re-dispatches nothing.** The transport calls during collect were exactly `agent list` → `agent prompt wH:p1 {"notification":"report","task_id":"PROBE-order","status":"completed",…}` → `agent wait`. No delegate payload, no task envelope, no `send-text`.
- **Collection resolves the pending record exactly as a first-time report (G629/G640 unchanged).** After collect: `status.report_arrived=True`, `report_status=completed`, outbox `delivered`.
- **Double collection is refused.** `cause=already-collected`, with **0** transport calls.
- **Supervision surfaces once and auto-sends nothing.** The observation is `undelivered-report-outbox` with `WakeSuppressed = true`, keyed per entry, its summary naming `intent-cli notify collect … --write` and "do not re-delegate the task"; the second cycle with unchanged state emits nothing. Grepping the added supervisor lines finds no transport call.
- **No terminal parsing.** Grepping every added source line finds no pane/screen/scrollback read and no `ORCH_RESULT` consumption; the terminal marker is referenced only as prose describing the recipient-side record.
- **Docs.** EN and JA both gain the report-outbox paragraph in parity, and **both** compatibility ledgers gain the `notify collect` command row and the `report outbox record` durable-artifact row, each labelled `preview-through-1.x`.
- `Closes #1413` is present and points at the linked source issue (G311); base is `main`, matching the packet's `direct-main` policy. No label knowledge leaked; no `intent-target` / `intent-pr-created` added by the diff; no AI provider launched.

**Intent consistency.** `intents/intent-cli/intent-tree/means/08-agent-message-orchestration.md` — the node's rule that a claim must correspond to observable record and evidence, and G619's write-before-send precedent, are honoured by the outbox mechanism itself. The finding above is where the change departs from node 08's companion rule that a report is a message which must not be silenced by bookkeeping.

### What was NOT verified — must not be claimed as collected

- **No real herdr socket, process, or pane was driven.** All behavioral evidence above comes from the repository's fake-runner fixtures at this head; the field-reported herdr-socket permission failure was not reproduced against a real transport.
- Passing tests is necessary but not sufficient and is not the basis of this verdict: the suite is fully green at this head, and the one test that covered the affected path was rewritten in this PR.
- Probes were added to the existing test files, executed, and reverted; `git status` was confirmed clean afterwards and nothing was pushed to this branch.

Reviewed read-only apart from this durable comment: no push, no merge, no label mutation, and the draft flag was left set.

tomohisa:
G653 lifecycle repair at 49b14786: outbox records now carry a durable entry ID and only collide within the same non-empty dispatch result nonce. A re-delegated task ID creates a new generation; unmatched reports and a settled question followed by completed remain messages and are delivered. An undelivered current-generation repeat remains fail-closed before transport and names: intent-cli notify collect --domain intent-cli --team intent-cli-dev --task-id <task-id> --write. Collection remains report-only and double collection refuses.\n\nRegression coverage restores G588 reused-task-ID schema cases and adds delegate1/report1, delegate2 same ID/report2, repeated unmatched, question-to-completed, failed transport/repeated-report collect remedy, collect-once, and second-collect refusal.\n\nValidated: focused Release G588/G629/G641/G613 37 passed; full Release solution suite passed; git diff --check passed.

tomohisa:
## G653 re-review — REQUEST_UPDATE (head `49b14786901f688c2cef8291be2e3c0d3b5e0513`)

The prior blocker's **main path is fixed** and I reproduced that independently. Two residual paths still strand a completed result, one of them introduced by this repair. Reviewed in a managed scratch worktree detached at this exact head; head SHA confirmed `49b14786…` before and after, worktree removed afterwards. Release suite green (`IntentSystem.Cli.Tests` 4601 passed / 1 pre-existing skip, all other assemblies green), `git diff --check` clean, exact-head CI success. Probes added to existing fixtures, executed, reverted; `git status` clean afterwards; nothing pushed.

### Prior blocker — main path resolved

Entries now carry a durable `entry_id`, `ReadCurrent` keys by it rather than by task id, and the collision test requires a matching **non-empty** `ResultNonce`. Verified by execution:

```
[A] re-delegate with a NEW result nonce
  delegate#1 (nonce-A1) exit=0 | report#1 exit=0 delivered=True
  delegate#2 (nonce-A2) exit=0 | report#2 exit=0 delivered=True      ← both carry (was: refused)

[C] repeated unmatched reports
  unmatched#1 delivered=True | #2 delivered=True | #3 delivered=True  ← G640 restored

[D] question then completed on one dispatch
  report question delivered=True | report completed delivered=True

[E] failed transport, then a repeat of the same generation
  report#1 exit=1 transport-failure   (entry retained, path reported)
  report#2 exit=1 report-outbox-write-failed  transportAttempted=0
    "An undelivered report outbox entry already exists … Recover it with
     'intent-cli notify collect --domain intent-cli --team intent-cli-dev --task-id E-task --write';
      do not re-delegate the task."                                   ← exact remedy named

[F] collect once, then again
  collect exit=0; calls = agent list | agent prompt wH:p1 | agent wait wH:p1   (report-only)
  status.report_arrived=True report_status=completed
  collect again exit=1 already-collected, 0 transport calls

[write-before-send] outbox write denied → exit=1, processCalls=0 []    ← zero calls, G619 shape
```

---

### Finding 1 — implementation-finding — supervision names a collect command that cannot collect the entry it surfaced

**Failed criterion.** AC: *"Supervision surfaces undelivered entries once per entry per cycle **with the collection command**"*, and *"Collection resolves the pending record like a first-time report"*.

`ExecuteCollect` now scopes the lookup to the **current** pending generation:

```csharp
var outbox = pending.Resolved && pending.Record is not null
    ? NotifyReportOutboxStore.Find(routingRoot, options.Domain!, options.Team!, options.TaskId!, pending.Record.ResultNonce)
    : NotifyReportOutboxStore.Find(routingRoot, options.Domain!, options.Team!, options.TaskId!);
```

but `ReadUndelivered` — which feeds the supervision finding — returns undelivered entries from **every** generation, and the finding's summary names a `notify collect --task-id <id>` command carrying no generation. When a generation is superseded, the two disagree:

```
[G] undelivered gen1 superseded by gen2
  delegate g1 (nonce-G1)         exit=0
  report g1 (transport down)     exit=1 transport-failure      → entry retained, undelivered
  delegate g2 (nonce-G2)         exit=0                        → new generation opens
  collect                        exit=1 cause=outbox-entry-unavailable
                                 "No persisted outbox entry exists for task 'G-task'."
  undelivered entries still surfaced by supervision: G-task/nonce-G1
```

Supervision keeps reporting the gen-1 entry as undelivered and keeps printing a recovery command that answers *"no persisted outbox entry exists"* — while the entry plainly exists and is the one being surfaced. The completed gen-1 result is now unrecoverable through any command.

This sequence is not exotic: it is precisely the two situations this packet is about, composed. A report fails at the transport (the field-reported herdr-socket case), the orchestrator does not notice and re-delegates — which the herdr operating guide names as the remedy: *"Use `intent-cli notify delegate` for the same logical role with the task id plus concrete repair delta"* (`HerdrOnlyOperatingGuide.cs:224`) — and the original result is stranded behind a finding whose remedy is dead.

**Required unblock action.** Make the surfaced entry and the collectable entry the same set. Either have the finding name a generation-qualified collect (and let `collect` accept it), or have `collect` fall back to the surfaced undelivered entry when the current generation has none, or stop surfacing entries that collection can no longer reach — but a finding must not name a command that cannot act on it.

### Finding 2 — implementation-finding — re-delegating with the same result nonce still strands the result, with no remedy named

Generation is defined as the result nonce, and `notify delegate` silently accepts a repeat of the same (task id, nonce) pair:

```
[B] re-delegate reusing the SAME result nonce
  delegate#1 (nonce-Y) exit=0 | report#1 exit=0 delivered=True
  delegate#2 same id + same nonce-Y  exit=0                     ← accepted silently
  report#2   exit=1 report-outbox-write-failed  transportAttempted=0
    "A report outbox entry already exists for task 'B2-task' and its current dispatch generation.
     No transport was attempted."                               ← no remedy named
  collect    exit=1 already-collected
```

Both paths refuse and the message takes the *delivered* branch, which — unlike the undelivered branch added by this repair — names no `notify collect` command. This is the prior blocker narrowed to the nonce-reuse case rather than eliminated. The operator has no signal at delegate time that reusing the nonce will make the next report unsendable, and no signal at report time about what to do instead.

**Required unblock action.** Either refuse or warn at `notify delegate` when a task id is re-dispatched with an already-used result nonce, or treat a re-opened pending record as a new generation regardless of nonce reuse. At minimum, the delivered-branch message should say what to do next, as the undelivered branch now does.

### Observation — not a blocker, but it changes what the restored G588 test proves

The repair comment states *"Regression coverage restores G588 reused-task-ID schema cases."* `NotifyRecordedRolesG588Tests` was reverted to re-seed `G588-demo` before each status, and it passes — but `Workspace.SeedPending` writes a pending record with **no `ResultNonce`**, so `!string.IsNullOrWhiteSpace(persisted.ResultNonce)` is false and the collision check is skipped entirely. That test therefore exercises the nonce-less path, not the generation logic. It is worth knowing that a dispatch without a result nonce silently loses the "undelivered current-generation repeat fails closed" protection altogether — the safe direction, but undocumented and untested.

### Re-confirmed and correct

- **Write-before-send ordered, not concurrent (G619).** Outbox write denied → `report-outbox-write-failed` with **0** process calls, not even `agent list`.
- **Undelivered durability and location reporting** unchanged and working; `outbox_entry_path` emitted and the retained path named in the summary.
- **Collection is report-only and never re-dispatches.** Calls during collect: `agent list` → `agent prompt wH:p1` → `agent wait wH:p1`; no delegate payload, no task envelope, no `send-text`.
- **G629/G640 pending resolution unchanged.** Collect resolves the record exactly as a first-time report (`report_arrived=True`, `report_status=completed`); unmatched reports carry with the advisory; double collection refuses with 0 transport calls.
- **Supervisor detect-only, once, no auto-send.** `NotifyMeasuredSupervisor` is untouched by this repair; the observation remains `WakeSuppressed = true`, once per entry, with no transport call in the supervise path.
- **No terminal parsing.** No added source line reads pane/screen/scrollback or consumes `ORCH_RESULT`.
- **Docs parity.** EN and JA both gained the generation paragraph with matching substance; the `preview-through-1.x` labelling and both compatibility-ledger rows from the previous head are retained.
- `Closes #1413`; base `main`, matching the packet's `direct-main` policy.

### What was NOT verified — must not be claimed as collected

- **No real herdr socket, process, or pane was driven.** All behavioral evidence comes from the repository's fake-runner fixtures at this head; the field-reported herdr-socket permission failure was not reproduced against a real transport.
- Passing tests is necessary but not sufficient and is not the basis of this verdict: the Release suite is fully green at this head and neither finding above is covered by it.

Reviewed read-only apart from this durable comment: no push, no merge, no label mutation, draft left set.

tomohisa:
G653 residual-generation repair at 97097616: notify delegate now refuses before pending write or transport when the task has any undelivered outbox report, naming the exact collect command. It also rejects an already-used task-id/result-nonce pair before work and requires a fresh --result-nonce or new task ID. Collection carries persisted unmatched or legacy reports too, but resolves pending state only when the collected entry matches the open dispatch generation; supervised findings are keyed per entry and their commands are collectable.\n\nNonce-bearing regressions cover: undelivered gen1 then attempted gen2 refusal and collect; same nonce after delivered refusal; different nonce after delivered success; same-generation retry then collect; exact supervisor remedy then successful collect; unmatched failed report collection without redispatch. Existing G588 reuse, G629/G640, no-auto-send, and no-terminal-parsing coverage remains.\n\nValidated: focused Release G588/G629/G641/G613 39 passed; full Release solution suite passed; git diff --check passed.

tomohisa:
CI repair for 97097616 at ede4a7cd: GuardDelegateOutboxLifecycle treated any matching pending nonce as reuse, including the still-open/no-outbox JSON-to-Markdown advisory resend in NotifyCommandG578Tests. The guard now permits only that idempotent open-generation resend. It still refuses the same nonce after settlement or any existing outbox, and it still blocks all re-delegation when the task has an undelivered report.\n\nRegression coverage distinguishes open same-nonce resend (allowed) from delivered same-nonce re-delegation (refused), while retaining different-nonce success and undelivered-task refusal.\n\nValidated in Release: exact CI failure 1/1; focused G618/G653/G588/G629/G641/G613 66/66; full solution suite; git diff --check.

tomohisa:
## G653 final re-review — APPROVE (head `ede4a7cd2e4473640ebb2e1c87316192ae90370c`)

Both residual blockers from [#issuecomment-5237456310](https://github.com/J-Tech-Japan/intent-system/pull/1414#issuecomment-5237456310) are **resolved**, and the open-resend CI regression is preserved rather than papered over. Reviewed in a managed scratch worktree detached at this exact head; head SHA confirmed `ede4a7cd…` before and after, worktree removed afterwards. Release suite green (`IntentSystem.Cli.Tests` **4604 passed / 1 pre-existing skip**, all other assemblies green), `git diff --check` clean, exact-head CI `Build and test (source contract)` success. Every result below was produced by my own probes against this head — added to the existing fixtures, executed, reverted, `git status` confirmed clean, nothing pushed.

### The fix chosen is the right shape

Rather than teaching `collect` to reach backwards, `notify delegate --write` now refuses to open a new generation while a result is stranded. That makes the previous defect structurally unreachable instead of patched: a superseded undelivered generation can no longer come into existence, so the supervision finding and the collectable set are the same set by construction. `ExecuteCollect` was widened to reach undelivered and unmatched/legacy entries, and it resolves pending state only when the collected entry matches the open dispatch generation.

### Prior Finding 1 — supervision named an uncollectable command — resolved end to end

I took the command the supervisor actually printed and **executed it verbatim**:

```
delegate S1 (nonce s-n1)        exit=0
report (transport down)          exit=1 transport-failure   → entry retained, undelivered
notify supervise --once          exit=0
  undelivered findings = 1   wake_attempted = False
  remedy = intent-cli notify collect --domain intent-cli --team intent-cli-dev \
           --task-id S1 --write --routing-root <root>
EXECUTING THE PRINTED REMEDY  →  exit=0  delivered=True
  calls = agent list | agent prompt wH:p1 | agent wait wH:p1      (report-only)
```

The remedy is now generated by the same `BuildCollectCommand` used by the delegate guard and the report refusal, it carries `--routing-root`, the finding key is per `entry_id`, and the command works. The `outbox-entry-unavailable` dead end is gone.

### Prior Finding 2 — same-nonce re-delegation stranded the result — resolved, and it now fails before work

```
[2] delegate T2 (m1) exit=0 | report (settles) exit=0
    delegate T2 same nonce m1  → exit=1 result-nonce-already-used
      "Supply a fresh --result-nonce or a new --task-id before delegating; no work was started."
      processCallsOnRefusedDelegate = 0
    delegate T2 fresh nonce m2 → exit=0 ; report gen2 → exit=0 delivered=True
```

The trap is now caught at the point where it is cheap and actionable — before any pending write or transport — and the message names both remedies. Previously the same sequence was accepted silently at delegate and only failed later, unrecoverably, at report.

### The six lifecycle requirements, independently reproduced

```
[1] undelivered gen1 blocks gen2 delegation BEFORE work, and the exact collect succeeds
    delegate g2 → exit=1 undelivered-report-outbox
      "…Recover it with 'intent-cli notify collect … --write --routing-root <root>';
        do not re-delegate the task. No work was started."
    processCallsOnBlockedDelegate = 0    pending nonce still gen1 (n1)   ← no state mutated
    collect → exit=0 delivered=True, calls = agent list | agent prompt | agent wait

[2] same task+nonce after settlement refuses before work; fresh nonce succeeds   (above)

[3] same OPEN task+nonce resend with no outbox remains allowed
    delegate T3 (k1) exit=0 ; delegate T3 (k1) again exit=0        ← G578 idempotent resend intact

[4] unmatched failed report collects without redispatch
    unmatched report (transport down) → exit=1 transport-failure, entry retained
    collect → exit=0 delivered=True, calls = agent list | agent prompt | agent wait
    (this was impossible at the previous head — collect only saw generation-scoped entries)

[5] same-generation retry then collect, then double collect
    report#2 → exit=1 report-outbox-write-failed, transportOnRetry = 0,
      naming the exact collect command
    collect → exit=0 ; status.report_arrived = True
    collect again → exit=1 already-collected, secondCollectCalls = 0

[6] every supervised undelivered entry is collectable                            (above)
```

The CI repair is a genuine narrowing, not a loosened guard: `[3]` confirms the open-generation resend that `NotifyCommandG578Tests` covers still passes, while `[2]` confirms the same nonce after settlement is still refused and `[1]` confirms an undelivered entry still blocks all re-delegation.

### Earlier G653 invariants re-confirmed at this head

- **Write-before-send is ordered, not concurrent (G619).** Outbox write denied → `report-outbox-write-failed` with `processCalls=0 []` — not even `agent list`.
- **Undelivered durability and location reporting** intact; the retained outbox path is named in the failure summary and `outbox_entry_path` is emitted.
- **Collection is report-only and never re-dispatches.** Across all three collect paths above the calls were exactly `agent list` → `agent prompt wH:p1` → `agent wait wH:p1`; no delegate payload, no task envelope, no `send-text`.
- **G629/G640 pending semantics unchanged.** Collect resolves the record as a first-time report when the generation matches (`report_arrived=True`), and deliberately does not resolve it when it does not; unmatched reports still carry as messages; double collection refuses with zero transport.
- **Supervision is detect-only, once per entry, auto-sends nothing.** `wake_attempted=False`, `WakeSuppressed = true`, one finding per `entry_id`, no transport call in the supervise path.
- **No terminal parsing.** No added line in either repair commit reads pane/screen/scrollback or consumes `ORCH_RESULT`.
- **Docs parity.** EN and JA both gained the delegate-guard paragraph with matching substance (undelivered refusal naming the same collection command, settled task-id/nonce refusal requiring a fresh nonce or task id, idempotent open-generation resend). The `preview-through-1.x` labelling and both compatibility-ledger rows are retained.

### Packet / AC / intent mapping

Execution unit **G653**, packet `.intent-cli/issues/G653/packet.yaml` (`issue_kind: bugfix`, target `src/IntentSystem.Cli/**, docs/en/**, docs/ja/**, tests/**`, `dependencies: []`).

- **AC 1** — outbox written before transport; unwritable outbox fails closed pre-transport; failed delivery leaves the entry `undelivered` with its location reported: verified.
- **AC 2** — supervision surfaces undelivered entries once per entry per cycle with the collection command, nothing auto-sent: verified, and the command was executed successfully.
- **AC 3** — collection resolves the pending record like a first-time report, second collection refused, nothing re-dispatched: verified across generation-matched, unmatched, and retry paths.
- **AC 4** — no terminal parsing; delivery/pending/marker contracts unchanged; preview labelling; G613 parity; full suite green: verified. The one deliberate change to delivery-adjacent behaviour — refusing a delegate that would strand a result — is fail-closed **before** any work and is documented in both languages.

**Out Of Scope not crossed.** No terminal is parsed; supervision auto-sends nothing; the completion-marker contract is untouched. The delegate guard does not change how a report is delivered — it prevents opening a generation that would orphan one, which is the packet's stated purpose ("a report must survive its own delivery").

**Intent evidence** — `intents/intent-cli/intent-tree/means/08-agent-message-orchestration.md`, G619's write-before-send precedent and the node's rule that a report is a message which must not be silenced by bookkeeping. Both hold at this head: nothing durable is destroyed by a transport failure, unmatched reports still carry, and the one new refusal happens before work rather than after it.

`Closes #1413` is present and points at the linked source issue (G311). Base is `main`, matching the packet's `direct-main` policy. No label knowledge leaked; no `intent-target` / `intent-pr-created` added by the diff; no AI provider launched.

### What was NOT verified — must not be claimed as collected

- **No real herdr socket, process, or pane was driven.** All behavioural evidence comes from the repository's fake-runner fixtures at this head; the field-reported herdr-socket permission failure was not reproduced against a real transport.
- **Tests passing is necessary but not sufficient** and is not what this approval rests on — every earlier head in this review was also fully green while carrying a stranding path. The approval rests on the packet/AC/intent mapping plus the lifecycle probes above, each executed against this head.
- Non-blocking observation, unchanged from the previous round and not an approval condition: a dispatch recorded with no `result_nonce` skips the generation collision check entirely, so the "undelivered current-generation repeat fails closed" protection does not apply to it. That is the safe direction, and the delegate guard's undelivered check still covers such tasks, but it remains undocumented.

Reviewed read-only apart from this durable comment and the canonical draft promotion: no push, no merge, no label mutation. Merge and closeout — including the node 08 write-back that `packet.yaml` requires — remain with the host orchestration path.

PR #1460, opened 2026-08-12
G675: make supervision scheduler artifacts self-sufficient

### LINKED ISSUE #1459: G675: a supervisor that cannot run its tools is degraded, not seeing losses — scheduler env completeness and truthful transport failure
# G675: a supervisor that cannot run its tools is degraded, not seeing losses — scheduler env completeness and truthful transport failure

## Goal

Make the `supervise install` artifact **carry its own runtime environment** — every binary the loop needs resolved absolutely or given an explicit recorded PATH at emission — and make supervise classify a **transport binary that fails to start** as its own degraded finding, never as `recipient-lost`, surfaced once per cycle rather than once per open delegation.

## Why This Slice Exists Now

**Measured on this host, 2026-08-12, in two acts.**

**Act one: the installed supervisor was never running.** The emitted launchd plist invoked `/usr/bin/env intent-cli`. launchd's minimal PATH does not include the dotnet tool directory, so every KeepAlive restart exited 127 — while `launchctl list` kept showing the label as loaded and the design seat believed supervision was installed. The gap was found only during an audit prompted by **another team's incident on the same machine the same day** (four accumulated supervisors, seats duplicated — their workaround for install friction was hand-run loops, which then outlived their sessions). G641's rule — a supervisor that is not running is indistinguishable from a healthy system — held again, this time with the scheduler itself as the silent non-runner.

**Act two: the revived supervisor reported ten false losses in one cycle.** After the executable path was fixed by hand, the loop started — into the same minimal environment. intent-cli invokes `herdr` via PATH; herdr was not on it. The transport failed with `No such file or directory` (working directory `/`), and supervise classified that as **`recipient-lost` — for every one of the ten open delegations at once**, escalating each. The recipient at the recorded pane was alive and mid-task. This is G648's lesson one layer down: **the inability to run the check is not the failure of the checked.** And the fan-out (one cause, ten escalations) repeats the exact per-delegation amplification G648 de-duplicated for registration loss.

**Why this belongs to the tool.** `supervise install` exists (G658) so teams stop hand-running loops. An artifact that works in the shell that emitted it and dies under the scheduler's environment converts the install path back into the hand-run path — with a worse failure mode, because the scheduler retries silently forever.

**Resolve at emission, verify after registration.** The artifact must not inherit the caller's shell: intent-cli absolutely, transports absolutely or via an explicit recorded PATH written into the artifact. And the emitted guidance gains a verification step — live PID plus a first cycle record — because "loaded" is not "running", as this host just measured.

**Cannot-check is its own truth.** A transport-start failure becomes `supervision-degraded: transport unavailable` naming the binary and error — a state the reader acts on (fix the environment) instead of ten wrong verdicts the reader must first disbelieve. G648's corroboration rules for real absence are untouched.

**Once per cycle.** One broken transport is one finding, not one per open delegation — the G648 de-duplication principle applied to this cause.

## Current Observed State

- The plist emission uses `/usr/bin/env intent-cli`; no PATH is written; transports are resolved from the runtime environment.
- Supervise classifies transport-start failure inside its liveness path; the resulting verdict is `recipient-lost` per open delegation.
- The emitted guidance ends at the registration command; nothing verifies the loop actually ran.

- G658's install surface (emission-only, no registration or process management by intent-cli) is correct and stays; this slice changes what the artifact contains and what the guidance says, not who registers it.
- G648's corroborated-absence rules and its per-cycle de-duplication for registration loss are correct; this extends the same shape to transport failure.
- G641's self-liveness reporting is correct and unchanged.

## Target Repo / Path / Part

Target part: the supervise-install artifact's runtime environment completeness, and supervise's classification of transport-start failure.

## In Scope

- **Emission-time resolution**: the artifact invokes intent-cli by absolute path; transport binaries the loop will need are resolved absolutely at emission or provided through an explicit PATH recorded in the artifact; unresolvable binaries are named in the emission output rather than left to fail at runtime.
- **A `supervision-degraded` transport finding**: binary name, error text, once per cycle; never `recipient-lost`; existing verdicts for genuinely absent recipients unchanged.
- **Post-registration verification** in the emitted guidance: scheduler shows a live PID and a first cycle record exists; the silent-exit failure shape named.
- **The incident recorded in guidance** as measured (both acts, this host, 2026-08-12).
- **Preview labelling** per G628; EN/JA parity.

## Out Of Scope

- Registering, starting, or managing the scheduler artifact from intent-cli (G658's boundary stands).
- Duplicate-supervisor detection (G676).
- Changing recovery sequences, verdict names, or wake transports.
- Windows/systemd emission beyond applying the same resolution rule to what G658 already emits.

The child PR makes the supervise-install artifact self-sufficient in environment: the intent-cli executable is resolved to an absolute path at emission, every transport binary the supervision loop needs is resolved absolutely at emission or provided through an explicit PATH recorded in the artifact, and a binary that cannot be resolved at emission is named in the emission output rather than left to fail silently under the scheduler. Supervise classifies a transport binary that fails to start as a supervision-degraded transport-unavailable finding naming the binary and the error, surfaced once per cycle regardless of how many delegations are open, and never emits recipient-lost on that basis, while the corroborated-absence rules for genuinely missing recipients are unchanged. The emitted guidance adds a post-registration verification step requiring a live scheduler PID and a recorded first cycle and names the silent-exit failure shape, the guidance records the measured incident in which this host believed supervision installed while the scheduler looped on exit 127 and the revived loop then escalated ten false losses in one cycle from one missing transport binary, intent-cli still registers and manages nothing, duplicate-supervisor detection is left to its own slice, recovery sequences, verdict names and wake transports are unchanged, the surfaces are documented as preview under the compatibility promise's preview lane, tests cover the emission-time resolution, the degraded finding replacing recipient-lost on transport-start failure, its once-per-cycle surfacing, and the unchanged genuine-absence path, EN and JA stay in parity under the G613 terminology policy, and the full suite is green.

## Acceptance Criteria

- Artifact self-sufficient: absolute intent-cli, transports resolved or explicit recorded PATH; unresolvable binaries named at emission.
- Transport-start failure → `supervision-degraded` (binary + error), once per cycle, never `recipient-lost`; genuine-absence path unchanged.
- Guidance: post-registration verification (live PID + first cycle record); silent-exit shape named; incident recorded.
- G658 boundary intact (nothing registered or managed); preview labelling; EN/JA parity under G613; full suite green.

## Verification

- Emit on a host with intent-cli and herdr present: assert absolute paths / recorded PATH in the artifact.
- Run supervise with the transport binary unavailable and open delegations present: assert one degraded finding, zero recipient-lost.
- Run with a genuinely absent recipient and transports healthy: assert the existing lost path unchanged.
- Render the emitted guidance: assert the verification step and the named silent-exit shape.

## Related Links

- Measured incident, this host, 2026-08-12: exit-127 KeepAlive loop behind a loaded label; then ten `recipient-lost` escalations in one cycle from `herdr` failing to start (`No such file or directory`, working directory `/`) while the recipient was alive and mid-task.
- Same-day, same-machine context: a second team dismantled four accumulated hand-run supervisors — the failure mode install exists to end (G676 takes the duplication itself).
- G658 — the install surface and its boundary; G648 — cannot-see is not absent, and the per-cycle de-duplication; G641 — the silent non-runner rule.

## Knowledge Maintenance

- ADR candidate: none.
- Diagram candidate: none.
- Docs update: ja/en install guarantees, degraded finding, verification step (in scope).
- Closeout writeback expected: yes — node 08 records both acts as one rule: an environment assumption is a measured fact, not a default.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1459

Summary:
- Resolve intent-cli and selected runtime transport executables at scheduler emission, record PATH and unresolved binaries, and preserve emit-only scheduler boundaries.
- Classify transport process-start failure before recipient liveness as cycle-level supervision-degraded / transport-unavailable without recipient-lost findings or per-delegation wakes.
- Preserve G648 absence corroboration and event-mode semantics; add EN/JA guidance, live-PID/first-cycle verification, measured attribution, and preview ledger coverage.

Verification:
- Targeted regression: 166 passed.
- Full suite: 4794 passed, 1 skipped (4795 total).
- CLI build: 0 warnings, 0 errors.
- git diff --check: clean.


tomohisa:
G675 review blocker (implementation-finding; exact head d057cf0345eb315cdf247bd4f1ed3d6484f8826d)

Failed acceptance criterion / packet goal: supervise-install emission must remain emit-only and usable while recording unresolved binaries; full suite must be green. Exact-head CI failed 7 pre-existing G658/G659 tests because NotifySuperviseInstallCommand now returns exit 1 before preview/artifact emission whenever intent-cli cannot resolve as the current process or a PATH-visible binary.

Current evidence: GitHub Actions run 31610111255 failed NotifySuperviseInstallG658Tests (default dry-run, 3 explicit platforms, Markdown) and NotifyEventSupervisionG659Tests (both event modes), each expected exit 0 but received 1. The CI host runs under dotnet/vstest, so ResolveIntentCliExecutable returns null. The same shape plausibly exists for supported dnx/dotnet-tool-exec invocation. This is not a flaky check.

Required unblock action: do not hard-fail when intent-cli cannot be resolved absolutely. Name intent-cli in unresolved_binaries and preserve the pre-existing bare intent-cli command fallback for preview/emitted invocation, as already done conceptually for unresolved herdr/bash. Keep absolute resolution when available. Add/adjust tests proving dry-run and every platform/event-mode artifact remain emit-only and exit 0 without a PATH-visible intent-cli; run the full suite and get exact-head CI green.

False-claim boundary: G675 is NOT merge-ready and full-suite green was NOT established at this head. Do not claim scheduler self-sufficiency is fully verified until the fallback/unresolved recording contract and CI are green.

The transport-start preflight ordering, once-per-cycle degraded finding, genuine G648 lost path, G658 management boundary, and G676 non-scope were reviewed as sound; keep this repair narrow.

PR #1485, opened 2026-08-13
G686: Add operator-recorded envelope profiles

### LINKED ISSUE #1484: G686: the host's chosen envelope is a recorded fact — drift compares against the choice, not the catalogue
# G686: the host's chosen envelope is a recorded fact — drift compares against the choice, not the catalogue

## Goal

Give the recipe-drift comparator a second layer of truth: **named, typed envelope profiles recorded in topology** (per kind, referenced or overridden per role), operator-recorded and never inferred, so a host whose operator sanctioned a launch shape stops producing a correct-but-useless drift alarm every cycle. No profile recorded → the kind recipe registry stays the baseline, byte-identically.

## Why This Slice Exists Now

**G684's detector fired on all three of this host's seats on its first cycle** — and every finding was true. The observed argv genuinely differs from the recorded kind recipe, because the operator decided that same day to align seats with the host-standard launch. The detection is correct; what is missing is **anywhere to record the sanctioned choice**. This is G684's own law (a detector firing on sanctioned choices buries real regressions) recurring one axis over: the wish/envelope split solved it for *model flags*; the *kind-recipe-versus-host-choice* axis has no recording surface at all.

**The survey (`G686-design-consult`, 2026-08-13, source head `5e6bf6b6`) fixed three design decisions with measurement:**

1. **`launch_args` must not be the profile.** Topology's optional `launch_args` already overlays the recipe for the comparator *and* feeds G630's recovery `StartRecipe` — reusing it as a drift baseline would silently change what recovery relaunches. The profile is comparator-only, a separate typed record.
2. **One host-wide profile per kind is not enough.** Measured on this host's three seats of one kind-family: p1 runs network with no approval flag and two roots, p2 network with approval-never and a different root, p3 under `--remote` with no local sandbox argv at all. A single kind profile broad enough for all three would **union away real drift** — so profiles are named variants with per-role references (or typed per-role overrides).
3. **G620's discipline stands.** `update-field` deliberately accepts only `delivery_method`; profiles must not reopen arbitrary JSON editing. Recording is a **dedicated typed CAS command** with explicit confirm and current digest, and the legacy parser must be taught (and test-locked) not to misread a top-level profile object as a role.

**Profiles are recorded, never learned.** Auto-deriving a profile from observed argv would convert every unnoticed drift into a new baseline — the detector would ratify what it exists to catch. The operator's choice enters through a command, with evidence and a timestamp, or the registry stays the baseline.

**Two layers, explicit precedence**: role's referenced profile if recorded → kind recipe registry otherwise. Unrecorded hosts and seats are byte-identical to today, locked by test.

**Fail closed on broken references**: a dangling profile ref or kind mismatch is a distinct `profile-invalid` finding — never a silent fallback that would make a typo look like compliance.

**Recovery is untouched**: `launch_args` semantics and `StartRecipe` behaviour do not change; the profile never feeds a relaunch.

## Current Observed State

- The comparator's only baseline is `AgentLaunchRecipeRegistry.Find(kind)` (plus `launch_args` overlay when present; absent on all three seats here).
- Three true-but-sanctioned drift findings fire per cycle on this host; node 08 records the operator decision they collide with.
- No topology surface can carry an envelope choice; `update-field` correctly refuses everything but `delivery_method`.

- G684's comparison algorithm, its envelope/wish split, and its once-per-seat-per-cycle surfacing are correct; this slice changes only which expected envelope is compared against.
- G647's recipe registry stays the catalogue of measured launch shapes and the fallback baseline.
- G630's recovery semantics and G620's registry limits are correct and untouched.

## Target Repo / Path / Part

Target part: the typed envelope profile in topology, its dedicated recording command, and the drift comparator's baseline selection.

## In Scope

- **Typed envelope profiles in topology**: named per-kind variants carrying sandbox mode, approval mode, roots policy, network access, transport mode, evidence and recorded_at (optionally a digest); per-role reference or typed per-role override selects one.
- **A dedicated recording command** (typed CAS: explicit confirm + current digest; naming within the topology command family), leaving `update-kind`/`update-field` registries unchanged.
- **Baseline selection in the comparator**: referenced profile → else registry; `launch_args` behaviour unchanged.
- **`profile-invalid`** as a distinct fail-closed finding for dangling refs and kind mismatches.
- **Parser safety**: the legacy topology parser never misreads a profile object as a role (envelope-property registration, test-locked).
- **No learning**: nothing derives a profile from observed argv.
- **Preview labelling** per G628; EN/JA docs for schema, command, precedence, finding.

## Out Of Scope

- Changing G684's comparison fields, surfacing cadence, or wish exclusions.
- Feeding profiles into recovery, launch, or any non-comparator consumer.
- Auto-migration of existing hosts; recording is always an explicit operator-side act.
- Widening `update-field`, or any generic topology JSON editing.

The child PR adds named typed envelope profiles to the topology record carrying sandbox mode, approval mode, roots policy, network access, transport mode, evidence and recorded time per kind variant, selected per role by reference or typed override, recorded only through a dedicated typed compare-and-set command with explicit confirmation and the current digest and never derived from observed argv. The drift comparator uses the role's referenced profile as its expected envelope when one is recorded and the kind recipe registry otherwise, hosts and seats without profiles behave byte-identically to the current build locked by test, launch_args semantics and recovery StartRecipe behaviour are unchanged, a dangling profile reference or kind mismatch fails closed as a distinct profile-invalid finding rather than falling back silently, the legacy topology parser is taught and test-locked not to misread a top-level profile object as a role, and the update-kind and update-field registries stay exactly as limited as today. The comparison fields, surfacing cadence and wish exclusions of the drift detector are unchanged, no non-comparator consumer reads profiles, no auto-migration occurs, the surfaces are documented as preview under the compatibility promise's preview lane, tests cover the profile-selected baseline, the registry fallback, the byte-identical unprofiled path, the profile-invalid finding, the parser safety and the CAS recording command, EN and JA stay in parity under the G613 terminology policy, and the full suite is green.

## Acceptance Criteria

- Profiles: named typed per-kind variants with the listed fields; per-role ref/override; recorded only via the dedicated CAS command; never learned.
- Comparator precedence: profile → registry; unprofiled paths byte-identical; `launch_args`/recovery untouched.
- `profile-invalid` fail-closed on dangling ref or kind mismatch; parser never misreads profiles as roles (test-locked); G620 limits intact.
- G684 fields/cadence/exclusions unchanged; preview labelling; EN/JA parity under G613; full suite green.

## Verification

- Record a profile matching this host's sanctioned launches; run a cycle: assert the three findings clear while a deliberately broken seat still drifts.
- Remove the profile ref: assert registry baseline returns, byte-identical to pre-change.
- Point a role at a missing profile: assert `profile-invalid`, not silence.
- Attempt to record via update-field or raw JSON paths: assert refusal; CAS command with stale digest: assert refusal.

## Related Links

- Survey `G686-design-consult` (2026-08-13, source head `5e6bf6b6`): the comparator path, the launch_args/recovery coupling, the three-seat diversity measurement, and the G620-preserving command shape — all adopted here.
- G684 — the detector this gives a second truth layer; G647 — the catalogue that stays the fallback; G630 — the recovery semantics preserved; G620 — the registry discipline preserved; node 08 (2026-08-12/13) — the operator decision and the first-cycle firing.

## Knowledge Maintenance

- ADR candidate: none.
- Diagram candidate: none.
- Docs update: ja/en schema, command, precedence, finding (in scope).
- Closeout writeback expected: yes — node 08 records that a detector shipped against a catalogue found its first real gap in the catalogue itself.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1484

## G686 contract mapping

- Adds a typed, named per-kind envelope-profile model with sandbox mode, approval mode, roots policy and writable roots, network access, transport mode, evidence, recorded-at, optional permission options/network URLs, and a canonical digest.
- Adds the dedicated session topology record-profile command with explicit confirmation, typed input, current-digest CAS, atomic persistence, and optional per-role reference or typed override. Existing update-kind/update-field surfaces are not expanded.
- Resolves comparator precedence as role override, then role profile reference, then registry recipe. Profile-invalid, dangling, kind-mismatched, or malformed profiles fail closed with no registry fallback; unprofiled roles retain the registry path.
- Integrates profile baselines into production supervision without changing G684 envelope fields, model/reasoning exclusions, cadence, wake suppression, or detect-only behavior.
- Updates the orchestrator guidance, command reference, EN/JA documentation, and preview-through-1.x compatibility ledger.
- Adds production-path regressions for typed profile CAS/refusal, profile precedence and registry fallback, profile-invalid fail-closed behavior, comparator equivalence, parser safety, and documentation parity.

## Repair mapping: discussion 3772808714

- ReadString now checks JsonValueKind.Object before TryGetProperty, so scalar and array role/profile values cannot throw during classification.
- Non-object or malformed override/profile entries become distinct profile-invalid findings before any process-runner call, with no registry fallback.
- Direct topology -> supervision regressions cover scalar and array envelope_profile_override, malformed object overrides, and scalar entries in accepted profile array/map shapes.

## Repair mapping: discussion 3772930135

- ReadProfileReference now distinguishes absent/null references from present non-string references for both envelope_profile and envelope_profile_ref.
- Present array, object, number, or boolean references name the property and JSON kind in a profile-invalid result; they never become absent and never fall back to the registry, valid alias, or override.
- Null remains the intentional unprofiled case and preserves the registry comparator path.
- Real topology validate and show plus supervision regressions cover both aliases, all four non-string shapes, null compatibility, invalid-reference precedence over a valid alias/override, and no agent list call.

## Verification

- Focused G686/G666/topology coverage: 71 passed (G686 23; G666/topology 48).
- Full IntentSystem.sln Release suite with isolated NuGet cache: CLI 4,921 passed, 1 skipped; all other solution assemblies passed; 0 failures.
- git diff --check: clean.
- Exact repaired head: af0da5a7b15687997cbb932bb5f66c17844fb38b.
- No AI provider was launched and no host metadata was read or mutated.


src/IntentSystem.Cli/Commands/NotifyRoleTopology.cs
```diff
@@ -225,6 +233,65 @@ private static NotifyTopologyResolution ResolveFromPath(
...
+                        $"Role '{property.Name}' has an invalid envelope profile reference: {profileReferenceError}");
+                }
+
+                AgentLaunchEnvelopeProfile? profileOverride = null;
+                if (property.Value.TryGetProperty("envelope_profile_override", out var overrideElement)
+                    || property.Value.TryGetProperty("profile_override", out overrideElement))
+                {
+                    if (!AgentLaunchEnvelopeProfileCodec.TryRead(
+                            overrideElement,
+                            ReadString(overrideElement, "name") ?? $"{property.Name}:override",
```
tomohisa, line 250: [P1] Fail closed for malformed profile overrides

The typed-profile contract requires an invalid profile to become a distinct profile-invalid finding. A scalar or array envelope_profile_override (for example, "envelope_profile_override": "bad") reaches this ReadString call before AgentLaunchEnvelopeProfileCodec.TryRead can reject the non-object. JsonElement.TryGetProperty throws InvalidOperationException for a non-object, and ResolveFromPath catches only I/O and JSON exceptions, so notify supervision aborts instead of emitting profile-invalid. Guard the element kind (or pass the fallback name to the codec without reading it first) and add regression coverage for malformed overrides. The same object assumption also exists for scalar entries in the accepted profile-array/map shapes.

tomohisa: Addressed in 56d5eb83. ReadString now checks JsonValueKind.Object before TryGetProperty, so scalar and array role/profile nodes cannot throw. The codec now receives every non-object or malformed override/profile entry and the production topology -> supervision path emits the distinct profile-invalid finding before any process-runner call, with no registry fallback. Added regressions for scalar/array overrides, malformed object overrides, and scalar entries in profile arrays/maps. Focused G686/G666/topology: 60 passed; full IntentSystem.sln Release: CLI 4,910 passed, 1 skipped, 0 failed.

src/IntentSystem.Cli/Commands/NotifyRoleTopology.cs
```diff
@@ -623,6 +699,125 @@ private static bool TrySelectTeam(
         return false;
     }
 
+    private static string? ReadProfileReference(JsonElement element, out string? error)
+    {
+        var direct = ReadString(element, "envelope_profile");
```
tomohisa: [P1] Fail closed for non-string profile references

The repair makes ReadString safe, but it now causes a present non-string binding to be treated as absent. With envelope_profile: ["not-a-reference"], session-layer topology validate succeeds and topology show resolves the role with no envelope profile; ReadRecipeDriftObservations then takes the registry recipe path. This violates the G686 acceptance criterion that a broken profile reference yields profile-invalid with no registry fallback. Distinguish an absent/null ref (the intentional unprofiled case) from a present Array/Object/Number/Bool for both envelope_profile aliases, return profile-invalid for the latter, and add the topology-to-supervision regression asserting no agent list call.

tomohisa: Addressed in af0da5a71e6ada07de031d5d669b9880b7626afe. ReadProfileReference now distinguishes absent/null from present non-string values for both envelope_profile and envelope_profile_ref. Array, object, number, and boolean references are reported as profile-invalid with the property name and JSON kind, before profile override/registry selection; null remains the intentional unprofiled registry case. Added real topology validate/show and supervision regressions for both aliases, all four shapes, null compatibility, invalid-reference precedence over a valid alias/override, and no agent list call. Focused G686/G666/topology: 71 passed; full solution: CLI 4,921 passed, 1 skipped, 0 failed; exact-head CI 31674291552 is green.

tomohisa (commented):
Final semantic rereview: approved at exact head af0da5a71e6ada07de031d5d669b9880b7626afe.

Both prior P1 blockers are resolved: #3772808714 now fail-closes malformed override/profile shapes, and #3772930135 distinguishes absent/null from present non-string direct and alias profile references. The latter is exercised through topology validate, topology show, and supervision; invalid values emit profile-invalid before any agent list call, while null retains the intentional registry path.

G686 packet acceptance criteria and boundaries were rechecked: named typed CAS-recorded profiles remain comparator-only; G630 recovery/launch_args, G620 update-field limits, G684 fields/cadence/wish exclusions, no-learning, preview labeling, EN/JA docs, and parser safety are preserved. Design alignment with node 08 is confirmed.

Verification: focused EnvelopeProfileG686Tests 23/23; full IntentSystem.sln 4,921 passed, 1 skipped; exact-head CI succeeded; git diff --check and merge-tree succeeded. Closes #1484 is present, and the canonical closeout plan reports ready=true.

Issue #1489, opened 2026-08-13
G683 is stricter than the work allows: the validated class vocabulary is three entries, shell command approval is not one, and the design answer path was removed in the same release


tomohisa:
## Summary

0.20.0 fixed #1469 — the prompt-class producer is live and the `<agent-kind>:<prompt-class>` pair is now validated against a real vocabulary. Thank you; that part works.

But the combination now in force is stricter than the work allows, and it stalls seats:

- the validated vocabulary is **three classes**, and the dialog class that actually wedges a working seat — a **shell command approval** — is not one of them;
- an unmatched class is **escalate-only** and no class may be fabricated (correct);
- and the **design answer path was removed in the same release line**.

So for the most common blocking dialog there is, by construction, **no in-contract answerer at all**. The seat waits for a human even when the command is trivially safe and the design thread is watching it happen.

## Environment

- `intent-cli 0.20.0-d860c33-G685` (from `0.19.0-G675`)
- Four-thread herdr-only team, domain/team `remote-herdr`, macOS
- Measured 2026-08-13

## Measurements

The vocabulary is now discoverable, which is a real improvement:

```
$ intent-cli notify supervise ... --pre-approve 'codex:bogus-class' --pre-escalate 'codex:other' --once --dry-run
invalid-notification: Unknown prompt policy pair 'codex:bogus-class'.
Known values: codex:github-comment-post, codex:launch-hook-trust, copilot:launch-limited-permissions
```

Three classes. None covers a shell command approval.

## The incident this produced, today

The implementation seat wedged on this dialog (verbatim from the pane):

```
Would you like to run the following command?
Environment: local

$ rm -f /tmp/rh041b.trx; CI=true dotnet test RemoteHerdr.sln --no-build \
    --filter 'Category!=PostgresIntegration' --logger 'trx;LogFileName=/tmp/rh041b.trx' ...
  <inline python3 heredoc parsing that trx with ElementTree, printing outcome counts>
  exit $rc

› 1. Yes, proceed (y)
  2. No, and tell Codex what to do differently (esc)
```

The only destructive element is `rm -f` on a temp file **the same command creates on its next line**. The rest is the project's own test suite and read-only XML parsing. No credential, no permission widening, no network, no repo write. It also matches this operator's 2026-08-04 pre-authorisation for an agent deleting its own same-wake scratch file.

[...]

PR #1492, opened 2026-08-13
G689: add scoped shell command approval vocabulary

### LINKED ISSUE #1490: G689: the vocabulary is the constraint — shell approval classes with scoped policies, and a registry you can read
# G689: the vocabulary is the constraint — shell approval classes with scoped policies, and a registry you can read

## Goal

Cover the dialog class that actually wedges seats — **shell command approval** — without ever making it wholesale-approvable: a **two-layer model** where the class recognises the dialog and extracts the command, and **scoped policy instances** (argv token prefixes, categories, path constraints, effect tags) decide what may be answered, with **shell-AST verification of every segment** of compound commands and fail-closed escalation for anything unparseable. Add the missing **inspection surface** (`prompt-class list/describe`) so the vocabulary stops being discoverable only through refusal messages.

## Why This Slice Exists Now

**Operator-filed feedback, issue #1489 (0.20.0, measured 2026-08-13), confirmed by consult against source `d860c33`.** G683's producer works — and its vocabulary is **three hard-coded classes** in `AgentLaunchRecipeRegistry`, none covering shell approval. Teams can only choose accept/escalate for known pairs; **no surface adds a class**, and no positive listing exists (the vocabulary is discoverable only by recording a bogus pair and reading the refusal). So for the most common blocking dialog the outcome is always escalate-only — which in practice means, in the reporter's words, **"a human presses a key"**. Their seats wedged this way repeatedly; today's verbatim case was `rm -f <its own temp file>; dotnet test …; python heredoc` — blocked until the operator answered, while both the design thread and orchestration correctly refused to fabricate a class.

**Why the naive fix is dangerous, and the consult's model is adopted whole.** A `codex:shell-command` class matched on dialog chrome with a fixed answer would make `--pre-approve codex:shell-command` mean *approve any command* — the exact "one approval becomes approval of everything shaped like it" failure. The layers must split:

- **Class layer**: recognises the dialog shape, extracts the command payload. Never approvable bare.
- **Policy layer**: team-recorded **scoped instances** — argv **token-sequence** prefixes (`dotnet test <recorded target>` must not textually match `dotnet test && rm -rf /`), category, cwd/root and path constraints, effect tags.
- **AST verification**: compound commands (`;`, `&&`, pipes, redirects, heredocs, subshells) decompose; **every segment** must be in scope; unknown syntax, command substitution or out-of-scope paths **fail closed**.

**Teams add scoped policies, not producers.** Dialog recognition follows provider UI and stays shipped/tested in the CLI; what a team legitimately varies is which scopes it trusts. This answers #1489's point 1 without opening class fabrication.

**The measured case remains not-wholesale-approvable, by design**: the compound `rm -f …; dotnet test …` needs both an `owned-scratch-delete` and a `project-test` scope to pass — and scratch deletion ships **orchestration-only** (destructive stays off any broader floor). The consult's minimal-safe alternative — a bounded No/Esc plus rerun-without-cleanup — remains the documented fallback.

**First scopes, shipped narrow**: `project-test` (recorded project root, first argv tokens match the recorded test invocation; not called "read-only" — tests run arbitrary code), `owned-scratch-delete` (exact path pre-recorded in the same wake's scratch ledger; never a bare `/tmp` prefix — symlinks and other work products), `exact-command-once` (normalized AST digest + current dialog hash, single use).

**Persistence is a human act**: pressing a provider's "don't ask again for commands that start with…" permanently mutates agent permission state — that choice is operator-only; the tool only answers bounded per-dialog Yes.

**Inspection**: `prompt-class list/describe` renders classes, scopes, and answerable roles — closing #1489's secondary discoverability gap and the guide's inert "verified read-only command approvals" wording, which is reconciled to the executable vocabulary.

## Current Observed State

- Three classes hard-coded; `TryValidateRule` refuses everything else; the refusal message is the only vocabulary listing.
- No shell producer; the guide's may-answer concept class is inert.
- The measured compound-command wedge required a human key press.

- G683's producer/validation/audit pipeline and its no-fabrication rule are correct and extended, not redesigned.
- G682's inapplicability honesty is correct; new classes clear it exactly as existing ones do.
- Answer-authority changes (who may answer) are G690's scope; everything here answers under today's orchestration-only rule.

## Target Repo / Path / Part

Target part: the shell-command prompt class with its scoped policy layer, the AST-verified compound-command matching, and the prompt-class inspection surface.

## In Scope

- The `codex:shell-command` class (dialog recognition + payload extraction); bare-class approval structurally impossible.
- Scoped policy schema (token-sequence prefix, category, cwd/root/path constraints, effect tags, persistence flag) and team recording of instances against shipped classes.
- Shell-AST decomposition with all-segments-in-scope acceptance; fail-closed on unknown syntax/substitution/out-of-scope redirects.
- Shipped scopes: `project-test`, `owned-scratch-delete` (orchestration-only), `exact-command-once`.
- Per-dialog bounded answers only; persistent allowances named operator-only.
- `prompt-class list/describe`; guide reconciliation; preview labelling per G628; EN/JA parity.

## Out Of Scope

- Team-defined producers/classes (recognition stays shipped).
- Changing who may answer (`answerable_by`, design authority) — G690.
- Any non-shell class additions beyond the guide reconciliation.
- Automatic pressing of persistent don't-ask-again options.

The child PR adds a codex shell-command prompt class responsible only for dialog-shape recognition and command payload extraction with bare-class approval structurally impossible, a scoped policy layer in which teams record instances carrying argv token-sequence prefixes that are never textual startsWith, category, cwd and root and path constraints and effect tags, and shell-AST decomposition of compound commands in which acceptance requires every segment to fall inside the matched scope while unknown syntax, command substitution and out-of-scope redirects or paths fail closed to escalation. The shipped first scopes are project-test bound to a recorded project root and recorded test invocation and documented as not read-only, owned-scratch-delete requiring the exact path pre-recorded in the same wake's scratch ledger and answerable by orchestration only, and exact-command-once keyed by normalized AST digest and current dialog hash for a single use. Answers are bounded per dialog and any persistent don't-ask-again option is documented operator-only and never pressed by the tool, a prompt-class list and describe surface renders the full vocabulary with scopes and answerable roles, the orchestrator guide's may-answer wording is reconciled to the executable vocabulary, G683's producer, validation, audit and no-fabrication rules and G682's inapplicability clearing are consumed unchanged, answer authority remains orchestration-only pending its own slice, the surfaces are documented as preview under the compatibility promise's preview lane, tests cover the two-layer split, the all-segments AST rule with fail-closed paths, each shipped scope including refusal of the measured compound command without both applicable scopes, the single-use digest scope, and the inspection surface, EN and JA stay in parity under the G613 terminology policy, and the full suite is green.

## Acceptance Criteria

- Shell class recognises/extracts only; bare-class approval impossible; scoped instances with the named fields are the only approval unit.
- AST all-segments rule with fail-closed unknowns; token-sequence prefixes never textual.
- Shipped scopes behave as specified (`owned-scratch-delete` orchestration-only; `exact-command-once` single-use).
- Per-dialog answers only; persistence operator-only; inspection surface + guide reconciliation shipped.
- G683/G682 semantics unchanged; authority unchanged; preview labelling; EN/JA parity under G613; full suite green.

## Verification

- Feed the measured compound dialog: assert refusal with only one scope recorded, acceptance with both, and audit records naming the matched scopes.
- Feed `dotnet test && rm -rf /` against a `dotnet test` prefix: assert escalation (token sequence, all-segments).
- Attempt `--pre-approve codex:shell-command` bare: assert structural refusal.
- Render `prompt-class list/describe`: assert vocabulary, scopes, roles; render the guide: assert reconciled wording.

## Related Links

- Operator-filed issue #1489: the three-class measurement, the verbatim wedge, the operator position; consult `design-consult-1489-authoring` (2026-08-13, source d860c33): the registry internals, the two-layer model, the shipped-scope designs adopted here.
- G683 — the pipeline extended; G682 — inapplicability clearing; G690 — the authority slice that follows; G666 — the layer model context.

## Knowledge Maintenance

- ADR candidate: none.
- Diagram candidate: none.
- Docs update: ja/en class model, scope schema, AST rules, inspection surface (in scope).
- Closeout writeback expected: yes — node 08 records that class and scope are different layers; collapsing them turns one approval into approval of everything shaped like it.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1490.\n\nImplements the measured shell-command prompt class, conservative shell-AST extraction, project-test / owned-scratch-delete / exact-command-once scoped policies, orchestration-only bounded answers, durable scope/digest/dialog audit fields, and prompt-class list/describe inspection. Includes English/Japanese guide and compatibility-ledger updates.


tomohisa (commented):
## G689 semantic review — request update

**Reviewed at exact head:** `2f3a350b434f83a5f25e0163ae9f298f4da3be06`
**Packet:** `.intent-cli/issues/G689` (packet.yaml / implementation.md / review-context.md) · **Linked issue:** #1490 (`Closes #1490.` present in the PR body and points at the linked source issue — G311 satisfied)
**Intent reference:** `intents/intent-cli/intent-tree/means/08-agent-message-orchestration.md` (node 08 owns the prompt/approval contract; `design_alignment_checked=true`)
**CI at exact head:** `Build and test (source contract)` — `conclusion: success`, completed `2026-08-13T20:47:10Z` (run `31742413364`, job `94588774542`). Tests passing is necessary, not sufficient; the findings below are packet/AC conformance gaps that a green suite does not cover.

### What the slice gets right

- **Bare-class approval is structurally impossible.** `AgentPromptClassRecipe.ScopedPolicyRequired = true` on the shell recipe makes `NotifyPreApprovalPolicyStore.TryValidateRule` refuse `codex:shell-command` as a bare pair, and `WithCurrentApplicability` marks any legacy bare rule `inapplicable-scoped-policy-required`. Review-context Q1 — satisfied.
- **Prefix matching is token-sequence, never textual.** `HasTokenPrefix` is `SequenceEqual` over argv tokens; `TryValidate` rejects shell operators and substitutions inside `argv_token_prefix`. `dotnet test && rm -rf /` fails because segment 2 has no candidate scope, not because of a string comparison. Review-context Q2 — satisfied for the parsed payload.
- **The parser fails closed** on `$`, backticks, `(`/`)`, single `&`, unclosed quotes, trailing escapes, missing heredoc delimiters, and substitution inside heredoc bodies; both segment scopes reject any segment carrying a redirect or heredoc.
- **The tool cannot press a persistent option.** `answer_keys` are pinned to the scope definition (`y`/`enter`); `Persistent` requires `operator_only_persistence=true` at validation and is then *refused* at evaluation (`persistent-allowance-operator-only`). Review-context Q3 — satisfied.
- Authority is unchanged (`answerable_by` forced to `orchestration`, G690 untouched), preview labelling is present in both compatibility ledgers, EN/JA doc parity holds, the guide's inert "verified read-only command approvals" wording is reconciled in `GuideOrchestratorThreadCommand` / `GuideDesignThreadCommand` and both `12-agent-message-orchestration.md` files, and `prompt-class list|describe` ships read-only.

---

### Finding 1 — implementation-finding: payload extraction silently truncates the command, so acceptance can authorize segments that were never evaluated

**Failed acceptance criterion:** *"Compound commands are decomposed by shell AST; acceptance requires **every segment** inside the matched scope; unknown syntax, command substitution, or out-of-scope redirects/paths fail closed to escalate."*

`ShellCommandPromptRecognizer.TryExtract` ends command extraction at the first choice-looking line and **drops every remaining line without inspecting it**:

```csharp
if (!insideFence && IsChoice(line))
{
    if (commandLines.Count > 0)
    {
        break;      // <-- everything after this point is discarded
    }
    continue;
}
```

`IsChoice` matches, among others, any line containing `[y/n]` and any line starting with `continue `, `allow `, `cancel `, `reject `. An unfenced multi-line command payload whose *second* line matches one of those causes the tail to disappear from the payload, and the AST all-segments rule is then applied only to the surviving prefix. Concretely, for a dialog rendering

```
dotnet test App.sln
read -p "continue? [y/n]" ans
rm -rf $HOME/work
```

the extracted payload is `dotnet test App.sln` alone. That prefix matches a recorded `project-test` scope, `Evaluate` returns `accept`, and the bounded `y` keypress answers the dialog for the **full** command — including the `rm -rf $HOME/work` segment which, had it been extracted, would have failed closed on `$` expansion. This is the "one approval becomes approval of everything shaped like it" failure the two-layer model exists to prevent, reintroduced at the extraction boundary rather than the policy boundary.

**Required unblock action:** make extraction fail closed rather than truncate — after the choice block begins, assert that every remaining non-empty, non-frame line is itself choice/frame text, and return `false` (or a `Parsed = false` payload) when residual non-choice content follows. Fenced payloads already behave correctly; the gap is the unfenced path.

### Finding 2 — implementation-finding: `owned-scratch-delete` is not bound to the wake, while shipped text states that it is

**Failed acceptance criterion / contract clause:** *"`owned-scratch-delete` requiring the exact path pre-recorded in **the same wake's scratch ledger** and answerable by orchestration only"* (child issue contract; AC2), and review-context question 4, which asks precisely whether this scope is really ledger-bound.

The exact-path half is implemented well: `TryValidate` requires non-empty `scratch_ledger_paths`, requires each of them to be an exact `path_constraints` entry, restricts `argv_token_prefix` to `rm` plus deletion flags, and `MatchesSegment` requires every `rm` target to be path-equal to a recorded ledger entry — a bare `/tmp` prefix cannot pass. The **wake binding does not exist**. `scratch_ledger_paths` is an operator-supplied field on `NotifyScopedPromptPolicy`, persisted into `.intent-cli/supervision/<domain>/<team>/pre-approval-policy.json`; when a later `notify supervise` run supplies no `--shell-policy`, `ResolvePreApprovalPolicy` reads `existing.ScopedPolicies` back off disk and uses them unchanged. A path recorded once therefore authorizes `rm -f <that exact path>` on every subsequent wake for the lifetime of that policy file. No scratch-ledger artifact exists in the repository for the code to consult (`git grep` over `src/` on `main` finds `scratch_ledger_paths` only in this PR).

That would be a defensible narrower design — except that four shipped surfaces assert the stronger property the code does not enforce:

- `ShellCommandPolicy.cs` scope description: *"An exact `rm -f` path recorded in **the same wake's** scratch ledger"*
- `docs/en/12-agent-message-orchestration.md`: *"requires exact paths from **the same wake's** scratch ledger"*
- `docs/ja/12-agent-message-orchestration.md`: *「**同じ wake** の scratch ledger にある exact path だけを要求し」*
- both compatibility ledgers: *"requires exact **same-wake** scratch-ledger paths"*

For a destructive, orchestration-only scope in a domain whose standing rule is that a recorded capability must not read as broader than what is enforced, the description and the enforcement have to agree.

**Required unblock action:** either (a) bind the scope to a wake — e.g. carry a wake/cycle identity on the scoped policy and refuse a match whose recorded ledger identity is not the current wake's — or (b) if the durable-record model is the intended design, correct all four texts to state what is actually enforced (operator-recorded exact paths in the durable per-team policy record, valid until that record is changed) and say plainly that the authorization is not wake-scoped.

---

### Informational (no change required)

- **Inspection surface under-reports the recognition vocabulary.** `ShellCommandPromptRecognizer.Headers` accepts six literal headers, while the recipe's `LiteralTextFragments` — the list `prompt-class describe` renders — carries one. `Classify` short-circuits to the shell branch before `MatchesActivePrompt` runs, so the five additional literals are live in recognition but invisible to the inspection surface that #1489's secondary discoverability gap asked for. Folding the full header list into the recipe would keep the registry the single source of the vocabulary.
- **Applicability granularity differs from the legacy path.** A legacy accept rule requires whole-policy `Applicable == true` (G682 aggregate clearing), whereas `Evaluate` filters scoped policies by their own per-instance `Applicable`. Per-instance is arguably the more honest granularity and each instance is independently validated, so this is noted rather than raised; worth a deliberate statement in node 08 either way. Separately, `WithCurrentApplicability` now returns `applicable=false` for a declaration-free policy with an "unavailable or invalid" reason and an empty `inapplicable_agent_kinds`, which reads as a defect report for a policy that simply declares nothing.
- **Closeout writeback** (`closeout_learning.write_back_required: true`, node 08) is not part of this diff; confirm it lands at closeout or is captured as a follow-up packet.

### Evidence boundary (what was NOT verified)

This review was performed from remote evidence only: `gh pr diff` / `gh pr view` for the complete changeset and `gh api .../commits/2f3a350b.../check-runs` for check results at the exact head SHA. **Local tests were not completed; exact-head GitHub Actions evidence was collected and verified.** Both findings above are derived from source reading, not from an executed reproduction — no runtime or manual execution evidence was collected and none is claimed. Out-of-scope boundaries from #1490 (team-defined producers/classes, `answerable_by` authority changes reserved for G690, non-shell class additions, automatic pressing of persistent options) were checked against the diff and are **not** crossed.

tomohisa:
Implementation repair required from the G689 semantic review at head 2f3a350b434f83a5f25e0163ae9f298f4da3be06. Finding 1: update src/IntentSystem.Cli/Commands/ShellCommandPromptRecognizer.cs so an unfenced payload never truncates at a choice-looking line; any residual non-choice command content after the choice block must fail closed, with regression tests proving no unevaluated segment can be authorized. Finding 2: update src/IntentSystem.Cli/Commands/ShellCommandPolicy.cs and its policy/store integration so owned-scratch-delete is bound to the current wake or cycle identity as required by G689 AC2; do not weaken the same-wake contract or docs. Add focused regressions, rerun the full suite, push the same PR, and complete the canonical pr-comment-fix lifecycle. Full evidence and rationale are in the preceding G689 semantic review.

tomohisa (commented):
## G689 rereview — APPROVED (both prior blockers closed)

**Reviewed at exact head:** `bc35c2f52ec2e137c7d047b8ebcfafe3e586404d` (repair commit `bc35c2f5` "fix: close G689 shell policy repair blockers", the single commit since the reviewed head `2f3a350b`)
**Execution unit / packet contract:** G689 — `.intent-cli/issues/G689/packet.yaml` (`target_part`: the shell-command prompt class with its scoped policy layer, AST-verified compound matching, and the inspection surface; `dependencies: []`), `implementation.md` Scope/Out-of-scope, `review-context.md` questions 1–4
**Linked issue:** #1490 — `Closes #1490.` is present in the PR body and points at the linked source issue (G311 satisfied)
**Intent reference:** `intents/intent-cli/intent-tree/means/08-agent-message-orchestration.md` — node 08 owns the prompt/approval contract; the repair strengthens rather than contradicts it (`design_alignment_checked=true`)
**CI at exact head:** `Build and test (source contract)` — `conclusion: success`, completed `2026-08-13T21:41:45Z` (run `31746636965`, job `94602660549`). Tests passing is necessary but not sufficient; the packet/intent evidence below is what carries the approval.

---

### Blocker 1 — payload extraction truncation: **CLOSED**

`ShellCommandPromptRecognizer.TryExtract` no longer breaks out of the loop at the first choice-looking line. It now sets `choiceBlockStarted` and, once the terminal choice chrome has begun, refuses the entire extraction if any residual non-choice line follows:

```csharp
if (!insideFence && choiceBlockStarted)
{
    if (IsChoice(line)) continue;
    return false;   // reject the whole extraction; no unevaluated segment can reach policy
}
```

`payload` stays `null`, so `AgentLaunchRecipeRegistry.Classify` produces no `ShellCommand`, `ShellCommandPolicyRegistry.Evaluate` is never reached, and the dialog is escalate-only. The exact scenario from my prior finding is now a test — `Classify_FailsClosedWhenChoiceBlockHasResidualCommandTail` feeds `dotnet test …` / `read -p "continue? [y/n]" answer` / `rm -rf <scratch>` and asserts `TryExtract` returns false, `payload` is null, and the resulting `PromptClass` is `unknown`. The all-segments AC invariant — *acceptance implies every segment of the command was evaluated* — now holds at the extraction boundary as well as the policy boundary.

### Blocker 2 — `owned-scratch-delete` wake binding: **CLOSED**

The scope is now bound to the active supervision cycle, and the identity is real rather than nominal:

- `NotifyScopedPromptPolicy.ScratchLedgerCycleId` (`scratch_ledger_cycle_id`) is a new persisted field.
- `RunOnceCore` generates `cycleId` **once at the start of the pass** and threads it through `ResolvePreApprovalPolicy`, `ReadObservedPrompts`, `Evaluate`, the observed prompt, and the prompt audit; the same value is what lands in the `cycles.jsonl` record (it was previously minted separately at cycle-record time), so the identity genuinely denotes this wake.
- `BindScopedPoliciesToCycle` stamps a freshly declared owned-scratch policy with the current cycle id, and **hard-refuses** a supplied policy carrying a different identity (`pre-approval-policy-cycle-binding-failed` / status `invalid-cycle-binding`).
- `ShellCommandPolicyRegistry.TryValidate` requires the identity by default and constrains its shape via `IsCycleIdentity`; `MatchesSegment` refuses any owned-scratch policy whose `ScratchLedgerCycleId` is not ordinal-equal to `currentCycleId`.
- A dedicated named refusal `owned-scratch-delete-stale-cycle` fires when an `rm` segment meets declared-but-not-current owned-scratch policies, so the failure is diagnosable rather than collapsing into generic `shell-segment-out-of-scope`.

**Missing** and **stale** identities both refuse, by two independent routes: a missing id fails `TryValidateScopedPolicy` (default `requireScratchLedgerCycleId: true`), which drives `WithCurrentApplicability` to `applicable=false` / `inapplicable-invalid-scoped-policy` and also blocks `NotifyPreApprovalPolicyStore.Record`; a stale id survives validation but fails the cycle comparison. The path that motivated the finding is the one that now bites: when a later supervisor run supplies no `--shell-policy`, `existing.ScopedPolicies` are read back from `pre-approval-policy.json` carrying the *earlier* cycle id and are refused. A scratch path recorded once can no longer authorize `rm` on a later wake.

The parse-time relaxation (`requireScratchLedgerCycleId: false` in `NotifyCommand.TryParse`/`Validate`) is correct and not a hole: the operator cannot know a future cycle id, the supervisor stamps it before `Record`, and `Record` itself validates with the requirement on.

This also makes the "same wake's scratch ledger" wording in `docs/en/12-agent-message-orchestration.md`, `docs/ja/12-agent-message-orchestration.md`, and both compatibility ledgers accurate, and the `ShellCommandPolicy.cs` scope description is updated to the sharper "current cycle's scratch ledger; stale wake identities are refused".

---

### Re-verified from the prior pass (unchanged by the repair)

Bare-class approval remains structurally impossible (`ScopedPolicyRequired` → `TryValidateRule` refusal, Q1); prefix matching remains argv `SequenceEqual` with no textual `startsWith` and every segment must find a scope (Q2); the tool still cannot press a persistent option (`answer_keys` pinned to the scope definition, `Persistent` refused at evaluation, Q3); the parser still fails closed on `$`, backticks, subshells, single `&`, unclosed quotes and heredoc substitution; `project-test` is still documented as test execution rather than read-only; authority remains `orchestration`-only with G690 untouched; G683 no-fabrication and G682 clearing semantics are consumed, not weakened; preview labelling and EN/JA parity hold; `prompt-class list|describe` stays read-only.

**Out-of-scope boundaries from #1490 were NOT crossed.** The repair adds no team-defined producer or class, no change to `answerable_by` or design authority (still G690's slice), no non-shell class addition, and no automatic pressing of a persistent option. No AI provider is launched by `intent-cli`; no `intent-target` / `intent-pr-created` label is added; no host metadata is touched by the diff. `NotifyPromptAudit.CycleId` extends the audit record, which strengthens the G683 audit trail rather than altering it.

### Informational — for closeout, not blocking

- **Fail-closed breadth.** Any dialog whose footer chrome after the choice list is not itself choice-like (e.g. a trailing "Esc to cancel" hint) now makes the whole class unrecognized rather than merely unanswerable. That is the safe direction, but it can quietly shrink the slice's real coverage; if measured dialogs carry such footers, recognizing terminal chrome explicitly would keep the class visible instead of silently inert.
- **Residual narrow tail.** A hidden tail composed *entirely* of choice-like tokens (a bare `run`, or a line starting with `cancel `/`reject `) is still skipped rather than refused. No destructive command shape fits that filter, so this is a hardening note, not a hole.
- **Blast radius of a mismatched supplied identity.** A `--shell-policy` JSON that carries a stale `scratch_ledger_cycle_id` aborts the entire supervision pass rather than marking that one instance inapplicable. Fail-closed and diagnosable, but a per-instance inapplicability would degrade more gracefully.
- **Inspection surface still under-reports the recognition vocabulary** (six recognizer headers vs. one `LiteralTextFragments` entry rendered by `prompt-class describe`) — carried over from the prior pass, unchanged by this repair.
- **Closeout writeback** (`closeout_learning.write_back_required: true`, node 08: class and scope are different layers) is not in this diff; confirm it lands at closeout or as a follow-up packet.

### Evidence boundary (what was NOT verified)

Remote evidence only: `gh api .../compare/2f3a350b...bc35c2f5` for the complete repair diff, `gh pr view` for PR state, and `gh api .../commits/bc35c2f5.../check-runs` for check results at the exact head. **Local tests were not completed; exact-head GitHub Actions evidence was collected and verified.** The focused 8/8 and full 4942-pass/1-skip figures reported by the implementation seat were **not** independently reproduced here and are not claimed as verified by this review; the green `Build and test (source contract)` run at `bc35c2f5` is the test evidence I hold. Both blocker closures above are established by source reading plus the named tests that ran in that green CI job — no runtime or manual execution evidence was collected, and none is claimed.

### Disposition

**Approved.** No implementation findings remain. The PR is still a draft: draft resolution and merge are the orchestration seat's calls at the merge boundary, not this review's — this seat delivers verdict and evidence only.

PR #1512, opened 2026-08-14
G699: supervision emission hygiene

### LINKED ISSUE #1511: G699: supervision emission hygiene
# G699: supervision emission hygiene

## Goal

Two emission-hygiene fixes measured on this host: repeated identical escalations back off into a **named parked state** (first-seen / last-seen / repeat count, visible on the status surface), and pane status classification **debounces transient flaps** with a declared consecutive-observation threshold before naming a seat blocked or stalled.

## Why This Slice Exists Now

Measured 2026-08-13/14 on this host:

- `supervision-recipient-G686-loop` recipient-lost escalations were emitted **55 times at a 10-minute cadence over ~9 hours** for the same unresolved recipient. The events stream became mostly repeats of one known fact — the G626/G670 lesson replayed: a detector alarming on a known state buries real regressions, because the reader stops reading.
- A claude pane transiently reported `blocked` while actually composing. An external watcher classified it as a permission stall and design intervened on a false positive; a 3-consecutive-poll debounce eliminated the false positives for the rest of the session. That debounce currently lives in a session-local script, not in the product.

Backoff must never become silence: the park is **named**, visible, counted, and reset by resolution or state change. The debounce threshold is recorded configuration, not a magic constant. A genuinely new escalation key is never delayed by another key's backoff.

## Current Observed State

- Escalations re-emit at full cadence for as long as the condition persists; identity keys (e.g. `recipient:G686-loop`) already exist.
- Status classification acts on single observations; flaps produce false stall findings.

- The escalation conditions themselves (recipient-lost etc.) detect correctly; this unit changes emission, not detection.
- The supervision status surface exists and can render parked entries.

## Target Repo / Path / Part

Target paths: `src/IntentSystem.Cli/**, tests/**, docs/en, docs/ja`

Target part: escalation backoff + status debounce.

## In Scope

- Same-key escalation backoff to a declared reduced cadence with a named parked state recording first-seen, last-seen, repeat count; reset on resolution or state change; parked entries visible on the supervision status surface.
- Consecutive-observation threshold (default 2–3 polls, recorded configuration) before blocked/stalled classification.
- Test: a new key fires immediately regardless of other keys' parks.
- EN/JA docs.

## Out Of Scope

- Changing detection conditions or adding new escalation kinds.
- Auto-resolving parked conditions.
- The continuation-chain surfaces (G695 — adjacent, unchanged).

The child PR makes repeated supervision escalations with the same identity key back off from their full cadence to a declared reduced cadence and transition into a named parked state that records first seen, last seen and repeat count, visible on the supervision status surface so that reduced emission reads as a named park rather than a lost signal, with the backoff reset by resolution or a state change of the underlying condition. It adds a declared consecutive-observation threshold, recorded as configuration with a small default, that pane and agent status classification must satisfy before naming a seat blocked or stalled, eliminating single-poll flap false positives. A genuinely new escalation key fires immediately and is never delayed by another key's backoff, proven by test. Detection conditions themselves are unchanged, EN and JA docs describe the park and debounce semantics under the G613 terminology policy, and the full suite is green.

## Acceptance Criteria

- Same-key backoff + named park (first/last/count, visible, reset on change).
- Declared debounce threshold before blocked/stalled classification.
- New-key immediacy test-proven.
- EN/JA parity; full suite green.

## Verification

- Test: 10 identical escalations → emissions follow declared backoff, park visible with count 10; condition change → reset.
- Test: 1-poll flap → no classification; threshold-consecutive → classification.

## Related Links

- The 55-emission measurement and the composing-flap false positive (this host, 2026-08-13/14).
- G626 (alarm honesty), G670 (do not prompt refused actions), G684 (drift emission), G686 (the loop that emitted).

## Knowledge Maintenance

- ADR candidate: none.
- Diagram candidate: none.
- Docs update: EN/JA (in scope).
- Closeout writeback expected: yes — "a detector that repeats a known fact 55 times has the same effect as one that never fired."

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1511

Implements bounded, readable repeated-observation supervision without changing detection predicates or the G695 continuation chain.

Configuration surface:
- `--interval <seconds>` is the full observation cadence.
- `--repeat-backoff-seconds <seconds>` (alias `--backoff-seconds`, default 1800) records repeat emission cadence.
- `--debounce-consecutive-observations <count>` (alias `--status-debounce-consecutive`, default 3) records pane-status classification threshold.
- Resolved values are written to `.intent-cli/supervision/<domain>/<team>/emission-policy.json`; active parked records expose `first_seen`, `last_seen`, `repeat_count`, and `emission_cadence_seconds`.

Guide reachability:
- `intent-cli guide orchestrator-thread --domain <domain> --target-repo <owner/repo> --agent claude --format json` exposes `design_workspace_supervision.emission_hygiene`.
- The regression runs the built CLI from a bare metadata-free directory and checks the commands, recorded configuration, operating semantics, negative checks, and absence of `.intent-cli/config.toml`.

Verification:
- Focused G699/G581/G586/G695/G659: 33 passed.
- Guide/Japanese fragment matrix: 164 passed.
- Full `IntentSystem.Cli.Tests`: 5014 passed, 1 skipped, 5015 total.
- `git diff --check`: clean.

Exact implementation head: `6990fec7`.


tomohisa (commented):
## Review verdict: REQUEST-UPDATE

Reviewed at exact head `6990fec775257cf9f68391b9f53a889c4cefff6a` against issue **#1511** as the standalone contract. Merge base `86f1ffdf`; `Closes #1511` present; base `main`; `git diff --check` clean; scope confined to `src/IntentSystem.Cli/Commands`, `tests/`, `docs/en`, `docs/ja`; exact-head CI `Build and test (source contract)` **success**; no automated reviewer comments.

The backoff/park half of this unit is well built and I verified it end to end. **One blocking correctness finding**: the status debounce does not delay classification — with a realistic herdr state-change sequence it **suppresses it permanently**, so a genuinely blocked seat is never named. The existing debounce test passes only because its fixture advances `state_change_sequence` on every poll, including polls where nothing changed.

Submitted as `COMMENTED` per the same-account convention.

---

## Blocking finding

### `implementation-finding` — the debounce never classifies a sustained blocked seat; it only suppresses

**Failed criterion.** Issue #1511 Verification: *"Test: 1-poll flap → no classification; threshold-consecutive → classification."* The first half holds; the second does not in production.

**Defect.** In `NotifyMeasuredSupervisor.ReadRecordedSeatTransitions`, the observed sequence is recorded **before** the gate:

```csharp
var seatKey = $"seat-state:{workspaceId}:{recorded.PaneId}";
observedSequences[seatKey] = sequence;          // written unconditionally
...
if (role is not ("implementation" or "review")
    || !string.Equals(runFrom, "working", StringComparison.Ordinal)
    || !settled
    || consecutiveCount < debounceConsecutiveObservations   // new G699 gate
    || priorSequence is null
    || sequence <= priorSequence.Value)                     // pre-existing guard
{
    continue;
}
```

`state_change_sequence` advances **once per state change**, not per poll. So a seat that goes `working → blocked` and stays blocked has exactly one new sequence value:

- poll 1 — new sequence `S2`, `consecutiveCount = 1` → suppressed by the new debounce gate, **but `observedSequences[seatKey] = S2` is already persisted to the cycle**;
- poll 2 — sequence still `S2`, `priorSequence = S2`, `consecutiveCount = 2` → suppressed;
- poll 3 — `consecutiveCount = 3` satisfies the threshold, but `sequence <= priorSequence` is now true → **`continue` forever**.

The debounce consumes the only sequence advance it needed in order to fire later.

**Measured on this head.** A review probe reproduced the shipped debounce scenario with the sequence held constant across the sustained blocked state (one real state change), 8 polls, threshold 3:

```
PROBE sustained-blocked constant-sequence: classified_on_polls=[]  consecutive_count=8
```

Never classified, with the threshold exceeded more than twice over. Control run, identical fixture, only `debounceConsecutiveObservations: 1` (pre-G699-equivalent):

```
PROBE sustained-blocked constant-sequence: classified_on_polls=[1]  consecutive_count=8
```

So this is a regression introduced by this PR, not a pre-existing gap: before G699 the transition was classified on the first poll (the false-positive behaviour the unit targets); after G699 it is classified **never**.

**Why the suite is green.** `OnePollFlapIsNotClassified_ButThresholdConsecutiveSettledStatusIs_G699` advances the sequence on every poll, including the three consecutive `blocked` polls:

```csharp
status = "blocked"; sequence++;   // poll A
sequence++;                       // poll B — no state change, yet sequence moves
sequence++;                       // poll C — classification asserted here
```

A *state-change* counter that increments while the state is unchanged is not a state-change counter. The fixture, not the product, is what makes the threshold reachable.

**Consequence for the adjacent G695 surface.** `RecordSettledTransitionChain(...)` is only reached for a `seat-state-transition` observation. Because no such observation is ever produced under a realistic sequence, the G695 "observed working→settled transition seeds the chain" path is silently disabled — and issue #1511 lists the continuation-chain surfaces as **out of scope / unchanged**. In my probe the chain never seeded.

**Net effect.** This is the G626 alarm-honesty failure in the opposite direction from the one the unit set out to fix: the false positive is gone because the true positive is gone too. A real permission stall now produces no finding at all.

**Required unblock action.**
1. Do not let a debounce-suppressed poll consume the sequence advance — e.g. only write `observedSequences[seatKey]` once the observation is actually emitted, or carry the pending "unclassified advance" forward (a `pending_sequence` alongside the existing `run_from`/consecutive-count carry-forward, which already solves the analogous problem for `priorStatus`).
2. Change the regression so the sequence advances **only on real state changes**, and assert classification fires on the threshold-th consecutive poll with a constant sequence. As written the test cannot fail for this defect.

---

## Verified and holding

### Guide reachability (treated as acceptance substance)

A CLI built from this head, run in a **metadata-free bare directory** (no `.intent-cli/`), exit 0, no config created:

```
$ intent-cli guide orchestrator-thread --domain intent-cli --target-repo … --agent claude --format json
design_workspace_supervision.emission_hygiene:
  summary            — full-cadence first finding; unchanged same-key stays a named parked record
  commands[]         — notify supervise … --interval 300 --repeat-backoff-seconds 1800
                       --debounce-consecutive-observations 3 --once --write
  recorded_configuration[] — flags, aliases, defaults (1800s / 3), and the
                       .intent-cli/supervision/<domain>/<team>/emission-policy.json write location
  operating_semantics[]    — park names first_seen/last_seen/repeat_count/emission_cadence_seconds;
                       resolution clears; fingerprint change resets; a new key is immediate;
                       sub-threshold flap is not classified
```

The declared operating recipe is reachable and matches the shipped flags. Note that `operating_semantics` currently promises "a pane status flap below the recorded consecutive threshold is not classified" — true — while the *above*-threshold half is what the finding above breaks, so the guide over-promises until it is fixed.

### Backoff, park and reset (AC#1, AC#3)

`RefreshEmissionState` computes a state fingerprint over the observation's canonical fields and derives `first_seen` / `last_seen` / `repeat_count` / `parked` / `emission_cadence_seconds` / `last_emitted_at`; emission is gated on `stateChanged || first observation || now - last_emitted >= backoff`. Verified by the shipped fixtures, which I ran and read:

- **Ten identical observations** at 10 s with a 20 s recorded backoff → **5 emissions** (t=0, 20, 40, 60, 80), `first_seen` = t0, `last_seen` = t0+90, `repeat_count` = 10, `parked` = true, `emission_cadence_seconds` = 20. Reduced emission is a *named park*, not silence.
- **Resolution** clears the active record; a reappearance starts a fresh window.
- **Condition/fingerprint change** resets `repeat_count` to 1, moves `first_seen` to now, clears `parked`, and emits immediately — full cadence restored.
- **New-key immediacy**: a genuinely interleaved new key emits on its first observation while another key is parked; there is no global emission gate, only per-record state.

### Policy is recorded configuration, not buried constants (AC#2)

`--repeat-backoff-seconds` (alias `--backoff-seconds`, default 1800) and `--debounce-consecutive-observations` (alias `--status-debounce-consecutive`, default 3) are CLI flags; `NotifySupervisionEmissionPolicy` names the defaults and a maximum; resolution order is explicit flag → previously recorded policy → default; resolved values are persisted to `emission-policy.json`, re-read on the next cycle, echoed on every `cycles.jsonl` entry, and returned on the pass. The fixture asserts the file exists and contains both keys.

### Detection predicates unchanged

The escalation conditions themselves are untouched — this PR changes emission bookkeeping (`RefreshEmissionState`, `PersistStallUpdate`) and adds the seat-status debounce gate. No new escalation kinds; no auto-resolution of parked conditions. The G695 continuation-chain code is unmodified — though see the finding above for its behavioural collateral.

### EN/JA parity (AC#4) and scope

Marker counts are identical across `docs/en` and `docs/ja`: `repeat-backoff-seconds` 2/2, `debounce-consecutive-observations` 2/2, `emission-policy.json` 2/2, `repeat_count` 2/2, `first_seen` 2/2, `G699` 2/2. Both compatibility ledgers gained the matching `repeated-observation emission hygiene` row. 14 changed files, all within the issue's declared target paths.

## Test evidence

Detached worktree at exact head `6990fec7`, isolated artifacts path:

```
--filter FullyQualifiedName~NotifySupervisionG699Tests
Passed  OrchestratorGuideNamesRecordedPolicyAndRunsFromMetadataFreeBareDirectory_G699 [281 ms]
Passed  OnePollFlapIsNotClassified_ButThresholdConsecutiveSettledStatusIs_G699        [119 ms]
Passed  TenSameKeyObservationsUseRecordedCadenceAndExposeParkedCounters_G699          [22 ms]
Passed  ResolutionAndConditionChangeResetSameKeyState_AndNewKeyIsImmediate_G699       [6 ms]
Failed  EnglishAndJapaneseGuidanceDeclareTheSameEmissionPolicy_G699                   [2 ms]
```

The single local failure is `RepoVersionPolicySource.RepoRoot()` → `Could not locate the repository root`, an artifact of the out-of-tree artifacts path this environment requires; it passes in CI. I substituted it by grepping both doc trees directly (parity table above).

**Negative check (measured).** Collapsing only the backoff comparison to zero at this head fails `TenSameKeyObservationsUseRecordedCadenceAndExposeParkedCounters_G699` — the backoff guard is genuinely exercised. By contrast, the probe above shows the debounce guard has **no** test that can fail for the defect it contains. The worktree was restored to the exact head after both experiments and is clean.

- The full suite was not re-run locally; exact-head CI is green, but note that CI green does not cover the finding above — the defect is invisible to the shipped fixture.
- No live herdr workspace was involved; the debounce behaviour was reproduced through the PR's own fixture harness with a realistic sequence, not against a real pane.
- No shared build process was terminated, no `/tmp` path was deleted (fresh uniquely named scratch only), and nothing was pushed, merged, or label-mutated. The PR remains a draft; promotion, approval label, merge, and closeout stay with orchestration.

tomohisa (commented):
## Re-review disposition: APPROVED

Re-reviewed at exact repaired head `65ae7ee6cf25853dc1bf8bf46e3dfce2bb753bad`, superseding the REQUEST-UPDATE at `6990fec7` (#pullrequestreview-4941976511). The blocking finding is fixed, and I confirmed the fix against production code rather than against the test's own model.

Exact-head CI `Build and test (source contract)` is **success**; head unchanged before and after this review; `Closes #1511`; base `main`; `git diff --check` against merge base `86f1ffdf` clean; review worktree clean at the exact head.

Submitted as `COMMENTED` per the same-account convention.

---

## Delta from the rejected head

```
$ git diff --stat 6990fec7 65ae7ee6
 src/IntentSystem.Cli/Commands/NotifyMeasuredSupervisor.cs | 23 ++++--
 tests/IntentSystem.Cli.Tests/NotifySupervisionG699Tests.cs | 83 +++++++++++++---
 2 files changed, 90 insertions(+), 16 deletions(-)
```

Two files. The production change is confined to the seat-transition block; docs, the emission-policy store, `RefreshEmissionState`, and the guide fragments are untouched, so the previously-approved backoff/park/reset/new-key/config/EN-JA behaviour carries forward structurally — and its tests still pass (below).

## Finding — RESOLVED: the debounce now delays classification instead of consuming it

`observedSequences[seatKey] = sequence` is no longer unconditional. It is written when the seat is **not** a pending settled transition (normal advance) **or** when the consecutive threshold has been reached (consume on emit); a debounce-suppressed poll leaves it alone, so the single state-change advance survives the debounce window. The gate is now expressed as `pendingSettledTransition` plus the threshold, which also makes the two concerns readable separately.

### Measured on production code, not on a model

An independent review probe drove the supervisor with **one** real state change (`working → blocked`, sequence 1 → 2) and then held the sequence constant, threshold 3, five polls:

```
baseline (working): seqEntry=1                       ← non-candidate state advances normally
poll1: classified=False count=1 status=blocked runFrom=working seqEntry=ABSENT
poll2: classified=False count=2 status=blocked runFrom=working seqEntry=ABSENT
poll3: classified=True  count=3 status=blocked runFrom=working seqEntry=2   ← consumed on emit
poll4: classified=False count=4 status=blocked runFrom=working seqEntry=2
poll5: classified=False count=5 status=blocked runFrom=working seqEntry=2
chainRecords=1 links=report-received, orchestration-wake-attempted, wake-delivered-or-observed
```

Every item the re-review asked for is confirmed:

- the transition advances the sequence **once** and stays constant;
- polls 1 and 2 are suppressed; **poll 3 classifies at threshold 3**;
- `status`, `consecutive count`, and `run_from` persist correctly across suppressed polls (`runFrom` stays `working` throughout, which is what keeps the working→settled precondition alive);
- the sequence is **consumed when emitted** (`seqEntry=2` from poll 3 onward), so polls 4–5 do not re-emit — the fix does not trade a suppressed finding for a repeated one;
- the **G695 continuation chain is seeded**: one record with `report-received` (source `herdr-state-transition`), `orchestration-wake-attempted`, and `wake-delivered-or-observed`. The adjacent surface issue #1511 declares unchanged is working again;
- baseline / non-candidate states still record their sequence normally.

### The repaired regression is a real guard, not a self-fulfilling simulation

I restored the **pre-repair ordering in production code** (`observedSequences[seatKey] = sequence` written unconditionally) at this head and re-ran the suite:

```
[FAIL] OnePollFlapIsNotClassified_ButConstantSequenceThresholdClassifiesAndSeedsG695Chain_G699
→ 4 passed, 2 failed   (the second failure is the known docs/repo-root artifact)
```

So the rewritten regression genuinely fails when the defect returns. It now holds the sequence constant across the three sustained polls, asserts classification on the third, asserts `LastObservedStateChangeSequences` equals the consumed sequence, and asserts the G695 chain links — it can no longer pass on a fixture that invents a sequence advance per poll.

## Re-confirmed unchanged by the delta

```
Passed  OrchestratorGuideNamesRecordedPolicyAndRunsFromMetadataFreeBareDirectory_G699 [263 ms]
Passed  OnePollFlapIsNotClassified_ButConstantSequenceThresholdClassifiesAndSeedsG695Chain_G699 [136 ms]
Passed  TenSameKeyObservationsUseRecordedCadenceAndExposeParkedCounters_G699          [32 ms]
Passed  ResolutionAndConditionChangeResetSameKeyState_AndNewKeyIsImmediate_G699       [5 ms]
Passed  PreRepairSequenceOrderingConsumesConstantSequenceAndNeverClassifies_G699      [11 ms]
Failed  EnglishAndJapaneseGuidanceDeclareTheSameEmissionPolicy_G699                   [< 1 ms]
```

Backoff/park (10 observations → 5 emissions, `first_seen`/`last_seen`/`repeat_count=10`/`parked`/cadence), resolution and fingerprint reset, new-key immediacy, and the recorded `emission-policy.json` configuration all still pass unmodified. The single failure is `RepoVersionPolicySource.RepoRoot()` under this environment's out-of-tree artifacts path; it passes in CI.

Affected-area net: **317 passed, 18 failed** across `Supervis|ContinuationChain|Herdr|NotifyEvent|OrchestratorThread`. All 18 classified exhaustively — 17 × `Could not locate the repository root`, 1 × `Could not locate docs/en/12-agent-message-orchestration.md`. **Zero assertion failures.**

### Guide primacy

A CLI built from this head, run in a metadata-free bare directory (no `.intent-cli/`), exit 0, no config created. `design_workspace_supervision.emission_hygiene` still carries 2 commands, 4 `recorded_configuration` entries, and 6 `operating_semantics` entries, including the two that matter here:

> a pane status flap below the recorded consecutive threshold is not classified; **the threshold-consecutive settled state is classified once** with the existing advance
>
> detection predicates and **G695 continuation-chain recording remain unchanged**

The guide no longer over-promises: the above-threshold half it declares is now the behaviour the code delivers, which was the gap I flagged last round.

## Informational (no change requested)

1. **The comment slightly overstates what is retained.** During the debounce window the seat's key is *absent* from `LastObservedStateChangeSequences` (probe: `seqEntry=ABSENT`), not "retained at the prior sequence". The emit decision still works because `GetValueOrDefault` yields `0` and `sequence > 0` holds. That is correct for herdr sequences, but it is an implicit dependency on sequences never being `0`; writing the prior sequence explicitly would match the comment and remove the assumption.

2. **`PreRepairSequenceOrderingConsumesConstantSequenceAndNeverClassifies_G699` is documentation, not a guard.** It re-implements the old ordering in a private test helper, so it is insensitive to production code — it **passed** during my experiment above while the defect was restored in the supervisor. The real protection is the main regression. Consider renaming it to say it models the historical ordering, or dropping it, so a future reader does not mistake it for coverage.

- The full suite was not re-run locally; exact-head CI success is the authority for suite-wide green.
- No live herdr workspace was involved; the sequence semantics were exercised through the PR's fixture harness with a realistic constant sequence, not against a real pane.
- The previously-approved EN/JA doc parity was not re-derived at this head — the delta touches no documentation, and that reasoning is stated rather than implied.
- No shared build process was terminated, no `/tmp` path was deleted (fresh uniquely named scratch only), and nothing was pushed, merged, or label-mutated. The PR remains a draft; promotion, approval label, merge, and closeout stay with orchestration.

PR #1539, opened 2026-08-16
G711: publish npm through Trusted Publishing (OIDC)

### LINKED ISSUE #1538: G711: publish to npm through Trusted Publishing (OIDC)
# G711: publish to npm through Trusted Publishing (OIDC)

## Goal

Authenticate the npm publish with **GitHub Actions OIDC** instead of a long-lived token, hold **no npm secret** in the repository, and make a publish that cannot authenticate **fail named** rather than silently skip.

## Why This Slice Exists Now

Operator decision (2026-08-15), taken at the token-creation screen: the offered token expiry was 2026-11-14 — **a credential whose expiry would silently break a release about three months later**. npm's own banner warns that tokens bypassing 2FA are being restricted (account changes Aug 2026, direct publishing Jan 2027), and the UI states plainly: *"For automation or CI/CD uses, please use Trusted Publishing instead."*

npm documentation (verified 2026-08-15) gives the requirements: the publishing job needs **`id-token: write`**, the runtime needs **npm CLI ≥ 11.5.1 and Node ≥ 22.14.0**, the trusted publisher is registered **per package** on npmjs.com by naming the GitHub repository and workflow file, and **provenance attestations are generated automatically** for GitHub Actions publishes.

There is a second defect to fix in the same step. The current publish step reads:

```
if [ -z "${NODE_AUTH_TOKEN:-}" ]; then
  echo "NPM_TOKEN is not set (fork or missing secret); skipping npm publish."
  ...
  exit 0
fi
```

That is exactly the shape this repository just shipped G708 to eliminate: **a step that publishes nothing while reporting success**. In the release pipeline it means a release can complete "successfully" with nothing on the registry — which is what happened for v0.22.0, and was only visible because the notes said so explicitly.

The release gate is unchanged: publishing still runs only on the release event, inside the operator-approved transaction, with tag-derived versions. What changes is **how the job proves who it is** — from a stored secret to a per-run proof — and that failure becomes loud.

Operator account actions are **out of scope for this unit**: the `j-tech-japan` organisation exists; the first publish of the four packages and the per-package trusted-publisher registration are operator steps.

## Current Observed State

- `release.yml`: top-level `permissions: contents: write`; npm job uses `actions/setup-node@v4` with `node-version: 22` and `registry-url: https://registry.npmjs.org`.
- Publish step passes `NODE_AUTH_TOKEN` from `secrets.NPM_TOKEN`, and exits 0 with a message when the secret is absent.
- No `NPM_TOKEN` secret is configured, and none will be.

- The npm requirements quoted above are current as documented on 2026-08-15.
- Packing, checksum verification and release-asset attachment already work and must keep working for every run, including dry runs.
- The four packages are `intent-system` plus `@j-tech-japan/intent-cli-{darwin-arm64,linux-x64,win32-x64}`.

## Target Repo / Path / Part

Target paths: `.github/workflows/release.yml, docs/en, docs/ja`

Target part: OIDC-authenticated npm publish.

## In Scope

- Job-scoped `id-token: write` on the npm publish job; publish through trusted publishing.
- Remove every npm credential path: no `NPM_TOKEN`, no `NODE_AUTH_TOKEN`, no registry auth secret.
- Pin an npm CLI ≥ 11.5.1 on Node ≥ 22.14.0, with the versions stated in the workflow rather than inherited implicitly from the runner.
- Replace the silent skip with a **named failure** identifying the package that could not be published and the missing configuration.
- Keep the release gate, tag-derived versions, packing, checksums and asset attachment unchanged.
- EN/JA docs: the one-time per-package trusted-publisher setup an operator performs, and the statement that no npm secret is stored.

## Out Of Scope

- Creating the npm organisation, publishing the first version of any package, or registering trusted publishers (operator account actions).
- Any credential handling.
- Changing NuGet publication or the binary release path.
- Changing what the packages contain.

The child PR changes the release workflow so the npm publish job authenticates through GitHub Actions OIDC trusted publishing, granting that job the id-token write permission without widening permissions elsewhere, and removes every npm credential path so that no npm token, node auth token, or registry secret is referenced anywhere in the workflow. It pins an npm command-line version of at least 11.5.1 running on Node of at least 22.14.0 and states those versions explicitly rather than relying on the runner default. It replaces the branch that printed a skip message and exited zero when a credential was absent with a named failure that identifies which package could not be published and what configuration is missing, so a release can no longer report success while publishing nothing. The release gate is unchanged: publication still occurs only on the release event with tag-derived versions inside the operator-approved transaction, and packing, checksum verification and release-asset attachment continue to run for every execution including dry runs. English and Japanese documentation describe the one-time per-package trusted-publisher registration that an operator performs by naming the repository and workflow file, note that provenance attestations are generated automatically for GitHub Actions publishes and can be verified by consumers, and state that the repository stores no npm secret. Documentation stays in parity under the G613 terminology policy and the full suite is green.

## Acceptance Criteria

- npm publish job uses OIDC with job-scoped `id-token: write`; zero npm secret references remain in the workflow.
- npm ≥ 11.5.1 on Node ≥ 22.14.0, versions stated in the workflow.
- Authentication failure is a named job failure naming the package and the missing configuration; no exit-zero skip path.
- Release gate, tag-derived versions, packing, checksums and asset attachment unchanged.
- EN/JA docs cover the operator's one-time setup and the no-stored-secret statement.
- Full suite green.

## Verification

- grep the workflow: no `NPM_TOKEN`, `NODE_AUTH_TOKEN`, or registry credential remains; `id-token: write` present on the publish job only.
- grep the workflow: pinned npm and Node versions meet the documented minimums.
- Inspect the publish step: the no-credential path exits non-zero with a named message; assert no `exit 0` skip remains on that branch.
- Dry-run path still packs, verifies checksums and attaches assets without attempting to publish.

## Related Links

- npm trusted publishing documentation (requirements verified 2026-08-15); the npm token screen's own CI/CD guidance and the 2FA-bypass restriction banner.
- G702 (npm channel), G708 (an output naming an action must describe the action taken), G685 (facts you store go stale; methods you re-derive do not).

## Knowledge Maintenance

- Intent placement: node 02 primary (distribution/release flow).
- ADR candidate: none — this applies an existing principle to release authentication.
- Diagram candidate: none.
- Docs update: EN/JA (in scope).
- Closeout writeback expected: yes — "a stored credential is a fact that rots on a timer; a derived credential is re-proved at every use."

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1538. Adds job-scoped GitHub Actions OIDC for npm publication, pins Node 22.14.0 and npm 11.5.1, removes npm token auth paths, and makes missing trusted-publisher configuration fail with the package name. Adds EN/JA operator guidance and regressions. Verification: focused workflow/npm tests 15 passed; full Release solution 5,105 passed, 1 skipped; actionlint and git diff --check passed. No npm publish, credential, tag, or Release mutation performed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,7 +17,7 @@
 #
 #   3. Prepares the npm entry package plus its per-platform optional packages,
 #      verifies checksums/version identity, and publishes them only on the same
-#      real-release event and guarded secret path as NuGet.
+#      real-release event through GitHub Actions OIDC trusted publishing.
 #
 #   4. Attaches the .nupkg, npm packages, and per-platform binary archives
 #      (plus checksums) to the triggering GitHub Release.
@@ -289,6 +289,9 @@ jobs:
     needs: [nupkg, binaries]
     runs-on: ubuntu-latest
     timeout-minutes: 30
+    permissions:
+      contents: write
+      id-token: write
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -298,8 +301,16 @@ jobs:
       - name: Setup Node.js for release packaging
         uses: actions/setup-node@v4
         with:
-          node-version: 22
+          node-version: 22.14.0
           registry-url: https://registry.npmjs.org
+
+      - name: Pin npm CLI for trusted publishing
+        shell: bash
+        run: |
+          set -euo pipefail
+          npm install --global npm@11.5.1
+          test "$(node --version)" = "v22.14.0"
+          test "$(npm --version)" = "11.5.1"
 
       - name: Resolve the same tag-derived release version as NuGet and binaries
         id: ver
@@ -384,22 +395,32 @@ jobs:
             .artifacts/npm-tarballs/* \
             --repo "${{ github.repository }}" --clobber
 
-      - name: Publish npm packages (guarded; same release gate as NuGet)
+      - name: Publish npm packages through OIDC trusted publishing
         if: ${{ github.event_name == 'release' }}
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
         shell: bash
         run: |
           set -euo pipefail
-          if [ -z "${NODE_AUTH_TOKEN:-}" ]; then
-            echo "NPM_TOKEN is not set (fork or missing secret); skipping npm publish."
-            echo "Packages were still packed, checksum-verified, and attached to the release."
-            exit 0
-          fi
-          for package in \
+          publish_package() {
+            local package_path="$1"
+            local package_name="$2"
+            local publish_output
+            if ! publish_output="$(npm publish "${package_path}" --access public 2>&1)"; then
+              printf '%s\n' "${publish_output}" >&2
+              echo "npm trusted publishing authentication failed for package ${package_name}: register this package's npmjs.com trusted publisher for repository ${GITHUB_REPOSITORY} and workflow .github/workflows/release.yml, and keep id-token: write on the npm job; no npm token fallback is supported." >&2
+              return 1
+            fi
+            printf '%s\n' "${publish_output}"
+          }
+
+          publish_package \
             .artifacts/npm-tarballs/j-tech-japan-intent-cli-darwin-arm64-*.tgz \
+            "@j-tech-japan/intent-cli-darwin-arm64"
+          publish_package \
             .artifacts/npm-tarballs/j-tech-japan-intent-cli-linux-x64-*.tgz \
+            "@j-tech-japan/intent-cli-linux-x64"
+          publish_package \
             .artifacts/npm-tarballs/j-tech-japan-intent-cli-win32-x64-*.tgz \
-            .artifacts/npm-tarballs/intent-system-*.tgz; do
-            npm publish "${package}" --access public
-          done
+            "@j-tech-japan/intent-cli-win32-x64"
+          publish_package \
+            .artifacts/npm-tarballs/intent-system-*.tgz \
+            "intent-system"
--- a/docs/en/13-npm-distribution.md
+++ b/docs/en/13-npm-distribution.md
@@ -73,6 +73,37 @@ NuGet. Pull-request CI runs package preparation, `npm pack`, checksum/version
 verification, and a packed-install smoke test; it never publishes to npm and
 does not require npm organization credentials.
 
+## Trusted publishing for releases (G711)
+
+The npm publish job runs only for a published GitHub Release, after the same
+tag-derived packaging, checksum verification, and release-asset steps used by
+the other distribution jobs. It authenticates with GitHub Actions OIDC
+trusted publishing. The job is explicitly pinned to Node `22.14.0` and npm
+`11.5.1`; it has job-scoped `id-token: write` permission and stores no npm
+token or registry credential in this repository.
+
+Before the first release, an operator must perform the one-time npmjs.com
+trusted-publisher registration separately for each package:
+
+- `intent-system`
+- `@j-tech-japan/intent-cli-darwin-arm64`
+- `@j-tech-japan/intent-cli-linux-x64`
+- `@j-tech-japan/intent-cli-win32-x64`
+
+For each package, select GitHub Actions as the trusted publisher and register
+the `J-Tech-Japan/intent-system` repository with the workflow file
+`.github/workflows/release.yml`. These are operator account actions; this
+repository neither creates the npm organization nor stores a publish secret.
+Successful GitHub Actions trusted publishes receive npm provenance
+attestations automatically, which consumers can verify from the published
+package metadata.
+
+If OIDC authentication or the per-package registration is missing, the
+release job fails with the package name and the missing trusted-publisher or
+`id-token: write` configuration. It never reports a successful skip. A
+workflow-dispatch dry run still prepares, packs, checksums, and verifies the
+packages without attempting publication.
+
 ## Coexisting with the .NET tool
 
 The npm route and the .NET global tool can coexist:
--- a/docs/ja/13-npm-distribution.md
+++ b/docs/ja/13-npm-distribution.md
@@ -65,6 +65,34 @@ npm パッケージ、ランタイム同梱バイナリ、およびバイナリ�
 Pull request CI は package preparation、`npm pack`、checksum/version 検証、packed-install smoke test
 までを実行します。npm への公開は行わず、npm 組織の credential も必要としません。
 
+## リリースの Trusted publishing (G711)
+
+npm publish job は公開済み GitHub Release の場合だけ動き、他の配布 job と同じ
+tag-derived version、package の checksum 検証、Release asset attachment を先に完了してから実行します。
+認証には GitHub Actions OIDC trusted publishing を使います。job は Node `22.14.0` と npm
+`11.5.1` を明示的に固定し、job 単位の `id-token: write` permission だけを追加します。
+repository に npm token や registry credential は保存しません。
+
+最初の release の前に、operator は npmjs.com で次の各 package に対して一度だけ
+trusted-publisher registration を行う必要があります。
+
+- `intent-system`
+- `@j-tech-japan/intent-cli-darwin-arm64`
+- `@j-tech-japan/intent-cli-linux-x64`
+- `@j-tech-japan/intent-cli-win32-x64`
+
+各 package で GitHub Actions を trusted publisher として選び、
+`J-Tech-Japan/intent-system` repository と workflow file
+`.github/workflows/release.yml` を登録します。これは operator の account action であり、
+この repository が npm 組織を作成したり publish secret を保存したりすることはありません。
+GitHub Actions の trusted publish が成功すると npm provenance attestation は自動生成され、
+consumer は公開 package metadata から検証できます。
+
+OIDC 認証または package ごとの registration が不足している場合、release job は対象 package 名と
+不足している trusted-publisher または `id-token: write` configuration を示して fail します。
+成功した skip として報告することはありません。workflow-dispatch の dry run は publication を試みず、
+package の prepare、pack、checksum、verification だけを続けます。
+
 ## .NET tool との共存
 
 npm 経路と .NET global tool は共存できます。
--- a/tests/IntentSystem.Cli.Tests/NpmDistributionG702Tests.cs
+++ b/tests/IntentSystem.Cli.Tests/NpmDistributionG702Tests.cs
@@ -79,7 +79,12 @@ public sealed class NpmDistributionG702Tests
         Assert.Contains("github.event.release.tag_name", release, StringComparison.Ordinal);
         Assert.Contains("VERSION=\"${RAW#v}\"", release, StringComparison.Ordinal);
         Assert.Contains("needs: [nupkg, binaries]", release, StringComparison.Ordinal);
-        Assert.Contains("NPM_TOKEN", release, StringComparison.Ordinal);
+        Assert.DoesNotContain("NPM_TOKEN", release, StringComparison.Ordinal);
+        Assert.DoesNotContain("NODE_AUTH_TOKEN", release, StringComparison.Ordinal);
+        Assert.Contains("id-token: write", release, StringComparison.Ordinal);
+        Assert.Contains("node-version: 22.14.0", release, StringComparison.Ordinal);
+        Assert.Contains("npm@11.5.1", release, StringComparison.Ordinal);
+        Assert.Contains("OIDC trusted publishing", release, StringComparison.Ordinal);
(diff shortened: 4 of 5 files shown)


tomohisa (commented):
## Review verdict: APPROVED

Reviewed at exact head `1392561d7b13a19d559d7a6cb5ab83f72a7b7f76` against issue **#1538** and the G711 review context. `Closes #1538`; base `main`; 5 files (+233/−15); head unchanged before and after this review; worktree clean at the exact head.

Exact-head CI, both required checks green (run `31924106353`, `head_sha` = `1392561d`):

```
Build and test (source contract)     pass  2m12s
npm package dry-run (no publish)     pass  1m18s
```

Every acceptance criterion holds, and I verified the two that matter most — the named non-zero failure and the permission scoping — by **executing the shipped shell** and by **parsing the workflow**, not by reading it. Two non-blocking findings below; neither contradicts an AC.

---

## AC-by-AC

### 1. OIDC with job-scoped `id-token: write`; zero npm secret references ✅

Verified structurally by parsing the YAML rather than grepping prose, so a match inside a comment or an echo string cannot fake it:

```
top-level permissions        {'contents': 'write'}
job nupkg    -> permissions  None
job binaries -> permissions  None
job npm      -> permissions  {'contents': 'write', 'id-token': 'write'}
```

`npm` is the only job that carries `id-token`, and it is the last of the three jobs. `contents: write` is re-declared on the job because a job-level block replaces the top-level grant and `gh release upload` still needs it — correct, not redundant.

Credential paths are gone repository-wide, not just from the diff:

```
$ grep -rn "NPM_TOKEN\|NODE_AUTH_TOKEN" .          (excluding .git)
tests/…/NpmDistributionG702Tests.cs       (DoesNotContain assertions)
tests/…/NpmTrustedPublishingG711Tests.cs  (DoesNotContain assertions + guard fixture)
docs/en|ja/09-developer-reference.md      (v0.22.0 historical record)
docs/en|ja/release-notes-v0.22.0.md       (v0.22.0 historical record)
→ zero live references in any workflow or under src/
```

The remaining doc hits are the frozen v0.22.0 record ("The **v0.22.0** npm registry publication remains skipped … `NPM_TOKEN` is absent"), explicitly scoped to a released version, so G711 does not make them stale. The workflow's only secret reference at all is `NUGET_API_KEY` in the `nupkg` job, unchanged.

```
$ grep -n "secrets\." .github/workflows/release.yml
132:          NUGET_API_KEY: ${{ secrets.NUGET_API_KEY }}
$ actionlint .github/workflows/release.yml    → clean
$ git diff --check origin/main...HEAD          → clean
```

### 2. npm ≥ 11.5.1 on Node ≥ 22.14.0, stated in the workflow ✅

```yaml
node-version: 22.14.0
- name: Pin npm CLI for trusted publishing
  run: |
    npm install --global npm@11.5.1
    test "$(node --version)" = "v22.14.0"
    test "$(npm --version)" = "11.5.1"
```

Worth calling out: the pins are not merely *declared*, they are **asserted at runtime**. If the runner or setup-node ever delivers something else, the job fails on the `test` line before it can attempt a publish with an unsupported npm. That is a stronger construction than the AC asked for.

### 3. Named non-zero failure, no exit-zero skip ✅ — executed, not read

I extracted the publish step's `run:` script from the workflow and ran it against a stub `npm` on `PATH`, with the four tarball names present.

**Authentication failure (E401):**

```
EXIT=1
stderr:
  npm error code E401
  npm error need auth This command requires you to be logged in
  npm trusted publishing authentication failed for package @j-tech-japan/intent-cli-darwin-arm64:
  register this package's npmjs.com trusted publisher for repository J-Tech-Japan/intent-system
  and workflow .github/workflows/release.yml, and keep id-token: write on the npm job;
  no npm token fallback is supported.
stdout: (empty)
```

Non-zero, names the exact package, names both missing configurations (per-package trusted publisher *and* `id-token: write`), and states no token fallback exists. It also fails on the **first** package rather than continuing — so a release cannot end with a partially published set. The raw npm output is printed *before* the interpretation, so the operator sees the registry's own words first.

**Success path:**

```
EXIT=0
+ .artifacts/npm-tarballs/j-tech-japan-intent-cli-darwin-arm64-0.23.0.tgz
+ .artifacts/npm-tarballs/j-tech-japan-intent-cli-linux-x64-0.23.0.tgz
+ .artifacts/npm-tarballs/j-tech-japan-intent-cli-win32-x64-0.23.0.tgz
+ .artifacts/npm-tarballs/intent-system-0.23.0.tgz
```

All four, with the three platform packages before the `intent-system` entry package — the order its `optionalDependencies` require.

The shell is also written correctly for `set -e` propagation: `local publish_output` is declared on its own line before the assignment, so the function's return status is the `npm publish` status rather than `local`'s. A `return 1` from the function therefore does terminate the step. That is the subtle way this pattern usually breaks, and it is done right here.

```
$ grep -n "exit 0" .github/workflows/release.yml
139:            exit 0        ← NuGet job, unchanged from base, out of scope for G711
```

No `exit 0` remains anywhere in the npm job.

### 4. Release gate, tag-derived versions, packing, checksums, assets unchanged ✅

The diff touches exactly four hunks in the workflow (`@@ -17`, `@@ -289`, `@@ -298`, `@@ -384`). Step-by-step comparison against `origin/main`:

| step | base | head |
|---|---|---|
| Checkout | ✓ | ✓ |
| Setup Node.js | `node-version: 22` | `node-version: 22.14.0` |
| **Pin npm CLI** | — | **added** |
| Resolve tag-derived version | ✓ | unchanged |
| Download binary artifacts | ✓ | unchanged |
| Rebuild npm platform packages | ✓ | unchanged |
| Pack and verify without publishing | ✓ | unchanged |
| Upload workflow artifact | ✓ | unchanged |
| Attach to GitHub Release | ✓ | unchanged |
| Publish | guarded token skip | **OIDC trusted publishing** |

One step added, one rewritten, one line changed. Packing, `verify-packages.js --require-real-binaries`, `sha256sum` checksum generation and `gh release upload` are byte-identical. Both the attach and publish steps remain gated on `github.event_name == 'release'`, confirmed by parsing rather than grep:

```
attach step  if: ${{ github.event_name == 'release' }}
publish step if: ${{ github.event_name == 'release' }}   env: None
```

So the workflow-dispatch dry run still packs, checksums and verifies, and reaches neither attach nor publish — the documented dry-run behaviour.

### 5. EN/JA operator docs and parity ✅

I machine-checked twenty-one properties across both documents rather than eyeballing the sections:

```
gate · OIDC named · Node pin · npm pin · job-scoped id-token · no stored secret ·
one-time per package · repository named · workflow file named · operator account action ·
no org creation · provenance automatic · consumer-verifiable · named failure ·
never a successful skip · dry-run behaviour · all four package names

EN 21/21   JA 21/21   mismatches = 0
```

Provenance wording stays inside npm's documented automatic behaviour — "Successful GitHub Actions trusted publishes receive npm provenance attestations automatically, which consumers can verify from the published package metadata." It claims no flag, no verification command, and no attestation format. Consistently, the workflow passes **no** `--provenance` flag (`grep provenance .github/workflows/release.yml` → 0 hits), which is exactly right for trusted publishing where npm generates it itself. The JA mirror keeps technical terms in original form, matching the surrounding G613-style prose.

### 6. Guide primacy — `no_role_facing_surface=true` substantiated ✅

Treating guide reachability as acceptance evidence rather than garnish, I built the CLI at this head and executed the router from an empty directory:

```
$ dotnet IntentSystem.Cli.dll guide help --format json     → exit 0, 21,889 B
advertised routes: guide workflow task {init-host, intent-interview, packet-draft,
  issue-publish, implementation-loop, review-next-slice-loop, bug-to-intent-repair},
  guide worker pr-comment-fix, automation summary
npm / NPM_TOKEN / npm publish / trusted publishing mentions in guide output: 0
```

No guide route covers npm distribution, so none is made stale by this change and none needs to be added. The PR touches **0 files under `src/`** and 0 under `packaging/`, so there is no CLI surface change at all — the declaration is accurate, and the operator path for this work is the EN/JA `13-npm-distribution.md` section the PR adds.

### 7. Scope boundaries ✅

Five files: the workflow, two docs, one amended guard, one new guard. No org creation, no first publish, no publisher registration, no credential value anywhere, no NuGet or binary-path change, no package-content change, and no AI-provider invocation. The docs state the registration is an operator account action in both languages.

## Negative checks — the guards are real

Baseline at the exact head: `NpmTrustedPublishingG711Tests` + `NpmDistributionG702Tests` → **9 passed, 0 failed**. Then one mutation at a time, reverted after each:

| mutation | result |
|---|---|
| re-add `NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}` to the publish step | 2 FAIL — `Workflow_UsesJobScopedOidcAndPinnedNodeAndNpmWithoutCredentialPaths`, `Ci_IsPackOnlyAndRelease_…` |
| re-add the `exit 0` silent-skip branch | 3 FAIL — the two above plus `Workflow_FailsNamedForThePackageWhenOidcConfigurationIsMissing` |
| move `id-token: write` to top-level `permissions` | 1 FAIL — job-scoping guard |
| grant `id-token: write` to the `binaries` job | 1 FAIL — job-scoping guard |
| downgrade to `node-version: 22` and `npm@11.0.0` | 2 FAIL — pin guards in both classes |

Every shape the unit exists to prevent is rejected. The PR also ships its own `WorkflowGuardRejectsTheFormerExitZeroSilentSkipShape`, which injects the old skip text and asserts the contract check throws naming `exit 0` — a self-testing guard rather than a claim, and I confirmed the same shape fails on disk too.

## Non-blocking findings

### 1. Every publish failure is reported as an authentication failure

`publish_package` treats any non-zero `npm publish` as a trusted-publishing auth problem. Measured with a stub returning a publish conflict:

```
npm error code EPUBLISHCONFLICT
npm error publish fail Cannot publish over previously published version 0.23.0.
npm trusted publishing authentication failed for package @j-tech-japan/intent-cli-darwin-arm64:
register this package's npmjs.com trusted publisher …
```

The version already existed; nothing was wrong with the registration, but the message sends the operator to the npmjs.com trusted-publisher screen. This is the same genus as G708 — an output naming a cause it did not determine. It is **not blocking** because the raw npm output is printed immediately above, so no evidence is hidden and the true cause is one line away; the AC's actual requirement (named, non-zero, no silent skip) is met in every case.

Cheap refinement when convenient: branch on the captured output, e.g. keep the trusted-publisher message when it matches an auth/OIDC signature and otherwise emit `npm publish failed for package <name>; see the npm output above` — same package naming, no asserted cause.

### 2. The `id-token` scoping guard only inspects the text before the npm job

```csharp
Assert.DoesNotContain("id-token: write", workflow[..workflow.IndexOf("\n  npm:")], …);
```

This is complete today only because `npm` happens to be the last job. I added a trailing job carrying the permission and re-ran:

```
jobs: ['nupkg', 'binaries', 'npm', 'postpublish']
postpublish permissions: {'contents': 'write', 'id-token': 'write'}
    → Passed! 9 passed, 0 failed          ← not detected
```

A later job could silently acquire OIDC. The shipped state is correct — I proved by YAML parse that `npm` is the only holder — so this is guard durability, not a defect in what is being merged. Suggested strengthening in the same idiom: assert the permission appears in exactly one job's `permissions` block, or check the whole workflow *minus* the extracted npm job rather than only the prefix.

- No npm publish, org creation, trusted-publisher registration, tag, Release, or credential handling was performed or attempted. The publish step was exercised only against a **stub `npm` on `PATH` in an isolated scratch directory**; nothing contacted registry.npmjs.org.
- I could not execute the real workflow, so OIDC token exchange against npmjs.com is verified by construction and by npm's documented requirements (`id-token: write`, npm ≥ 11.5.1, Node ≥ 22.14.0 — all present), not by observation. First-run behaviour still depends on the operator completing the per-package registration the docs describe.
- The full suite was not re-run locally; exact-head CI success is the authority. My broader local slice (`Npm|Release|Workflow|Packaging`) was **397 passed, 1 failed**, and that one failure is an environment artifact, not a PR defect: `PackagedInvocationSmokeTests.CliProject_DeclaresDotNetToolPackagingMetadata` resolves the repo root with a fixed five-level `..` walk (`PackagedInvocationSmokeTests.cs:544`), which overshoots under an out-of-tree `--artifacts-path` layout and produces `/…/.intent-cli/worktrees/src/IntentSystem.Cli/IntentSystem.Cli.csproj`. The file is untouched by this PR and CI builds in-tree and is green. This is the recurring fixed-depth/marker root-resolution limitation previously reported on G698/G703/G704/G705.
- The `exit 0` at `release.yml:139` is the **NuGet** publish guard, identical to base and explicitly out of scope for #1538. The shape the issue condemns therefore still exists for NuGet; worth a future slice, not a change here.
- No label was applied, nothing was pushed, merged, promoted from draft, or closed out; no parent host packet file was read. Every mutation probe was reverted, the build artifacts directory was removed, and `git status` is empty at `1392561d7b13a19d559d7a6cb5ab83f72a7b7f76`.

PR #1541, opened 2026-08-17
G712: keep supervise lifetime session-scoped

### LINKED ISSUE #1540: G712: supervision must not survive a reboot — stop emitting a login-persistent LaunchAgent
# G712: supervision must not survive a reboot — stop emitting a login-persistent LaunchAgent

## Goal

Nothing intent-cli installs should start again after a reboot. The supervisor's lifetime should be bounded by the session that asked for it, and intent-cli should not appear in macOS Login Items.

## Why This Slice Exists Now

The operator found `intent-cli` listed in System Settings → General → Login Items & Extensions, annotated `Item from unidentified developer.`

That is an accurate description of what we install. `notify supervise install` writes a launchd property list into `~/Library/LaunchAgents/`, which macOS loads at every login, with `RunAtLoad = true`. After a reboot the supervisor is running again, started by nobody, on behalf of a session that ended.

The intent behind it is good — supervision exists so that a stalled thread is noticed while the operator is away. But from the operator's side, a process that starts without any session asking for it is indistinguishable from something that installed itself. Good intent does not change how it reads in System Settings, and a tool asking people to adopt it cannot also be a thing they find in their login items and do not remember approving.

## Current Observed State

Measured on macOS, 2026-08-16.

`~/Library/LaunchAgents/intent-cli.supervise.aic.aic-herdr.plist`, written 21:08:

```
KeepAlive        = true
RunAtLoad        = true
ThrottleInterval = 30
ProgramArguments = ~/.dotnet/tools/intent-cli notify supervise
                   --domain aic --team aic-herdr --repo J-Tech-Japan/aic
                   --owner-role orchestration --bound 900 --interval 300
                   --event-mode --write --format json
```

`launchctl list` showed three supervise jobs loaded:

```
intent-cli.supervise.aic.aic-herdr
intent-cli.supervise.sekiban-as-a-service.sekiban-as-a-service-orch
intent-cli.supervise.remote-herdr.remote-herdr
```

Only one plist remained on disk. Removing a plist does not unload an already-bootstrapped job, so what is loaded and what is on disk drift apart, and no intent-cli command reconciles them.

The current guidance describes this surface as emitting an operator-managed scheduler artifact *without registration or process execution*. Writing into the directory launchd auto-loads is registration in every sense that matters to the person who then finds it in Login Items.

- Supervision exists to notice a stall while nobody is watching. A previously recorded incident had a supervisor sleep 21h40m because of a misconfigured bound/interval. Whatever lifetime is chosen has to be weighed against that purpose rather than ignoring it.
- The distributed binary is unsigned and un-notarized, so the `unidentified developer` annotation is correct and is not itself the defect being fixed here.
- Bounding the supervisor to the provisioning session is the preferred outcome. If that is shown to defeat the purpose of supervision, the supervisor may outlive the provisioning session — but it must still be gone after a reboot. This fallback is explicitly permitted.
- `launchctl bootstrap` into the current GUI session, without an artifact in an auto-loaded directory, is one shape that satisfies both constraints. It is not mandated; pick what is defensible and say why.

## Target Repo / Path / Part

- Part: the lifetime of the process intent-cli asks the operator to run, and what that installation registers with the operating system

## In Scope

- The lifetime and registration mechanism of the supervise install path.
- A reconcile/uninstall path that unloads what is loaded and removes what is on disk.
- Install-time output that states what is created, where, how long it lives, and how to remove it.
- Correcting the guidance that currently claims no registration takes place.
- Handling existing installations on upgrade.

## Out Of Scope

- Signing or notarizing the binary.
- Changing supervision's detection logic, bound/interval semantics, or escalation behaviour.
- Any platform other than the one the defect was observed on, beyond keeping the chosen mechanism honest about what it does on each.

Self-contained. The observed plist contents, the loaded job list, the constraint set, and the permitted fallback are all stated above.

## Acceptance Criteria

- [ ] Nothing intent-cli installs, emits, or registers causes a process to start after a reboot — verified by rebooting, or by showing that no artifact is placed in any path launchd auto-loads at login and no login item is created.
- [ ] intent-cli no longer creates an entry in macOS Login Items & Extensions, and where an existing installation created one there is a documented one-command removal.
- [ ] The supervisor's default lifetime is bounded by the provisioning session; if the fallback is taken instead, the issue states which was chosen and why, and the guide tells the operator which one they are getting.
- [ ] A command reconciles loaded jobs with on-disk artifacts. Given the observed state — three loaded, one present — it leaves them in agreement and reports what it unloaded. An operator must not need to know `launchctl` to undo what intent-cli set up.
- [ ] Install-time output names its own persistence: what is created, where, how long it lives, how to remove it. An operator reading only that output can predict what they will find in System Settings.
- [ ] The guidance describing this surface is corrected in the same change. Per the guide-primacy rule, a feature whose guide route misdescribes it is not shipped.
- [ ] Re-running install after upgrade does not leave the old login-persistent artifact behind.

## Verification

- Show the artifact set produced by a fresh install, and show that none of it sits in a login-auto-loaded path.
- Show `launchctl list` before and after the reconcile command against a deliberately drifted state (a job loaded with its plist deleted).
- Show the install-time output text.
- Show the corrected guide route output.

## Related Links

- Reported from a System Settings screenshot showing `intent-cli` under Login Items & Extensions.

## Knowledge Maintenance

Node 08 gains the supervision lifetime contract: bounded by the provisioning session by default, never crossing a reboot.

## Base Branch Policy

When the fix removes persistence, do not replace it with something that merely hides the persistence — a helper that re-registers on next run is the same defect with more steps. If supervision genuinely cannot do its job without surviving the session, say so in the issue and take the documented fallback rather than working around the constraint.

tomohisa:
Closes #1540

## G712 implementation

`notify supervise install` now emits a session-scoped artifact under the routing repository's `.intent-cli/supervision/<domain>/<team>/install/` and rejects a custom macOS output path under `~/Library/LaunchAgents`. The generated macOS plist omits `RunAtLoad`; Windows and Linux generated artifacts also omit login/default-target auto-start triggers. Install authors the artifact and proves a bounded first cycle but does not execute lifecycle registration.

The permitted GUI-session fallback is chosen because supervision observes operator-owned orchestration health while the operator session is active. The printed `launchctl bootstrap gui/$(id -u) '<artifact-path>'` is an explicit current-session operator action; logout/reboot removes the GUI domain and requires an explicit new bootstrap. `notify supervise reconcile --write` and `uninstall --write` enumerate loaded-before/after state, boot out managed jobs, remove managed artifacts and legacy login-persistent plists, and preserve unrelated jobs.

## Guide reachability

From a new bare directory with no `.intent-cli/config.toml`, the Release-built CLI routes:

```text
intent-cli guide orchestrator-thread --format json
intent-cli guide orchestrator-thread --format markdown
```

Both outputs expose the G712 lifetime, install/bootstrap/reconcile/uninstall commands, artifact location, `RunAtLoad` absence, three-loaded/one-plist drift verification, and the lifecycle authority boundary.

## Isolated launchctl evidence

The retained fixture used label scope `intent-cli.supervise.g712.` and target `gui/501`:

- before: 3 loaded jobs (`g712.a`, `g712.b`, `g712.c`) and 1 plist on disk;
- write reconcile: unloaded all 3 and removed the 1 remaining plist;
- after: 0 scoped loaded jobs and 0 scoped plist artifacts;
- unrelated `com.example.unrelated` remained loaded;
- preserved stdout/stderr evidence logs remain under `.artifacts/g712-run-a4b4ecb7/bare/logs/`;
- durable local evidence record: `.artifacts/g712-run-a4b4ecb7/launchctl-drift-before-after.json`.

The repaired install path was also exercised against the original legacy `~/Library/LaunchAgents/intent-cli.supervise.aic.aic-herdr.plist`: the canonical install command removed that legacy plist, did not unload the pre-existing shared job, and failed closed at the one-second first-cycle proof because no lifecycle registration was executed. The new artifact and probe evidence remain in `.artifacts/g712-live-legacy-clean-a4b4ecb7/`.

## Verification

- Release CLI build: passed, 0 warnings / 0 errors.
- G712/G658/G704 focused tests: 22 passed.
- CI-compatibility/G712 guide matrix: 26 passed, 0 failed.
- affected guide/session/parity matrix: 371 passed, 0 failed.
- `git diff --check`: passed.
- full CLI test suite was attempted with a repo-local `TMPDIR` so no `/tmp` path was deleted; the current shared checkout produced 4,880 passed and 230 failures from unrelated existing environment/state-sensitive tests (including retained untracked-artifact safe-stash assumptions). No G712 or affected guide test failed after the declaration and snapshot updates.
- exact-head GitHub Actions run `31998974826` for `27b85ba953a2ca47b8c917c4386d9e151dbf74e1`: source-contract build/test passed and npm package dry-run passed without publishing.

No real service registration, package publish, signing, notarization, or shared launchctl job termination is performed by this PR.


tomohisa (commented):
## Review verdict: APPROVED

Reviewed at exact head `27b85ba953a2ca47b8c917c4386d9e151dbf74e1` against issue **#1540**. `Closes #1540`; base `main`; 19 files (+1175/−71); head unchanged before and after this review; worktree clean at the exact head.

Exact-head CI is green (run `31998974826`, `head_sha` = `27b85ba9`), but per the review context I did **not** treat that as acceptance evidence. Everything below was measured by building the PR CLI on macOS and executing it against a real, deliberately drifted `launchctl` fixture on this host.

Every acceptance criterion holds. Four non-blocking findings follow the evidence.

---

## Independent launchctl evidence

I built the PR CLI (Release) and created an isolated fixture in a **separate label namespace** (`intent-cli.supervise.g712rev.*`) so that this host's three live supervisors could serve as controls rather than casualties.

### Before — the exact drift shape from #1540: three loaded, one plist

```
$ launchctl list | awk '{print $NF}' | grep -E '^intent-cli\.supervise\.|^com\.g712rev\.|^com\.tomohisa\.codex\.'
com.g712rev.unrelated                                            ← control (unrelated prefix)
com.tomohisa.codex.intent-cli-design-reminder                    ← control (name contains "intent-cli")
com.tomohisa.codex.intent-system.issue-worker                    ← control
intent-cli.supervise.aic.aic-herdr                               ← control (managed prefix, other domain)
intent-cli.supervise.g712rev.alpha                               ← fixture
intent-cli.supervise.g712rev.bravo                               ← fixture
intent-cli.supervise.g712rev.charlie                             ← fixture
intent-cli.supervise.remote-herdr.remote-herdr                   ← control (managed prefix, other domain)
intent-cli.supervise.sekiban-as-a-service.sekiban-as-a-service-orch  ← control (managed prefix, other domain)

loaded fixture jobs : 3
managed artifacts   : 1   (…/g712rev/alpha/install/intent-cli.supervise.g712rev.alpha.plist)
legacy plists       : 1   (~/Library/LaunchAgents/intent-cli.supervise.g712rev.delta.plist, placed by me, never bootstrapped)
```

The three `intent-cli.supervise.*` jobs from **other domains** are the strongest control available: they share the managed prefix and differ only in scope. `com.tomohisa.codex.intent-cli-design-reminder` is a second useful control — its name contains `intent-cli` but not the managed label prefix.

### Dry run mutates nothing and reports honestly

```json
"command_mode": "dry-run", "scope": "intent-cli.supervise.g712rev.",
"loaded_before":  [alpha, bravo, charlie],
"unloaded":       [],
"would_unload":   [alpha, bravo, charlie],
"artifacts_before": ["~/Library/LaunchAgents/intent-cli.supervise.g712rev.delta.plist",
                     "…/g712rev/alpha/install/intent-cli.supervise.g712rev.alpha.plist"],
"removed_artifacts": [], "loaded_after": [alpha, bravo, charlie], "artifacts_after": [both]
```

`loaded_after` equals `loaded_before` and both artifacts survive — the preview genuinely previews.

### `reconcile --write` drives managed state to zero

```json
"command_mode": "write",
"unloaded":          [alpha, bravo, charlie],
"removed_artifacts": ["~/Library/LaunchAgents/intent-cli.supervise.g712rev.delta.plist",
                      "…/g712rev/alpha/install/intent-cli.supervise.g712rev.alpha.plist"],
"loaded_after": [], "artifacts_after": [], "errors": [],
"summary": "reconcile unloaded 3 supervise job(s) and removed 2 artifact(s).
            Remaining loaded jobs: 0; remaining artifacts: 0."
```

### After — remeasured independently, not read from the command's own output

```
$ diff before.txt after.txt
5,7d4
< intent-cli.supervise.g712rev.alpha
< intent-cli.supervise.g712rev.bravo
< intent-cli.supervise.g712rev.charlie
```

Three lines removed, nothing else. Independently:

```
managed artifacts under the artifact root : 0
~/Library/LaunchAgents/intent-cli.supervise.*.plist : 0
com.g712rev.unrelated                     : still loaded
com.tomohisa.codex.* (both)               : still loaded
intent-cli.supervise.{aic,remote-herdr,sekiban-as-a-service}.* : still loaded, same PIDs as before
```

The other-domain supervisors kept their **original PIDs** (25588 / 62984 / 57790), so they were not merely re-listed — they were never stopped or restarted. Managed jobs and artifacts reach zero; unrelated jobs are untouched.

I then booted out my own `com.g712rev.unrelated` fixture. `~/Library/LaunchAgents` is byte-for-byte the same 11-file listing it had before this review, and no `g712rev` label remains anywhere in launchd.

## AC-by-AC

### Nothing starts after a reboot ✅

The AC permits proof by showing no artifact reaches a login-auto-loaded path and no login item is created. Measured on all three platforms:

**macOS** — the generated plist, in full:

```
Label, ProgramArguments, KeepAlive=true, ThrottleInterval=30,
WorkingDirectory, StandardOutPath, StandardErrorPath, EnvironmentVariables/PATH

RunAtLoad               absent
StartCalendarInterval   absent
StartInterval           absent
WatchPaths              absent
QueueDirectories        absent
```

`KeepAlive` without `RunAtLoad`, in a directory launchd does not scan, restarts the job **only while the operator has explicitly bootstrapped it into the live GUI domain** — which is the supervision purpose — and cannot start it at login.

**Linux** — the `.service` unit has **no `[Install]` section and no `WantedBy=`**, so `systemctl enable` has nothing to hook; only an explicit `start` runs it.

**Windows** — the task XML has **no `<Triggers>` element at all** (no `LogonTrigger`, `BootTrigger`, or `CalendarTrigger`); `<Enabled>true</Enabled>` is a Settings key, not a trigger.

The path guard is enforced, not just documented — install refuses a custom output under the auto-loaded directory and writes nothing:

```
$ … install --output ~/Library/LaunchAgents/intent-cli.supervise.g712rev.alpha.plist --write
"error": "login-auto-loaded-path: refusing to emit a macOS supervisor artifact under
          '/Users/…/Library/LaunchAgents'; use the session-scoped default under
          .intent-cli/supervision and bootstrap the current GUI session explicitly."
$ ls ~/Library/LaunchAgents | grep intent-cli.supervise   → nothing
```

A nested path (`~/Library/LaunchAgents/sub/x.plist`) is refused identically, so the check is on the directory tree rather than an exact string.

A default `--write` install landed its artifact at `<routing-repo>/.intent-cli/supervision/g712rev/alpha/install/intent-cli.supervise.g712rev.alpha.plist` and nowhere else — the whole produced file set was one file.

*Boundary:* I could not read the Background Task Management database directly — `sfltool dumpbtm` requires admin authorization, which I did not request. I verified the alternative the AC explicitly allows instead.

### One-command removal of an existing installation ✅

Shown above: a legacy `~/Library/LaunchAgents/intent-cli.supervise.*.plist` is enumerated by the same scoped command and removed by `--write`, alongside the session artifact. The operator needs no `launchctl` knowledge.

### Lifetime chosen, and the guide says which one you get ✅

The permitted fallback is taken and declared. From the guide route itself:

> G712 chooses the permitted GUI-session lifetime: supervision may be explicitly bootstrapped into the current GUI domain, but it is never login-auto-loaded and does not survive logout or reboot.

The rationale — supervision observes operator-owned orchestration health while the operator session is active — is defensible against #1540's own baseline: the incident the baseline protects against is a supervisor sleeping while the operator is *away from the keyboard*, which the GUI-session scope still covers; it stops only at logout/reboot, when no orchestration session exists to supervise. This is the documented fallback, not a workaround that hides persistence — there is no re-registering helper anywhere in the diff.

### A command reconciles drift ✅

Measured above against exactly the three-loaded/one-present shape from the issue.

### Install-time output names its own persistence ✅

```
- lifetime: current GUI session only; no LaunchAgents login auto-load and no reboot persistence
- artifact path: `…/.intent-cli/supervision/g712rev/alpha/install/intent-cli.supervise.g712rev.alpha.plist`
- supervise runtime directory / stdout log path / stderr log path: …
- registration command (operator action): `launchctl bootstrap gui/$(id -u) '<artifact-path>'`
- unregistration command (operator action): `launchctl bootout gui/$(id -u)/intent-cli.supervise.g712rev.alpha`
- reconcile/removal command (operator action): `intent-cli notify supervise reconcile --write`
- legacy login-persistent artifacts removed: none
- install lifecycle command executed by intent-cli: false
```

An operator reading only this can predict what System Settings will show: nothing, because nothing is registered until they run the bootstrap line themselves.

### The guidance is corrected, without transferring lifecycle authority ✅

Both guide routes run from a **bare metadata-free directory** and create nothing:

```
$ intent-cli guide orchestrator-thread --format json      → exit 0, 221,163 B
$ intent-cli guide orchestrator-thread --format markdown  → exit 0, 186,196 B
$ ls -a <bare-dir>   → . ..   (unchanged)
```

The old claim is gone — `"without registration or process execution"`, `"without registration"`, `"no registration takes place"` all return **0 hits** in both formats. What replaces it keeps the boundary explicit:

> **Lifecycle authority boundary:** Install authors and first-cycle-probes only. Registration is an explicit operator action; reconcile/uninstall is the bounded current-session unload/removal command and does not grant workflow or recovery authority.

That authority claim is true in practice, not just in prose: after a `--write` install, `launchctl bootstrap` of the produced plist **succeeded**, which it could not have done had install already registered the job.

### Re-running install after upgrade removes the old artifact ✅

```
before: ~/Library/LaunchAgents/intent-cli.supervise.g712rev.upgrade.plist   (RunAtLoad, KeepAlive)
$ … install --domain g712rev --team upgrade --write
after:  ls ~/Library/LaunchAgents | grep intent-cli.supervise   → nothing
```

And `--dry-run` with the same legacy file present removes nothing and reports `legacy login-persistent artifacts removed: none` — the preview is honest in this direction too.

### Docs and tests

EN/JA parity across the new G712 substance in `12-agent-message-orchestration.md` and `08-command-reference.md`: **12/12 properties present in both, 0 mismatches** (G712 named, GUI-session lifetime, no `~/Library/LaunchAgents`, `RunAtLoad` omitted, bootstrap command, reconcile/uninstall commands, legacy plists, artifact location). The 1.0 compatibility ledger row is updated to `G658/G712` in both languages. `git diff --check` clean.

Affected suites at the exact head: `NotifySuperviseReconcileG712Tests` + `NotifySuperviseInstallG658Tests` + `HerdrPaneTopologyG586Tests` + `HerdrWakeSourcesG581Tests` → **24 passed, 0 failed**.

### Preserved implementation evidence — corroborated, then re-derived

`…/g712-run-a4b4ecb7/launchctl-drift-before-after.json` records the same shape (three `intent-cli.supervise.g712.*` loaded, one artifact, `com.example.unrelated` surviving, both counts to zero) and the same `legacy plist removed by repaired install path; first-cycle-proof-failed` observation. I did not rely on it: my fixture used a different namespace, a different artifact root, and *live third-party supervisors* as controls rather than a synthetic unrelated job.

## Non-blocking findings

### 1. A `--write` install can delete a file from `~/Library/LaunchAgents` and not say so

Legacy removal happens at `NotifySuperviseInstallCommand.cs:143`, **before** the first-cycle proof at `:194`. `LegacyArtifactsRemoved` is only carried on the success result (`:245`, rendered at `:749`); the failure paths go through `EmitFailure`, whose payload is just `{operation, success, error, manages_process}`. Measured:

```
before: ~/Library/LaunchAgents/intent-cli.supervise.g712rev.upgrade.plist  present
$ … install --team upgrade --startup-bound 5 --write --format json
  output keys = ['error', 'manages_process', 'operation', 'success']
  error       = first-cycle-proof-failed: …
after:  the plist is gone — and the output never named it
```

The deletion is the point of the unit, so this is not a wrong action; it is an undisclosed one on the path where the operator is already being told something went wrong. AC 5 asks install-time output to name its own persistence, and here it names the failure but not the file it removed. Carrying `legacy_artifacts_removed` onto the failure payload would close it.

### 2. The one-command undo that install prints is global across every domain

Install prints, and the guide repeats, the unscoped form:

```
- reconcile/removal command (operator action): `intent-cli notify supervise reconcile --write`
```

On this host that command is not scoped to the install that printed it. Measured with `--dry-run` only (nothing was written):

```
$ notify supervise uninstall --dry-run
scope         = null
loaded_before = [intent-cli.supervise.aic.aic-herdr,
                 intent-cli.supervise.remote-herdr.remote-herdr,
                 intent-cli.supervise.sekiban-as-a-service.sekiban-as-a-service-orch]
```

Three live supervisors across three unrelated domains. The behaviour is disclosed in prose — `docs/en/12…:822` says "boot out every managed `intent-cli.supervise.*` job" — and `--domain/--team` and `--label-prefix` exist and work correctly (my entire fixture run depended on them). But those flags appear only in `--help`; neither the guide nor the docs mention them, so the operator's discoverable path is the widest one. Since install knows its own domain and team, printing `reconcile --domain <d> --team <t> --write` would make the undo match the install that emitted it, with the global form available for a deliberate full cleanup.

### 3. `reconcile` and `uninstall` are the same operation

`Execute(...)` takes `operation` only to label output: it appears in `Operation = "supervise-" + operation`, in `EmitFailure`, and in `BuildSummary`, and nowhere in the control flow. Both verbs unload every in-scope job and delete every in-scope artifact. Nothing is wrong with the behaviour — the review context expects all managed state to reach zero — but "reconcile" conventionally means *make loaded agree with on-disk*, which would keep a job whose artifact is still present. An operator reaching for the gentler-sounding verb gets the same removal. Worth either a one-line note that the two are aliases, or making `reconcile` genuinely converge toward disk.

### 4. `--routing-root` is accepted, validated, then discarded by reconcile/uninstall

```csharp
_ = Path.GetFullPath(options.RoutingRoot ?? context.RepoRoot);   // :63 — result thrown away
artifactRoot = context.ResolveSupervisionArtifactRootPath();
```

It is advertised in the usage string and rejects a malformed value with `invalid-routing-root`, but a well-formed value changes nothing. This is **not** a wrong-target hazard: install resolves its artifact location through the same `context.ResolveSupervisionArtifactRootPath()`, so install and reconcile always agree on where artifacts live. It is a flag that looks like it scopes the operation and does not. Either honour it or drop it from the usage line.

## Not a finding, but worth recording

The G712 fixture labels in `NotifySuperviseReconcileG712Tests` are this operator's **real** job names, copied from #1540's observed state (`aic.aic-herdr`, `sekiban-as-a-service.sekiban-as-a-service-orch`, `remote-herdr.remote-herdr`). Isolation is correct — a fake `INotifyProcessRunner` is installed before every call, `UserProfileDirectoryFactory` is redirected to a repo-local `home`, `MacOsDetector` is forced, and `Dispose` restores all three — and I confirmed every test class that mutates `NotifyCommand.ProcessRunnerFactory` sits in the same `WorkerNextActionSharedState` collection, so xUnit serialises them. The redirect demonstrably worked: the tests created `.artifacts/g712-reconcile-*/home/Library/LaunchAgents`, not the real one. The only remark is defence-in-depth: because the fixture labels match reality on this machine, a future test that forgot the fake would boot out real supervisors *and still pass its assertions*. Synthetic labels would make that failure mode loud.

- **The only `--write` reconcile I executed was scoped**, and its own record proves the blast radius: `scope = intent-cli.supervise.g712rev.`, `unloaded = [g712rev.alpha, g712rev.bravo, g712rev.charlie]`. Every other reconcile/uninstall invocation was `--dry-run`. I did not run the unscoped form with `--write` at any point.
- I did not reboot this machine, so reboot-survival is proven by construction (no artifact in any auto-loaded path; no `RunAtLoad`, `LogonTrigger`, `BootTrigger`, `CalendarTrigger`, `[Install]`/`WantedBy`), which is the alternative #1540 explicitly permits — not by observing a boot.
- I did not read Login Items & Extensions / the BTM database; `sfltool dumpbtm` requires admin authorization and I did not request it.
- Windows and Linux artifacts were inspected as **files authored from macOS**; I did not load a systemd unit or register a Task Scheduler task. They remain `emitted-but-unverified`, exactly as the artifacts themselves state.
- Every fixture I created was removed: three bootstrapped supervise jobs and one unrelated job booted out, two legacy plists deleted, build output and test `.artifacts/` removed. `~/Library/LaunchAgents` matches its pre-review listing exactly and `git status` is empty at `27b85ba953a2ca47b8c917c4386d9e151dbf74e1`.
- **One observation I cannot attribute, recorded for orchestration:** `intent-cli.supervise.aic.aic-herdr` (PID 25588) was loaded throughout the fixture run and verified still loaded immediately after my scoped `--write` reconcile; it is no longer loaded at the end of this review. The other two live supervisors remain. Nothing I ran targeted it — my only write was scoped to `g712rev`, the tests use a fake runner and a redirected home, and dry runs mutate nothing — so I am reporting the change rather than explaining it. Orchestration should confirm whether it re-provisioned that supervisor itself.
- No label was applied, nothing was pushed, merged, promoted from draft, or closed out; no parent host packet file was read.

PR #1542, opened 2026-08-17
G712: restore supervision-setup guide route

### LINKED ISSUE #1540: G712: supervision must not survive a reboot — stop emitting a login-persistent LaunchAgent

tomohisa:
Closes #1540

## G712 guide-reachability repair

The post-merge G712 contract declared `intent-cli guide workflow task supervision-setup`, but the merged CLI did not expose that task. This PR restores the declared route from the exact `origin/main` base containing `037c4acc`.

### Production route

- Adds `GuideWorkflowTaskSupervisionSetupCommand` with structured JSON and Markdown output, contract version `g712-session-scoped-supervision/v1`.
- Registers `supervision-setup` in the workflow dispatcher, supported-task failures, workflow help, command catalog, top-level guide pointers, and the EN/JA command references.
- Reuses the shipped G712 session-scoped supervision contract: install, explicit current-GUI-session bootstrap, reconcile, and uninstall.
- The route is read-only and metadata-free. It does not read `.intent-cli/config.toml`, queue-state, packets, or intents, and it executes none of the emitted lifecycle commands.
- The rendered contract preserves the GUI-session fallback, no `RunAtLoad`/`~/Library/LaunchAgents` persistence, bounded first-cycle proof, managed-only reconcile/uninstall, and no workflow/recovery authority.

### Reachability evidence

From a fresh bare directory containing only `.` and no `.intent-cli` configuration:

```text
dotnet .../IntentSystem.Cli.dll guide workflow task supervision-setup --format json     # exit 0; metadata_free=true; read_only=true
dotnet .../IntentSystem.Cli.dll guide workflow task supervision-setup --format markdown  # exit 0; session-scoped commands and negative checks rendered
find . -maxdepth 3 -print                                                               # .
```

The JSON and Markdown outputs both name the install, explicit `launchctl bootstrap gui/$(id -u) '<artifact-path>'`, reconcile, and uninstall commands plus the no-registration/no-login-persistence boundaries.

### Verification

- Focused route regressions: 4 passed.
- External-user route audit plus new route regressions: 10 passed.
- Affected guide/G712 matrix: 349 passed.
- Release full CLI suite: 5,115 passed, 1 expected skip, 0 failed.
- CI-equivalent Release solution build: passed with 0 warnings and 0 errors.
- CI-equivalent Release solution tests: all 11 assemblies passed (5,445 passed, 1 expected skip, 0 failed).
- `git diff --check`: passed.

No G712 supervision detection or lifecycle implementation was changed; this PR only restores the shipped guide route, its discoverability registrations, documentation parity, and negative reachability coverage.

### DIFF
--- a/docs/en/08-command-reference.md
+++ b/docs/en/08-command-reference.md
@@ -484,6 +484,23 @@ Operator-dogfooding prompt templates that wire these loops entirely through the
 deterministic worker/metadata commands live under
 [`docs/automation-templates/`](../automation-templates/README.md).
 
+## Session-scoped supervision setup (G712)
+
+The declared supervision setup route is executable from a bare directory with
+no `.intent-cli/config.toml` or host metadata:
+
+```bash
+intent-cli guide workflow task supervision-setup --format json
+intent-cli guide workflow task supervision-setup --format markdown
+```
+
+It renders the shipped session-scoped contract: `notify supervise install`
+authors and proves the artifact without registering a process, the printed
+`launchctl bootstrap gui/$(id -u) '<artifact-path>'` is an explicit current-GUI
+session action, and `notify supervise reconcile --write` / `uninstall --write`
+reports before/after state and removes only managed drift. The route is
+read-only and does not execute any of those lifecycle commands.
+
 ## Recovery
 
 ```bash
--- a/docs/ja/08-command-reference.md
+++ b/docs/ja/08-command-reference.md
@@ -447,6 +447,23 @@ intent-cli guide oneshot --kind host-review-next-slice    --domain <name>
 worker/metadata コマンドだけでループを回す operator dogfooding 向けプロンプトテンプレートは
 [`docs/automation-templates/`](../automation-templates/README.md) にあります。
 
+## セッションスコープの supervision セットアップ（G712）
+
+宣言された supervision setup route は、`.intent-cli/config.toml` や host metadata
+のない bare directory から実行できます。
+
+```bash
+intent-cli guide workflow task supervision-setup --format json
+intent-cli guide workflow task supervision-setup --format markdown
+```
+
+この route は shipped の session-scoped contract を表示します。`notify supervise install`
+は artifact を作成して first-cycle proof を確認するだけで process を登録せず、表示された
+`launchctl bootstrap gui/$(id -u) '<artifact-path>'` は現在の GUI session で operator が明示的に
+実行する action です。`notify supervise reconcile --write` / `uninstall --write` は before/after
+を報告し、managed drift だけを削除します。route 自体は read-only で、これらの lifecycle command
+を実行しません。
+
 ## 復旧
 
 ```bash
--- a/src/IntentSystem.Cli/Commands/CommandRouter.cs
+++ b/src/IntentSystem.Cli/Commands/CommandRouter.cs
@@ -793,8 +793,8 @@ internal static class CommandRouter
     /// G334: canonical one-line pointers for the six workflow phases the
     /// issue names (init / interview / packet / issue / automation /
     /// bug repair), the G338 loop-setup pair (implementation-loop /
-    /// review-next-slice-loop), and the G339 bug-to-intent-repair
-    /// chain. Tests assert each phase appears in the top-level help
+    /// review-next-slice-loop), the G339 bug-to-intent-repair
+    /// chain, and the G712 supervision-setup route. Tests assert each phase appears in the top-level help
     /// output so external agents can find the workflow entry without
     /// reading local rules.
     /// </summary>
@@ -809,6 +809,7 @@ internal static class CommandRouter
         "implementation-loop — `intent-cli guide workflow task implementation-loop --target-repo <r> --agent claude --frequency 5m --format markdown` (paste-ready child implementation-loop prompt with current label/claim/complete rules, G338).",
         "review-next-slice-loop — `intent-cli guide workflow task review-next-slice-loop --domain <d> --target-repo <r> --agent claude --frequency 20m --format markdown` (paste-ready host review / next-slice-loop prompt with current host-sync preflight + packet/issue lifecycle rules, G338).",
         "bug-to-intent-repair — `intent-cli guide workflow task bug-to-intent-repair --format json` (report → triage → plan → intent-repair → implementation-repair chain; classifies implementation-mismatch / intent-gap / packet-gap / rule-gap / metadata-workflow-gap; recommends packet creation when the bug is in intent-cli rules/guidance, G339).",
+        "supervision-setup — `intent-cli guide workflow task supervision-setup --format json` (metadata-free G712 session-scoped install/current-GUI registration/reconcile/uninstall contract; renders guidance only).",
         "improve (design-thread reflection) — `intent-cli improve --domain <d> --format markdown` returns the realignment guide. Preview `improve window ... --write` declares the independent recency bound; `improve record ... --write` appends evidence after human/agent review and never grades quality, G456/G457/G662. NOT a scheduler, auto-run, stalled-work class, or operational recovery.",
         "grill (persistent interview mode) — `intent-cli grill --domain <d> --format markdown` (alias of `intent-cli guide grill`): once the user asks to grill a topic, stay in grill mode — generate an open-question backlog from current intents/packets/ADRs/docs, ask one question at a time, and continue after each answer until a stop condition holds, G463. Built on the interview artifacts; NOT clarification (blocker resolution) and NOT improve (retrospective realignment).",
         "stack (packet backlog creation) — `intent-cli stack --domain <d> --target-repo <r> --format markdown` (alias of `intent-cli guide stack`): forward planning — create an ordered packet backlog from the current intents (often ~10), commit/push durable state, and publish AT MOST the first GitHub issue by default, G464. Distinct from improve (retrospective realignment), grill (open-question interview), clarification (blocker resolution), and runtime queue transitions.",
--- a/src/IntentSystem.Cli/Commands/GuideCommandsListCommand.cs
+++ b/src/IntentSystem.Cli/Commands/GuideCommandsListCommand.cs
@@ -358,6 +358,15 @@ internal static class GuideCommandsListCommand
             RecommendedCaller = CallerChatAgent,
             Purpose = "Canonical host review / next-slice-loop prompt generator: `intent-cli guide workflow task review-next-slice-loop --domain <d> --target-repo <r> --agent claude --frequency 20m --format markdown` emits the paste-ready host loop prompt with current preflight + packet/issue lifecycle rules (G338)."
         },
+        new CommandGroupEntry
+        {
+            Name = "guide workflow task supervision-setup",
+            Role = RoleHostReview,
+            Classification = ClassificationPrimary,
+            Mutability = MutabilityReadOnly,
+            RecommendedCaller = CallerChatAgent,
+            Purpose = "G712 metadata-free supervision setup route: `intent-cli guide workflow task supervision-setup --format json|markdown` renders the shipped session-scoped install, current-GUI registration, reconcile, and uninstall contract without reading host metadata or executing lifecycle commands."
+        },
     };
 
     public static int Execute(CliContext context, string[] args, TextWriter writer)
--- a/src/IntentSystem.Cli/Commands/GuideHelpCommand.cs
+++ b/src/IntentSystem.Cli/Commands/GuideHelpCommand.cs
@@ -42,7 +42,8 @@ internal static class GuideHelpCommand
     /// <c>interview</c>, <c>packet</c>, <c>issue</c>,
     /// <c>automation</c>, <c>bug-repair</c>, <c>implementation-loop</c>
     /// (G338), <c>review-next-slice-loop</c> (G338),
-    /// <c>bug-to-intent-repair</c> (G339).
+    /// <c>bug-to-intent-repair</c> (G339), and <c>supervision-setup</c>
+    /// (G712).
     /// </summary>
     internal static readonly IReadOnlyList<WorkflowGuidePointer> WorkflowGuides = new[]
     {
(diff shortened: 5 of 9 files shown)


tomohisa (commented):
## Review verdict: APPROVED

Reviewed at exact head `475f40ee00dc041cd4f152691e1eb0a29ce98138` against issue **#1540** and the post-merge guide-reachability contract. `Closes #1540`; base `main`; 9 files (+434/−8); one commit on top of `037c4acc` (the #1541 merge); head unchanged before and after this review; worktree clean at the exact head.

Exact-head CI green (run `32002887554`, `head_sha` = `475f40ee`): `Build and test (source contract)` 2m3s, `npm package dry-run (no publish)` 1m29s.

The blocking acceptance — the packet-declared direct route, executed from a fresh bare metadata-free directory — passes in both formats, and the contract it renders is accurate against the shipped code because it **references the same constants** rather than restating them. One non-blocking guard gap follows the evidence.

---

## The repair premise is real

At the merged G712 state, `supervision-setup` existed only as a `guide next` **recommended action id** (`GuideNextCommand.cs:45` and its decision-set prose) — never as an executable `guide workflow task`. So the packet declared a route the merged CLI could not run. This PR adds the route on top of the merge:

```
$ git log --oneline -3
475f40ee G712: restore supervision setup guide route
037c4acc Merge pull request #1541 …
27b85ba9 Fix G712 CI guide compatibility

$ git merge-base --is-ancestor 037c4acc HEAD   → yes
```

## Blocking acceptance — the direct route from a bare directory

I built the PR CLI in Release and ran it with the working directory set to a freshly created empty directory with no `.intent-cli` anywhere in it or above it.

```
$ find . -maxdepth 3 -print
.

$ dotnet …/IntentSystem.Cli.dll guide workflow task supervision-setup --format json
exit 0      4,978 bytes

$ dotnet …/IntentSystem.Cli.dll guide workflow task supervision-setup --format markdown
exit 0      4,508 bytes

$ find . -maxdepth 3 -print
.
```

Exit 0 in both formats and **the directory is still empty** — nothing created, no `.intent-cli`, no cache, no state. The JSON declares its own posture and the render agrees with it:

```
task              = supervision-setup
contract_version  = g712-session-scoped-supervision/v1
metadata_free     = True
read_only         = True
contract_statements[0..3], commands[0..3], artifact_location, authority_boundary, negative_checks[0..3]
```

## The routing is accurate — and structurally cannot drift

This is the part I most wanted to falsify, since #1540 was filed precisely because guidance misdescribed the surface it pointed at. The route does not restate the contract in its own words:

```csharp
internal static readonly IReadOnlyList<string> ContractStatements = new[]
{
    SupervisionGuideText.InstallBoundRule,
    SupervisionGuideText.InstallArtifactRule,
    SupervisionGuideText.InstallEvidenceRule,
    SupervisionGuideText.SessionLifetimeRule,
};
```

Those are the same constants the shipped install output and the orchestrator-thread guide render. A change to the supervision contract changes this route in the same edit — the class of drift that produced #1540 is closed by construction rather than by a matching assertion.

I still verified the claims by **executing what the route names**, not by trusting the shared source:

| route claim | measured |
|---|---|
| `--bound` < `--interval` is fail-closed as `bound-below-interval` | `"error": "bound-below-interval: declared bound 100s is smaller than interval 300s; a healthy supervisor is structurally judged absent (supervisor-not-running)…"` |
| `--startup-bound` default is 30s | failure text reads `within 30s` |
| artifact lives under `.intent-cli/supervision/<domain>/<team>/install/` | the only file produced was `…/supervision/g712b/t1/install/intent-cli.supervise.g712b.t1.plist` |
| never under `~/Library/LaunchAgents` | `ls ~/Library/LaunchAgents \| grep -c '^intent-cli.supervise'` → **0** |
| generated plist omits `RunAtLoad` | `grep -c RunAtLoad <plist>` → **0** |

The rendered contract also now carries the *reason* for the fallback in a shipped artifact:

> This GUI-session fallback is chosen because supervision observes operator-owned orchestration health while the operator session is active; a reboot or logout must require an explicit new bootstrap.

That closes the gap I recorded on #1541, where the rationale lived only in the PR description while the guide stated the choice without justifying it.

## Registration on every declared surface — executed, not grepped

```
$ guide workflow task <bad-name>
Unknown 'guide workflow task' name 'no-such-task'. Supported: init-host, intent-interview,
packet-draft, issue-publish, implementation-loop, review-next-slice-loop,
bug-to-intent-repair, supervision-setup.          ← dispatcher + supported-task failure

$ guide help --format json | grep -c 'guide workflow task supervision-setup'      → 2

$ guide commands list --format json
catalog entries: 1
  name    = guide workflow task supervision-setup
  purpose = G712 metadata-free supervision setup route: … renders the shipped session-scoped
            install, current-GUI registration, reconcile, and uninstall contract without
            reading host metadata or executing lifecycle commands.
```

Every one of those ran from the bare directory, which stayed empty. EN and JA command references both carry the two invocations (`docs/en/08-command-reference.md:493-494`, `docs/ja/08-command-reference.md:456-457`) with matching surrounding prose, and `ExternalUserHappyPathAuditTests` gains the route as a `ChildCwdSafe` phase so the external-user audit now covers it.

## Negative checks — six mutations, one at a time, each reverted

Baseline at the exact head: `GuideWorkflowTaskSupervisionSetupG712Tests` + `ExternalUserHappyPathAuditTests` + `NotifySuperviseReconcileG712Tests` + `NotifySuperviseInstallG658Tests` → **24 passed, 0 failed**.

| mutation | result |
|---|---|
| delete the dispatcher arm so the route is unreachable | **6 FAIL** — both route tests plus `EveryRequiredWorkflowGuide_ExistsAndReturnsNonEmptyContent`, `ChildImplementationGuides_WorkWithoutDotIntentCliInCwd`, `EveryWorkflowGuide_NamesAtLeastOneIntentCliCommand`, `NoWorkflowGuide_InstructsAgentToReadLocalSkillFilesOrCopiedPrompts` |
| rename the catalog entry | 1 FAIL — `CatalogAndTopLevelHelp_AdvertiseTheSameReachableRoute` |
| corrupt the registration command to the login-persistent `launchctl load -w ~/Library/LaunchAgents` | 2 FAIL — JSON and Markdown route tests |
| delete the "must not be emitted under `~/Library/LaunchAgents`" claim | 1 FAIL — Markdown route test |
| invert the authority boundary to "grants full workflow and recovery authority" | 1 FAIL — JSON route test |
| rename every `RunAtLoad` mention in the route to `NoSuchKey` | **0 FAIL — not detected** (see finding) |

Removing the route is loud in six places, including the external-user audit, so a future merge cannot drop it silently the way the merged state did. The three most load-bearing claims — the bootstrap command, the LaunchAgents prohibition, and the authority boundary — each have a real assertion behind them.

## Non-blocking finding

### The route's `RunAtLoad` boundary is the one unguarded claim

`RunAtLoad` appears exactly once in the route source — `negative_checks[1]`, "The generated macOS artifact must omit RunAtLoad and must not be emitted under `~/Library/LaunchAgents`". The second half of that sentence is asserted; the first half is not:

```
$ perl -pi -e 's/RunAtLoad/NoSuchKey/g' GuideWorkflowTaskSupervisionSetupCommand.cs
$ grep -c RunAtLoad …  → 0
    → Passed! 10 passed, 0 failed
```

`RunAtLoad` is the single launchd key #1540 was filed about, so it is the claim most worth pinning in the route that teaches operators about it. The impact is limited, and deliberately so: the **behaviour** is guarded where it counts — `Assert.DoesNotContain("RunAtLoad", artifact)` appears in both `NotifySuperviseReconcileG712Tests.cs:158` and `NotifySuperviseInstallG658Tests.cs:98`, and I independently confirmed the generated plist omits the key using this build. So the artifact cannot regress; only the route's description of it could. Adding `RunAtLoad` to the existing Markdown/JSON assertions costs one line.

### Carried forward from #1541, unchanged and still non-blocking

The route renders `notify supervise reconcile --write --format json` and `uninstall --write --format json` in their unscoped form, which on a multi-domain host boots out every managed `intent-cli.supervise.*` job across all domains. The route is honest about it — "boots out **every managed** `intent-cli.supervise.*` job" — and I measured that breadth on #1541 (`--dry-run` enumerating three live supervisors from three unrelated domains). The `--domain/--team` and `--label-prefix` scoping flags still appear only in `--help`. Nothing in this PR makes it worse; it is simply the natural place to surface the scoped form once someone fixes it.

## No lifecycle or detection change

```
$ git diff --stat 037c4acc HEAD -- NotifySuperviseInstallCommand.cs NotifySuperviseReconcileCommand.cs \
                                    NotifyMeasuredSupervisor.cs SupervisionGuideText.cs
(empty)
```

Byte-identical to merged main. The nine changed files are the new route command, its registration in the dispatcher / workflow help / top-level help / command catalog, the EN and JA command references, and two test files. `CommandRouter.cs` gains one pointer line and a doc-comment clause. `git diff --check` clean. The #1541 lifecycle and launchctl evidence therefore stands unchanged and is not re-litigated here.

- I did **not** re-run the #1541 drift fixture. That evidence — three loaded jobs and one plist driven to zero while three live other-domain supervisors kept their original PIDs — remains the approved record at #1541 (`#pullrequestreview-4948865658`), and this PR changes no code it covered.
- No launchd job was created, bootstrapped, or booted out during this review. The two `install --write` invocations wrote only into an isolated scratch routing root; `~/Library/LaunchAgents` still contains zero `intent-cli.supervise.*` plists and no `g712*` label exists in launchd.
- The full suite was not re-run locally; exact-head CI success at `475f40ee` is the authority. My focused run was 24 passed / 0 failed.
- The route was exercised on macOS only; its rendered contract names macOS-specific commands, which is consistent with the shipped platform boundary rather than a gap introduced here.
- Every mutation probe was reverted immediately, the build output was removed, nothing was pushed, merged, label-mutated, promoted from draft, or closed out, and no parent host packet file was read. `git status` is empty at `475f40ee00dc041cd4f152691e1eb0a29ce98138`.

PR #1548, opened 2026-08-18
Fix herdr-only host-state routing guidance

tomohisa:
Closes #1547

## What changed

- Corrected the rendered herdr-only Codex recipe: the role-work-root is an ordinary-file root, `.git` writes remain denied even when `.git` is inside that root, and adding another `--add-dir` does not grant repository-metadata access.
- Added an explicit Codex host-state duty route for issue publication, host-state pushes, and approved-PR closeout. Codex routes the bounded request and verifies the returned canonical state; it does not widen the sandbox or ask design to perform routine transitions.
- Made managed-worktree and ordinary `/private/tmp/<role>-<unit>` fallback instructions performable, including the required git-ignored managed root and host-owned cleanup. Mirrored the rendered contract in English/Japanese docs and expanded rendering tests.

## Rendered herdr-only evidence

These excerpts are from the release-built CLI rendering with `--agent codex --orchestrator-agent codex --format markdown` and a team-scoped `herdr-only` session record.

### Codex recipe

```text
#### Codex (measured recipe, G647)

herdr agent start <logical-role> --kind codex --pane <pane-id> -- --sandbox workspace-write --ask-for-approval never --add-dir <role-work-root> [--add-dir <host-routing-root>]

- **role-derived roots** — Use one bounded --add-dir <role-work-root> for the role checkout/worktree; add the host routing root only for a role whose canonical report surface needs it. The role-work-root is the seat's ordinary-file root, not permission to write repository metadata: Codex cannot write `.git` even when `.git` is inside a declared root. A non-sandboxed host-state role must therefore prepare registered worktrees and perform host-state git operations before delegation.
- **denial semantics** — Measured envelope asymmetry: writes outside declared roots are denied, while reads outside declared roots are not denied. Writes to `.git` are also denied even when `.git` is inside a declared root (for example `.git/index`, `.git/FETCH_HEAD`, and worktree metadata); adding another `--add-dir` does not make those operations available. Treat both asymmetries as explicit security facts, not as permission guarantees, and route repository-metadata work to the non-sandboxed host-state role.
```

### Orchestrator duty list

```text
## Role boundary (design authors; orchestrator coordinates)

### Host-state duty routing

[...]

PR #1553, opened 2026-08-19
G716: correct the .git envelope rule and state the routing decision

### LINKED ISSUE #1552: G716: correct the shipped `.git` envelope rule — declaring `<repo>/.git` does grant the write
# G716: correct the shipped `.git` envelope rule — declaring `<repo>/.git` does grant the write

Resolves the consumer report in #1550. Corrects guidance this repo shipped one day earlier in #1548 (G714).

## Goal

The rendered envelope text states what was measured, and the duty routing is chosen against that rather than against a claim that does not hold.

## Why This Slice Exists Now

G714 shipped a sentence with two halves. Re-measured independently on the design host, they fare differently — and the false half is the one that closed off the cheaper resolution.

Shipped text:

> Writes to `.git` are also denied even when `.git` is inside a declared root … another `--add-dir` does not make those operations available

| half | verdict |
| --- | --- |
| `.git` writes denied even inside a declared root | **holds** |
| another `--add-dir` cannot make them available | **false** |

## Current Observed State

Re-measured 2026-08-19, `codex-cli 0.147.0`, macOS. Fresh `git init --bare` + clone outside `/tmp`. `FETCH_HEAD` deleted and re-checked around every fetch case.

| # | cwd | declared roots | command | result |
|---|---|---|---|---|
| B | `work` | `work` | `touch .git/PROBE` | denied |
| B′ | `work` | `work` | `git fetch origin` | **denied** (`cannot open '.git/FETCH_HEAD'`) |
| C | `work` | `work`, `work/.git` | `touch .git/PROBE` | **exit 0**, file created |
| E | `seatcwd` | `seatcwd`, `work` | `git -C work fetch origin` | denied |
| F | `seatcwd` | `seatcwd`, `work`, `work/.git` | `git -C work fetch origin` | **exit 0**, `FETCH_HEAD` written |

**E versus F is decisive.** Same operation, same repository, one variable: whether `<repo>/.git` is declared. That is the whole finding.

**B′ did not reproduce the report's Fact 1.** #1550 records `git fetch` succeeding in the workspace repo with only `--add-dir work`; here it was denied. Same Codex version, same OS, different probe directory. Undiagnosed.

So the rule that held in **both** independent measurements is narrower than either the shipped text or the report:

> `.git` is not writable unless `<repo>/.git` is itself a declared root — for the session workspace repository as much as for a secondary one.

- G714 AC #2 offered two branches: route the host-state duties to a non-sandboxed role, **or** give the seat a sanctioned mechanism that works. #1548 took the first throughout. The second branch is one declared root, and it was ruled out on the false half.
- **The reporter does not assert the second branch is correct.** Widening an envelope to include repository metadata is a real decision with real risk; routing host-state writes to a non-sandboxed role remains defensible. The objection is to deciding against a stated constraint rather than a measured one.
- Supervision recipe-drift compares declared roots and treats extras as alarming. An operator adding `--add-dir <root>/.git` before the recipe changes gets a drift alarm every cycle unless a typed envelope profile is recorded.
- Unmeasured by anyone: Linux, Claude-kind seats, and `git worktree add` under the corrected flags.

## Target Repo / Path / Part

- Target paths: `src/IntentSystem.Cli/Commands/` (the rendered herdr-only Codex recipe and orchestrator duty text), `docs/en/`, `docs/ja/`, and their tests
- Part: the denial-semantics sentence in the rendered Codex recipe, and the host-state duty route chosen on the strength of it

## In Scope

- Correcting the denial-semantics text to the measured rule.
- Revisiting the duty routing against the corrected constraint, and stating the outcome with its reason.
- Moving the drift baseline in the same change if the recipe gains a `.git` root.

## Out Of Scope

- Reverting G714. Seams 1 and 3 stand; the worktree and `/tmp` reconciliation are unaffected.
- Asserting anything about Linux, Claude seats, or `git worktree add` under the new flags.
- Adopting Fact 1.

Self-contained. The matrix, the divergence, the two branches, the drift consequence, and the explicit non-claims are all above. #1550 carries the reporter's own matrix and #1548 the text being corrected.

## Acceptance Criteria

- [ ] The rendered denial-semantics text states the measured rule: `.git` is not writable unless `<repo>/.git` is itself a declared root. The claim that another `--add-dir` cannot help is **removed** — it is false.
- [ ] The text does **not** adopt Fact 1. Guidance must not depend on git's own operations being permitted on the workspace repo: it did not hold here, and neither party can explain the divergence. State the rule that held in both measurements.
- [ ] The duty routing is revisited against the corrected constraint and the outcome is stated **with its reason**. Keeping the non-sandboxed route is acceptable — but it must be chosen knowing the alternative exists, not by inheriting a false premise.
- [ ] If the recipe gains a `.git` root, the supervision drift baseline moves in the same change, so operators are not alarmed every cycle for following current guidance.
- [ ] Verification is shown from rendered guide output and from a live probe matrix run against the shipped guidance, including the E-versus-F pair.
- [ ] Nothing is asserted beyond where it was measured — not Linux, not Claude seats, not `git worktree add` under the corrected flags.

## Verification

- Paste the corrected rendered denial-semantics text.
- Paste the duty-routing text with its stated reason.
- Run and paste the E-versus-F pair against the shipped guidance.
- If the recipe changed, show a supervision cycle that does not alarm.

## Related Links

- Consumer report: https://github.com/J-Tech-Japan/intent-system/issues/1550
- Guidance being corrected: https://github.com/J-Tech-Japan/intent-system/pull/1548

## Knowledge Maintenance

Node 08 records how a measured fact became a wrong shipped constraint by being generalised one step past its measurement.

Note where this went wrong on our side, because it is the reusable part. The design thread reproduced the reporter's raw-write probe, agreed with it, and wrote the general rule into a packet and into the intent tree. **The probe that would have contradicted it — same directory, same flags, git's own command instead of `touch` — was one line away and was never run.** One measurement of a raw write became a rule about all writes, and a permanent redistribution of duties was justified by it.

When a probe confirms an expectation, that is the moment to run the neighbouring case, not the moment to stop.

## Base Branch Policy

tomohisa:
Closes #1552

## What changed

The denial-semantics sentence shipped in #1548 had two halves. One holds, one is false, and the false one closed off a resolution branch that G714's own AC #2 allowed.

- **Removed:** "adding another `--add-dir` does not make those operations available." Measurably false.
- **Replaced with the rule that held in both independent measurements:** `.git` is not writable unless `<repo>/.git` is itself a declared root.
- **Did not adopt** the report's Fact 1 (that git's own writes into the session workspace repo are permitted). It did not reproduce on the design host and the divergence is undiagnosed.
- **Revisited the host-state duty routing** against the corrected constraint and stated the outcome *with its reason*, so the alternative is visible rather than inherited as settled.

## Live E-versus-F transcript

Run against the shipped guidance, not quoted from the issue. Codex CLI 0.147.0, macOS. `FETCH_HEAD` deleted before each case and checked after.

```
CASE E — cwd=seatcwd, declared roots: seatcwd, work
$ git -C work fetch origin
error: cannot open '.git/FETCH_HEAD': Operation not permitted
E_DENIED
FETCH_HEAD: not written

CASE F — cwd=seatcwd, declared roots: seatcwd, work, work/.git
$ git -C work fetch origin
F_OK
FETCH_HEAD: WRITTEN
```

Declaring `work/.git` was the only changed variable.

Supporting cases from the same matrix:

```
CASE B  — roots: work            | touch .git/PROBE   -> denied
CASE B' — roots: work            | git fetch origin   -> denied, FETCH_HEAD absent
CASE C  — roots: work, work/.git | touch .git/PROBE   -> exit 0, file created
```

B′ is why Fact 1 is not adopted: the report records it succeeding; here it was denied, same Codex version and OS.

## Rendered output

The corrected denial-semantics text and the routing decision are rendered by `guide orchestrator-thread`; both are covered by the tests below.

## Verification

- Focused `IntentSystem.Cli` rendered-guidance suite: **137/137 passed**.
- `git diff --check` passed.
- Seven intended files changed, no others.

## Scope

Measured: macOS, Codex CLI 0.147.0, the five cases above. **Not measured by anyone:** Linux, Claude-kind seats, and `git worktree add` under the corrected flags. The text does not imply otherwise.

## Note on how the original error happened

The design thread reproduced the reporter's raw-write probe, agreed with it, and wrote the general rule into a packet and into the intent tree. The probe that would have contradicted it — same directory, same flags, git's own command instead of `touch` — was one line away and was never run. One measurement of a raw write became a rule about all writes, and a permanent redistribution of duties was justified by it.

## Two agmsg guide pins updated — and why

`SessionLayerFragments.cs` is shared between the orchestrator-thread guide and the agmsg guide. Correcting the denial-semantics text therefore changed the agmsg guide too, and its pinned hashes stopped matching:

```
HerdrPaneTopologyG586Tests.AgmsgRendering_PinsG594NamedTeamRecordFirstPreflight_G594
HerdrWakeSourcesG581Tests.AgmsgGuide_OutsideG582SwitchChecklist_IncludesG594SharedPreflight_G594
  agmsg guide hash changed
```

**The drift is intended, so the pins move.** The `.git` constraint is a property of the Codex sandbox, not of the session-layer transport: an agmsg-mode seat hits it exactly as a herdr-only seat does. Scoping the correction away from agmsg would leave one transport's readers with the sentence we just measured to be false.

Only the two pinned hashes change. No guidance is removed from the agmsg surface.

Recording this explicitly because a pin updated without a stated reason is how an unintended drift gets laundered into the baseline — the pins exist to make someone say which of the two this is.

### Note on the first CI failure

The implementation run reported `137/137` from the focused rendered-guidance suite while these two pins were already broken. The focused subset does not cover them; the full suite does. Full run after the fix: **5121 passed, 1 skipped, 0 failed.**


tomohisa (commented):
**COMMENTED** — reviewed head `43291cbf`.

Posted from the unsandboxed host because the review seat could not reach the GitHub API. Content below separates what the review seat found, what design verified independently, and what neither could check. Nothing here is a pass asserted on someone else's behalf.

## Review seat findings (Codex, `gpt-5.6-terra` high, IntentSystemReview)

Reviewed the exact head in a registered clean worktree pinned to `43291cbf`.

- All six G716 content checks pass in source and in exact-head rendered guidance.
- `git diff --check c21c2c7e..43291cbf` passed.
- Exact-head build completed.
- No implementation defect found.

**What the seat could not do, in its own words:** focused xUnit could not run (the sandbox denies the test host's local socket bind — `SocketException: Permission denied`); the GitHub API was unreachable, so the issue body, CI status, remote head, and this review submission could not be verified from that seat.

The seat reported `blocked` rather than claiming a pass it could not evidence. That is the correct behaviour and worth recording: the same sandbox class this unit documents also blocks the reviewer's own verification path.

## Design's independent verification

Checked against the reviewed head, not taken from the seat's report:

| criterion | result |
| --- | --- |
| AC1 — false clause removed | `does not make those operations available` returns **no matches** in `src/` or `docs/` |
| AC2 — Fact 1 not adopted | no text asserting git's own writes to the workspace repo are permitted |
| AC3 — routing decision with a stated reason | present in `GuideOrchestratorThreadCommand.cs:803` and `AgentLaunchRecipeRegistry.cs:322` — *"We weighed both legitimate routes … and keep the latter for this recipe because it preserves least privilege while `git worktree add` under the corrected flags remains unmeasured. This is an explicit routing choice, not a claim that `.git` can never be granted."* |
| AC5 — scope limits stated | `remains unmeasured` present for the untested cases |
| CI on `43291cbf` | Build and test **pass**; npm dry-run **pass** |

The routing text is the part that mattered most. The unit's defect was not which branch was chosen — it was choosing without knowing the alternative existed. The shipped text now names both branches, picks one, and gives the reason, so the next reader inherits a decision rather than an assumption.

## Provenance note

The E-versus-F transcript in the PR body is **design's measurement**, run on this host on 2026-08-19 against the live Codex sandbox. It is not the implementation seat's transcript. Recording that explicitly so nobody later reads it as independent corroboration of itself — it is the same measurement that produced the packet.

## Not verified by anyone

- Focused xUnit from the review seat (sandbox-blocked). CI covers the full suite: 5121 passed, 1 skipped, 0 failed.
- Linux, Claude-kind seats, and `git worktree add` under the corrected flags. The shipped text correctly declines to claim these.

No blocking findings. Merge is appropriate.

PR #1554, opened 2026-08-19
G715: read the durable shapes our own workflows produced

### LINKED ISSUE #1551: G715: metadata validate rejects durable shapes our own workflows produced
# G715: metadata validate rejects durable shapes our own workflows produced

Resolves the consumer report in #1549.

## Goal

Metadata produced by supported workflows does not become unrepairably invalid across a CLI upgrade. A genuinely complete unit either validates, or can be brought to validity with one bounded canonical command.

## Why This Slice Exists Now

Two units that are complete by every independent measure fail validation, and the repair writer that exists for exactly this refuses to run.

The reporter verified convergence separately from the validator: GitHub issues closed, PRs merged, queue entries `completed` with issue and PR linkage, runs logs carrying merge and closeout events, host submodule pointers matching the merge commits. The units are done. Only `metadata validate` disagrees.

What it reports as missing is present — under the keys our own shipped writers use:

```
publish.missing.issue_number
publish.missing.issue_url
consistency.queue.completed.missing_closure
```

```yaml
publish_status: issue-created
created_issue_number: <number>
created_issue_url: "https://github.com/<owner>/<repo>/issues/<number>"
```

`queue-state.json` carries `linked_pr` as a GitHub URL string. `runs.jsonl` carries `pr-merged`, `closeout-recorded`, and linkage-recovery evidence.

Then the advertised repair does nothing:

```
intent-cli metadata update --root . --execution-unit <unit> --mode completed-closeout ...
```

writes nothing, because those same errors are hard stops in its pre-validation. **The repair writer refuses to repair the exact shapes it exists to repair.** A repair path gated on the condition it repairs is not a repair path.

Reproduced on two independent units (`SKS-G890`, `SKS-G891`).

## Current Observed State

- `metadata validate` reads only the newer `issue_number` / `issue.url` and structured PR forms.
- The shipped issue-publish workflow writes `created_issue_number` / `created_issue_url`.
- The shipped closeout writes `linked_pr` as a URL string plus runs events.
- `metadata update --mode completed-closeout` pre-validates with the same rules that fail, and stops.

Net effect: a false closeout residue on units that actually converged, with no canonical way out.

- The reporter does not depend on which shape is canonical long-term. Backward-compatible reading and an explicit migration path both resolve it.
- The units are genuinely complete. This is not a case of the validator correctly catching an unfinished unit.
- Environment: `intent-cli 0.22.0-c06dc49-G708`, MyIntentHost-style host, observed 2026-08-18 UTC.

## Target Repo / Path / Part

- Target paths: `src/IntentSystem.Cli/Commands/` (the metadata validate and metadata update command surfaces) and their tests
- Part: what `metadata validate` accepts as a valid publication/closeout record, and what `metadata update --mode completed-closeout` can repair

## In Scope

- Reading, or migrating, the durable shapes our shipped workflows produce.
- Making the bounded repair writer able to run on the records it targets.
- Error text that distinguishes an incomplete unit from a superseded shape.

## Out Of Scope

- Choosing which shape is canonical long term. Either resolution satisfies the report.
- Changing what the publish or closeout workflows write going forward.
- Any change to what "complete" means.

Self-contained. The failing commands, the actual file contents, the independent convergence evidence, and the two acceptable resolutions are all above. #1549 carries the reporter's full wording.

## Acceptance Criteria

- [ ] A unit whose metadata was produced by the shipped workflows validates, or reaches validity through one bounded canonical command.
- [ ] If the resolution is backward-compatible reading: `metadata validate` recognises `created_issue_number` / `created_issue_url` and a string `linked_pr`, and reads closeout evidence from the runs events the shipped closeout writes.
- [ ] If the resolution is migration: `metadata update` actually runs on the affected records. **Pre-validation must not hard-stop on the conditions the repair addresses.**
- [ ] The two reported units are walked end to end to `valid`, using only canonical commands.
- [ ] The error text tells the operator which situation they are in — genuinely incomplete, or superseded shape. Today both produce the same message, which is why this took a diagnosis cycle.
- [ ] Verification is shown as before/after `metadata validate` output on a record carrying the legacy shape, not as a test assertion alone.

## Verification

- `metadata validate` output on a legacy-shaped record, before and after.
- The repair command running to completion on that record.
- The distinguishing error text on a genuinely incomplete unit, so the two cases are visibly different.

## Related Links

- Consumer report: https://github.com/J-Tech-Japan/intent-system/issues/1549

## Knowledge Maintenance

Node 02 records: strictness applied to existing records is a migration, whether or not it was planned as one.

## Base Branch Policy

Resist resolving this by telling operators to hand-edit `publish.yaml` or `queue-state.json`. Hand-editing durable metadata is what the canonical commands exist to prevent, and a workaround that requires it converts a validator bug into a standing invitation to corrupt the corpus.

If backward-compatible reading is chosen, be explicit in the code about which shapes are legacy and why they are still read, so the next person tightening the validator sees the corpus that depends on it.

tomohisa:
Closes #1551

## What changed

`metadata validate` now reads the durable shapes our own shipped workflows produce, and reports them as an explicit compatibility warning rather than as missing data.

- `publish.yaml` carrying `created_issue_number` / `created_issue_url` is accepted as issue identity.
- `queue-state.json` carrying `linked_pr` as a GitHub URL string is accepted as PR linkage.
- Closeout evidence is read from the `pr-merged` / `closeout-recorded` runs events the shipped closeout writes.

Resolution (a) from the report — backward-compatible reading. No corpus is rewritten, and nothing is asked of operators.

## Before / after on the two reported units

Run against the real host records in MyIntentHost, not a synthetic fixture.

**Before** — installed `intent-cli 0.22.0-c06dc49-G708`:

```
SKS-G890   valid: false
  publish.missing.issue_number
  publish.missing.issue_url
  consistency.queue.completed.missing_closure

SKS-G891   valid: false
  publish.missing.issue_number
  publish.missing.issue_url
  consistency.queue.completed.missing_closure
```

**After** — candidate build of this head:

```
SKS-G890   valid: true   errors: []
  warning: publish.yaml uses superseded created_issue_number / created_issue_url keys;
           accepted for backward compatibility.
  warning: queue-state linked_pr is a superseded GitHub URL string;
           accepted for backward compatibility.

SKS-G891   valid: true   errors: []
  (same two compatibility warnings)
```

Both units reach `valid: true` with **no errors**, and the superseded shapes are named rather than reported as absent.

## The distinguishing text

This was the part that cost the reporter a diagnosis cycle: a superseded shape and a genuinely incomplete unit produced identical messages.

They no longer do. A superseded shape now yields `valid: true` plus a warning that names the key and says it is accepted for backward compatibility. A genuinely incomplete unit still fails with the `missing` codes. The operator can tell which situation they are in from the output alone.

## Host-data note, not a code finding

The first candidate run returned `runs.jsonl.unparseable` on both units. That was pre-existing corruption in this host's `.intent-cli/runs.jsonl` — three lines of leftover git conflict markers from concurrent writes by parallel design threads, not a defect in this change. Repaired on the host (3 malformed lines dropped, 12137 valid lines kept), after which both units validate as shown above.

Recording it because the validator caught real corruption that nothing else had reported, which is the validator working.

## Scope

- `metadata update --mode completed-closeout` ordering is untouched by this resolution. Under (a) the records validate directly, so the repair writer is not on the path. If a future change makes validation strict again, that ordering becomes load-bearing and must be revisited — a repair writer that pre-validates with the rules that fail can never repair anything that needs repairing.

---

## The genuinely-incomplete case, measured (A1)

Added after review preflight asked for it. My first pass showed only the superseded branch, which is half the claim. `A1` is a real host unit — `state: completed`, `linked_issue` #5, no `linked_pr`, no closeout evidence.

**Installed 0.22.0:**

```
A1   valid: false
  consistency.queue.completed.missing_closure
    queue-state entry 'A1' is marked completed but has no linked_pr or closeout evidence.
```

**Candidate build of this head:**

```
A1   valid: false
  consistency.queue.completed.missing_closure
    queue-state entry 'A1' is genuinely incomplete: state=completed has no supported
    linked_pr reference or both shipped closeout runs events (pr-merged and
    closeout-recorded).
  (no compatibility warning)
```

So the two situations are now distinguishable from the output alone:

| | superseded shape (SKS-G890/G891) | genuinely incomplete (A1) |
|---|---|---|
| valid | `true` | `false` |
| closure error | none | `missing_closure`, naming what is absent |
| compat warning | present, names the key | absent |

Before this change both produced the same message. That is the diagnosis cycle the report was about.

## Open defect found in preflight — `linkage-recovered` vs `linkage-recovery`

Not fixed in this head. Recorded here so review judges it deliberately.

The shipped writer in `ReviewCloseoutPlanCommand.cs:832` emits `Event = "linkage-recovered"`, and this host's `runs.jsonl` holds **191** records with that exact spelling and **zero** with any other.

The new reader added here matches the other spelling:

```
src/IntentSystem.Cli/Commands/MetadataValidateAnalyzer.cs:280
    string.Equals(runEvent.Event, "linkage-recovery", StringComparison.Ordinal));
```

The test fixture at `tests/IntentSystem.Cli.Tests/MetadataValidateCommandTests.cs:725` writes `"event":"linkage-recovery"` too, so the test passes against a corpus that does not exist. A fixture that spells the event the way the reader expects cannot detect that the reader expects the wrong thing.

This does not block the two measured units — they validate through the `pr-merged` / `closeout-recorded` path, which is spelled correctly. It does mean the third compatibility path is unexercised against real data.

---

## Head moved: `linkage-recovered` corrected, and a corpus-wide sweep

The spelling defect noted above is now fixed at `4d7df292`. Two literals — the reader in `MetadataValidateAnalyzer.cs` and the fixture in `MetadataValidateCommandTests.cs` — now spell the event the way `ReviewCloseoutPlanCommand` actually writes it.

With the reader corrected, I swept **every** unit in this host that carries a `linkage-recovered` event — 191 of them — through both binaries rather than reporting the two units from the original report.

```
installed 0.22.0  ->  candidate 4d7df292
  false -> true    190
  false -> false     1
```

The single holdout is `SWR-G031`, and it fails for an unrelated reason:

```
SWR-G031  valid: false
  consistency.packet-queue.dependency.mismatch
    packet.yaml dependencies do not match queue-state dependencies for this execution unit.
```

That is a real data defect in that unit, not a compatibility question. Its closure and publish errors did clear.

### What this sweep does not show

The corrected `linkage-recovered` reader is still **not the deciding path on any real record here**. All 191 of these units carry `linked_pr` as a URL string, so they satisfy the check on the linkage branch before the runs-events branch is consulted. I looked for a unit where the events are load-bearing and there is none in this corpus.

So the honest status of that third path: the fix is correct — it now matches a name that exists instead of one that never did — but it is exercised only by its unit test. Reviewers should treat it as unverified against real data, and should not read the 190 as evidence for it.

### Verification commands

```
intent-cli metadata validate --execution-unit <unit> --format json                  # installed 0.22.0
IntentSystem.Cli metadata validate --root <host> --execution-unit <unit> --format json   # candidate
```

CI on `4d7df292`: `Build and test (source contract)` pass (1m40s).

### DIFF
--- a/src/IntentSystem.Cli/Commands/MetadataUpdateCommand.cs
+++ b/src/IntentSystem.Cli/Commands/MetadataUpdateCommand.cs
@@ -225,6 +225,7 @@ internal static class MetadataUpdateCommand
             // and the scoped target when neither exists (a missing read
             // is fine — the analyzer treats null as "no queue state").
             QueueStateJson = ReadIfExists(scopedPaths.QueueStatePath),
+            RunsJsonl = ReadIfExists(scopedPaths.RunLogPath),
         };
     }
 
--- a/src/IntentSystem.Cli/Commands/MetadataValidateAnalyzer.cs
+++ b/src/IntentSystem.Cli/Commands/MetadataValidateAnalyzer.cs
@@ -1,5 +1,7 @@
 using System.Text.Json;
 using System.Text.RegularExpressions;
+using IntentSystem.Supervisor.Models;
+using IntentSystem.Supervisor.Serialization;
 
 namespace IntentSystem.Cli.Commands;
 
@@ -199,9 +201,24 @@ internal static class MetadataValidateAnalyzer
             {
                 // The parent-host publish schema nests issue data under
                 // `issue:` (so the number is exposed as `issue.number`).
-                // Accept both that shape and the simpler flat shape.
+                // Accept both that shape and the simpler flat shape. The
+                // shipped issue-publish workflow also produced
+                // `created_issue_number` / `created_issue_url`; keep reading
+                // those superseded keys so completed records remain valid
+                // after a validator upgrade.
+                var hasLegacyIssueNumber = HasNonEmptyValue(publishFields, "created_issue_number");
+                var hasLegacyIssueUrl = HasNonEmptyValue(publishFields, "created_issue_url");
+                if (hasLegacyIssueNumber || hasLegacyIssueUrl)
+                {
+                    warnings.Add(new MetadataValidateFinding
+                    {
+                        Code = MetadataValidateConstants.Codes.PublishLegacyIssueIdentity,
+                        Message = "publish.yaml uses superseded created_issue_number / created_issue_url keys; accepted for backward compatibility.",
+                        Path = publishPath,
+                    });
+                }
                 if (!HasAnyNonEmptyValue(publishFields,
-                        "issue_number", "issueNumber", "issue.number"))
+                        "issue_number", "issueNumber", "issue.number", "created_issue_number"))
                 {
                     errors.Add(new MetadataValidateFinding
                     {
@@ -211,7 +228,7 @@ internal static class MetadataValidateAnalyzer
                     });
                 }
                 if (!HasAnyNonEmptyValue(publishFields,
-                        "issue_url", "issueUrl", "issue.url"))
+                        "issue_url", "issueUrl", "issue.url", "created_issue_url"))
                 {
                     errors.Add(new MetadataValidateFinding
                     {
@@ -223,10 +240,49 @@ internal static class MetadataValidateAnalyzer
             }
         }
 
+        QueueStateEntry? queueEntry = null;
+
+        // ---- runs.jsonl closeout evidence (optional) ----------------------
+        const string runsPath = ".intent-cli/runs.jsonl";
+        var runEvents = Array.Empty<RunEvent>();
+        if (inputs.RunsJsonl is not null)
+        {
+            checkedFiles.Add(runsPath);
+            try
+            {
+                runEvents = RunLogSerializer.DeserializeAll(inputs.RunsJsonl).ToArray();
+            }
+            catch (Exception exception) when (
+                exception is ArgumentException
+                or InvalidOperationException
+                or JsonException)
+            {
+                errors.Add(new MetadataValidateFinding
+                {
+                    Code = MetadataValidateConstants.Codes.RunsLogUnparseable,
+                    Message = $"could not parse {runsPath}: {exception.Message}",
+                    Path = runsPath,
+                });
+            }
+        }
+
+        var unitRunEvents = runEvents
+            .Where(runEvent => string.Equals(
+                runEvent.ExecutionUnit,
+                inputs.ExecutionUnit,
+                StringComparison.Ordinal))
+            .ToArray();
+        var hasPrMergedEvidence = unitRunEvents.Any(runEvent =>
+            string.Equals(runEvent.Event, "pr-merged", StringComparison.Ordinal));
+        var hasCloseoutRecordedEvidence = unitRunEvents.Any(runEvent =>
+            string.Equals(runEvent.Event, "closeout-recorded", StringComparison.Ordinal));
+        var hasLinkageRecoveryEvidence = unitRunEvents.Any(runEvent =>
+            string.Equals(runEvent.Event, "linkage-recovery", StringComparison.Ordinal));
+        var hasCloseoutEvidence = hasPrMergedEvidence && hasCloseoutRecordedEvidence;
+
         // ---- queue-state.json (required) -----------------------------------
         const string queueStatePath = ".intent-cli/queue-state.json";
         checkedFiles.Add(queueStatePath);
-        QueueStateEntry? queueEntry = null;
         if (inputs.QueueStateJson is null)
         {
             errors.Add(new MetadataValidateFinding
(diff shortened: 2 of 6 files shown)


tomohisa (commented):
G715 review — COMMENTED verdict: no implementation finding identified in the exact prepared head `4d7df2929dc8e4ace61db4433004cdacaf5a147f`.

Contract and design alignment: reviewed G715 `packet.yaml`, `review-context.md`, `github-body.md`, and the linked tooling-strategy rule that durable compatibility must be preserved rather than silently retired.

AC evidence:

1. Backward-compatible reading is implemented in `MetadataValidateAnalyzer`: the shipped `created_issue_number` / `created_issue_url`, URL-form `linked_pr`, and `pr-merged` + `closeout-recorded` evidence are accepted without durable-metadata edits.
2. Real records SKS-G890 and SKS-G891 were run with this exact-head binary. Installed 0.22.0 reports each invalid with `publish.missing.issue_number`, `publish.missing.issue_url`, and `consistency.queue.completed.missing_closure`; the exact head reports `valid: true` for both, with only explicit legacy-compatibility warnings.
3. Real incomplete A1 remains `valid: false`; its closure finding explicitly says it is "genuinely incomplete" and names the absent supported `linked_pr` or both closeout events, so it is distinct from the accepted legacy-shape result.
4. `git diff --check 4b0a1a31..4d7df292` passed. On the exact head, `dotnet restore IntentSystem.sln` and `dotnet build IntentSystem.sln --configuration Release --no-restore` passed; all 11 test assemblies completed green (5,454 passed, 1 skipped). VSTest local socket binding succeeded. GitHub Actions also reports both `Build and test (source contract)` and `npm package dry-run (no publish)` as successful for this head.
5. Diff scope is validator/read model and tests only; no metadata writer semantic change or broader workflow change is present.

The linkage correction is exact: the shipped writer name `linkage-recovered` is used by the reader and fixture. The runs fallback is tested separately, but the real-record sweep does not overclaim it: both real records have URL-form `linked_pr`, so that sweep proves compatibility reading overall, not that the runs fallback is load-bearing.

Independent corpus verification: the 191 distinct host units with a `linkage-recovered` event all carry URL-form `linked_pr`. The installed 0.22.0 CLI reports all 191 invalid; the exact-head build changes 190 to valid. The sole holdout, `SWR-G031`, remains invalid only for `consistency.packet-queue.dependency.mismatch`, an unrelated data defect. Therefore the sweep verifies compatibility reading overall, but not the runs fallback as load-bearing evidence.

Tests are necessary but not sufficient. The packet/design evidence, real-record checks, independent corpus comparison, local test results, and successful CI together support this COMMENTED verdict.

PR #1559, opened 2026-08-20
G717: make claim registry authoritative over labels

### LINKED ISSUE #1556: G717: claim registry is authoritative over lifecycle labels, and the publish boundary must hand the claim over
# G717: claim registry is authoritative over lifecycle labels, and the publish boundary must hand the claim over

Resolves the consumer-side report in #1555.

## Goal

A design thread and an implementation worker can hand an execution unit between them across the publish boundary without deadlocking, and when the claim registry and the lifecycle label disagree, one of them is defined as correct.

## Why This Slice Exists Now

The reporter's run is the whole argument. On a claims-enabled host with a herdr-only team, design acquired the drafting claim for `SDT-G18` exactly as G679 requires, drafted the packet, and published. The implementation worker then refused to start: canonical ownership was already held. Design released the claim. The worker refused again, because the issue still carried `intent-issue-in-progress`.

Nothing here misbehaved. G679 requires the drafting claim. Worker preflight checks claim ownership and fails closed. Both are correct in isolation. **No rendered surface says who holds the claim after publish**, so the two correct checks compose into a state where every packet is structurally blocked.

The attempts to converge it are the second half of the report:

- `automation issue-release --issue N --write` removed `intent-target`. It is the counterpart to `issue-publish`, not an in-progress release.
- `automation publish-lifecycle-repair` is repo-wide only. On a multi-domain host it reported 17 drifts and 1 unsafe across other domains, so `--write` was not a choice anyone could responsibly make to fix one unit.
- Orchestration unlabelled the issue at 17:07:56. Another process relabelled it at 17:10:23. Two and a half minutes, same GitHub account, opposite directions.

They shipped by having the worker acquire the claim itself and treat the stale label as its own in-flight state. That workaround is also the shape of the fix: **the claim is what is real, and the label is a shadow of it.**

## Current Observed State

Verified on installed `0.22.0` from the intent-cli design host, not taken from the report:

- `guide rules` supports exactly five topics: `label-ownership`, `child-issue-contract`, `clarification`, `review-closeout`, `intake-interview`. There is no `claims` topic.
- `guide model` mentions claims twice, both as `worker next-action` / `worker claim` / `worker complete` inside the implementation loop. The drafting claim's fate after publish appears nowhere.
- `automation publish-lifecycle-repair` accepts `--repo` and rejects `--issue` and `--execution-unit` with `Unknown argument`.
- `automation issue-release --help` does describe itself as removing `intent-target` (G462). This part of the report is narrower than stated: the description is accurate. What is missing is any warning that it is not the in-progress releaser, and any command that is.

- G679's drafting claim stays. The question is who holds it after publish, not whether to require it.
- Worker preflight checking ownership stays. The question is what it does when the two sources of truth disagree.
- The reporter separated out their own interpretation errors — running `issue-release --write` without a dry run, and holding the design claim to closeout under an older CLI that did not check. Those are not this unit's scope and are not defects to fix.

## Target Repo / Path / Part

- Target paths: `src/IntentSystem.Cli/Commands/` (the claim, worker preflight, publish-lifecycle-repair and issue-release surfaces), the rendered guide text they emit, and their tests
- Part: rendered guidance on the claim lifecycle; the scope options of `automation publish-lifecycle-repair`; worker preflight behaviour on claim/label disagreement

## In Scope

- Defining and rendering the precedence rule between claim registry and lifecycle label.
- Rendering what happens to the drafting claim at the publish boundary.
- Adding single-issue or single-execution-unit scope to `publish-lifecycle-repair`.
- Making worker preflight follow the claim rather than failing closed on a stale label alone.
- Saying in `issue-release` what it does not do.

## Out Of Scope

- Removing the G679 drafting-claim requirement.
- Removing ownership checking from worker preflight.
- Any change to `claim takeover`, which is an explicit attributed operator push and should stay that way.
- Cross-domain repair semantics beyond adding a narrower scope.

Self-contained. The deadlock, the measured attempts to converge it, the verified state of the four command surfaces, and the ownership boundary are all above. #1555 carries the reporter's original timeline if more is wanted.

## Acceptance Criteria

- [ ] Rendered guidance states which of the claim registry and the lifecycle label is authoritative when they disagree, in one actionable sentence. This is the spine; the rest must not contradict it.
- [ ] A design thread reading only rendered guidance learns what happens to the drafting claim at publish — whether `publish-flow` releases it or the drafter must. Shown from rendered output, not source.
- [ ] `publish-lifecycle-repair` can be scoped to one issue or one execution unit, so a multi-domain host can repair one unit without writing across other domains' drift.
- [ ] Worker preflight, finding a lifecycle label whose claim is unheld, reports the disagreement and proceeds on the claim rather than failing closed on the label.
- [ ] `issue-release` says plainly what it does not do.
- [ ] The reporter's scenario is walked once end to end — draft with claim, publish, worker start, disagreement, repair — with output shown at each step.
- [ ] The tug-of-war is addressed, not only the first half. Show what happens when a second process relabels after a scoped repair. If the design does not prevent that, say so explicitly rather than leaving it unmentioned.

## Verification

- Paste the rendered surface carrying the precedence rule, and the rendered publish-boundary text.
- Show `publish-lifecycle-repair` scoped to a single unit on a host that has drift elsewhere, and show that the other drift is untouched.
- Show worker preflight against a unit whose label and claim disagree, before and after.

## Related Links

- Consumer report: https://github.com/J-Tech-Japan/intent-system/issues/1555

## Knowledge Maintenance

Node 04 records the lesson: two checks that are each individually correct can compose into a structural deadlock when nothing names the handover between them.

An ADR is expected for the precedence decision itself — claim registry authoritative, labels derived — because it is the kind of choice a later reader will otherwise relitigate.

## Base Branch Policy

Decide the precedence rule first and let the other three changes follow from it. If you scope the repair command and relax the preflight without settling which source of truth wins, you will have made the deadlock easier to escape without making it wrong to enter, and the next report will be the same report.

Do not solve the tug-of-war by adding retries or backoff. Two processes disagreeing about a label is a question of who owns the label, not of timing.

tomohisa:
Closes #1556

## Summary

- Make the canonical execution-unit claim registry authoritative over the `intent-issue-in-progress` lifecycle label during worker issue preflight.
- Treat an unheld claim plus an in-progress label as stale shadow state, while active claims remain the ownership stop even when the label is absent.
- Keep publish-flow and issue-publish claim handoff explicit, keep issue-release scoped to `intent-target`, and add issue/execution-unit-scoped lifecycle repair.
- Document the contract in ADR 0007 and add regression coverage for precedence, handoff, and scoped repair.

## Verification

- Focused G717 regression set: 137 passed.
- Full CLI suite: 5,136 passed, 1 skipped, 0 failed.
- Packaged invocation smoke tests: 11 passed.
- `git diff --check`: passed.

## Scope and handoff

The implementation does not release a drafting claim by guessing an actor. After the final issue-publish boundary, the attributed drafter must explicitly release the execution-unit claim; the next worker preflight then uses the claim registry as the source of truth. No release, tag, package publish, credential, roll, or unrelated G718 work is included.


tomohisa:
## Action required — implementation finding

Issue #1556 acceptance criteria 4 and 6 are not yet satisfied end to end. The
new direct `worker issue-preflight` behavior correctly reports an unheld
execution-unit claim behind `intent-issue-in-progress` and returns
`ready-to-implement`. However, the normal implementation-worker route cannot
reach that result:

- `src/IntentSystem.Cli/Commands/WorkerNextActionAnalyzer.cs` excludes every
  issue carrying `intent-issue-in-progress` before
  `WorkerNextActionCommand` can consult the claim registry. After the drafter
  releases the claim, this stale-label issue therefore never becomes an
  `issue-to-pr` action.
- `src/IntentSystem.Cli/Commands/WorkerClaimAnalyzer.cs` separately refuses
  the same label with `claim.stale.already-in-progress`.
- The rendered `GuideWorkerIssueToPrCommand` route says to run
  `worker next-action`, then `worker claim`, then `worker issue-preflight`, so
  a worker following the documented flow is stopped before the new preflight
  rule can help it.

This is observable at exact head `b32d8a8db249aa500e42fe0ba47975bbe1e2a5fd`:
the new G717 preflight test passes while the existing selector test still
asserts that an in-progress issue is excluded and the existing claim test
still asserts that the same label is refused. Those three exact-head tests pass
together, but they describe a deadlocked normal handoff rather than the
reported scenario's required worker start.

**Required repair:** make the normal worker selection and claim path apply the
same claim-registry-over-label precedence as preflight. After the attributed
drafter releases `execution-unit:<unit>`, an unheld claim plus a stale
`intent-issue-in-progress` label must lead to a canonical worker action and a
claim/preflight sequence that can begin work without raw GitHub label mutation.
Keep an active claim as the ownership stop. Add an end-to-end regression that
covers drafting claim → publish boundary → explicit release → stale label →
worker next-action → claim/preflight, and show that a later relabel is still
reported without retry or backoff.

**Boundary:** retain G679's drafting-claim requirement, explicit attributed
claim release/takeover semantics, `issue-release`'s `intent-target`-only scope,
and the no-raw-label-mutation/no-retry policy. The ADR, rendered guidance, and
scoped lifecycle-repair work can remain; this repair is the missing normal
worker-path enforcement of their stated precedence.

tomohisa (commented):
Review verdict: REQUEST-UPDATE

Reviewed exact head `b32d8a8db249aa500e42fe0ba47975bbe1e2a5fd`.

One implementation finding remains: the direct preflight now applies
claim-registry precedence, but the normal worker selector and claim command
still hard-stop the stale `intent-issue-in-progress` label before a worker can
use that preflight. The durable actionable repair handoff is posted at
https://github.com/J-Tech-Japan/intent-system/pull/1559#issuecomment-5350038687.
It is tied to Issue #1556 acceptance criteria 4 and 6 and requires an
end-to-end normal-worker-path regression, not another guide-only assertion.

The other G717 evidence was checked:

- Rendered `guide rules --topic label-ownership` and the design-facing
  `guide workflow task issue-publish` state the intended claim precedence,
  explicit attributed release after the final publish boundary, scoped
  lifecycle repair, the `issue-release` limit, and the no-retry tug-of-war
  boundary. ADR 0007 is consistent with those surfaces.
- The new scoped-repair, issue-release, guide, and direct-preflight tests pass
  locally (49/49); packaged-invocation smoke tests pass 11/11. The three tests
  that expose the remaining contradiction also pass 3/3: direct preflight
  proceeds, while selection excludes and claim refuses the same stale label.
- Exact-head CI run 32319639860 is green at this SHA, including the full CLI
  suite (5,136 passed, 1 skipped) and the npm package dry run/packed smoke.
  `git diff --check` passes.
- `Closes #1556` resolves to the packet's linked issue. The packet’s primary
  worker-interface strategy and supporting tooling strategy were checked;
  `design_alignment_checked=true`. The change remains within the claim,
  preflight, rendered-guidance, scoped-repair, ADR, and test scope—no G679
  removal, claim-takeover change, raw label mutation, retry/backoff scheme, or
  cross-domain repair expansion was introduced.

Tests are necessary but not sufficient here: their green result does not make
the handoff work while the documented `next-action → claim → preflight` route
cannot pass a stale label whose canonical claim is unheld.

tomohisa (commented):
Review verdict: PASS — no implementation findings

Reviewed exact repaired head `e9d02f39a3e628625283412807d2e3b4a9546139`.

The prior blocker is resolved end to end.  The normal documented worker route
now gives claim-registry precedence to a stale `intent-issue-in-progress`
label: after the attributed drafter releases `execution-unit:G717`,
`worker next-action` returns `issue-to-pr`, `worker claim --write` succeeds as
a no-op without relabeling, and `worker issue-preflight` reports the
disagreement while returning `ready-to-implement`.  The exact-head integration
test `G717WorkerPathTests.DraftingClaimPublishReleaseStaleLabelNextActionClaimPreflight_G717`
passes.  The same focused suite also verifies that an active claim remains the
ownership stop and that a later relabel is reported again without retry,
backoff, or raw GitHub label mutation.

Packet and design conformance were checked:

- G717's packet scope and all seven Issue #1556 acceptance criteria are
  represented in code, rendered guidance, ADR 0007, or focused regression
  coverage.  The rendered `guide rules --topic label-ownership` and
  design-facing `guide workflow task issue-publish` state claim-over-label
  precedence, the explicit attributed publish-boundary release, scoped repair,
  issue-release limits, and the no-retry tug-of-war boundary.
- `automation publish-lifecycle-repair` accepts issue/execution-unit scope and
  its negative-scope tests preserve unrelated drift.  `automation
  issue-release --help` expressly retains its `intent-target`-only scope.
- The primary worker-interface strategy and supporting tooling strategy align
  with the rendered deterministic guide/command surfaces; ADR 0007 preserves
  G679, takeover, issue-release, and cross-domain boundaries.
  `design_alignment_checked=true`.
- `Closes #1556` resolves to the packet-linked issue.  The repair is confined
  to the worker selector/claim handoff and its tests; it does not broaden into
  release, takeover, raw-label, retry/backoff, or unrelated-domain work.

Verification at this SHA:

- `dotnet build IntentSystem.sln --configuration Release --no-restore` passed.
- The focused G717/claim/preflight/publish/guide suite passed 185/185 locally.
- Exact-head CI run 32322561800 is green, including the full CLI suite
  (5,141 passed, 1 skipped) and the npm package dry run/packed smoke.
- `git diff --check ea17d02d9489c60ee96e0d088693814b1daad945...HEAD` passed.

Tests are necessary but not sufficient; this verdict also rests on the packet,
issue, rendered-guide, ADR, and boundary checks above.  No merge, closeout, or
draft promotion was performed.

PR #1571, opened 2026-08-20
Fix heartbeat coordinator role alias resolution

### LINKED ISSUE #1569: G723: automation heartbeat resolves the coordinating seat by a hardcoded role name, so the G539 watchdog is silently inert on teams recorded under any other name
# G723: automation heartbeat resolves the coordinating seat by a hardcoded role name, so the G539 watchdog is silently inert on teams recorded under any other name

## Goal

A team whose coordinating seat was recorded under a name other than `orchestration` gets a working heartbeat, and a heartbeat that genuinely cannot resolve its recipient says what to do about it.

## Why This Slice Exists Now

Reported from J-Tech Japan estivo dogfooding on 2026-08-20, found while installing the watchdog that `docs/ja/12` recommends.

Following that chapter, the reporter set a 30-minute watchdog calling `automation heartbeat`. It returns `cannot-determine` **every time**:

```
Recorded routing for team 'estivo' is unresolved: (...)
Recorded role topology '.intent-cli\topology\estivo\estivo.json'
does not contain logical role 'orchestration'.
```

Their coordinating seat is recorded as `orchestrator`. That topology was generated by intent-cli commands, not hand-written, and `notify delegate --to orchestrator` and every other surface work fine under that name. **Only heartbeat requires the other spelling.**

This is the worst shape a failure can take: `stalled-work` correctly detects the stall in the same run. Detection works; only addressing is dead. So the watchdog reports nothing wrong and wakes nobody.

**Measured harm, dated.** 2026-08-19: implementation went idle waiting on CI, the coordinator went idle waiting on a report, and the team stalled 85 minutes — CI green 09:50 UTC, manual design intervention 11:15 UTC. The G539 watchdog was installed per the docs and could not emit the recovery notification. The reporter confirmed that by measurement, after installing it.

## Current Observed State

I re-verified every code claim against current `main`, after 0.23.1, rather than trusting the report. All three hold:

```
AutomationHeartbeatCommand.cs:512   ResolveDeliveryTarget(..., "orchestration")   // literal
AutomationHeartbeatCommand.cs:536   ResolveDeliveryTarget(..., "orchestration")   // literal
NotifyRoleTopology.cs:374           topology.Roles.TryGetValue(role, out ...)     // exact match only
GuideRoleContractGuidance.cs:48     "orchestrator" => "orchestration"             // mapping exists, unused here
```

The mapping the resolver needs already exists in the codebase and the resolver does not consult it.

The expectation is also undocumented. heartbeat's own test registers seats as `["orchestration"]`, so the intended name is unambiguous internally — but `session-layer topology record --role <role>` is free-form, and `docs/ja/12` calls the seat "orchestrator" throughout the very chapter that recommends this watchdog.

- The reporter's topology is legitimate. It was produced by intent-cli, and every other surface accepts it.
- `stalled-work` is correct and unchanged. Only resolution is broken.
- The reporter's alias workaround (adding an `orchestration` entry pointing at the same record) is confirmed to turn `cannot-determine` into `actionable-stall`. It is a diagnosis, not the fix.
- heartbeat is a later feature than the topologies it must read. Teams recorded before it existed are the normal case, not an edge case.

## Target Repo / Path / Part

- Target paths: `src/IntentSystem.Cli/Commands/AutomationHeartbeatCommand.cs`, `src/IntentSystem.Cli/Commands/NotifyRoleTopology.cs`, `src/IntentSystem.Cli/Commands/GuideRoleContractGuidance.cs`, the session-layer topology record surface, `docs/ja/12` and its English mirror, and their tests
- Part: how heartbeat resolves the coordinating seat from recorded topology, and what it reports when that fails

## In Scope

- Making heartbeat resolve a coordinating seat recorded under a different name.
- Making a genuine resolution failure actionable instead of silent.
- Saying what the coordinating role must be called, where the name is chosen.
- Correcting the docs chapter that recommends the watchdog while using a name the watchdog rejects.

## Out Of Scope

- `stalled-work` detection.
- Renaming anyone's recorded roles, or requiring a migration.
- Reworking role topology as a concept.
- The reporter's alias workaround as a shipped mechanism.

Self-contained. The verbatim failure, the three code sites re-verified on current main, the 85-minute measured stall, the backward-compatibility dimension, and the confirmed workaround are all above.

## Acceptance Criteria

- [ ] A team whose coordinating seat is recorded under a name other than `orchestration` gets a working heartbeat. **Demonstrated on a topology recorded the way the reporter's was** — not on a fixture already using the expected name.
- [ ] The failure stops being silent. When the seat genuinely cannot be resolved, the output names what is missing and the act that clears it. A reader must be able to act without reading source.
- [ ] The alias mapping is applied in **one** place every consumer reaches, not copied next to the existing one. `GuideRoleContractGuidance` already holds it and the resolver ignores it; two copies would drift.
- [ ] The naming expectation stops being implicit — either the recorded name no longer has to match, or the surface that records it says what the coordinating role must be called at the moment of recording.
- [ ] Existing topologies are not invalidated and need no operator migration. A team recorded as `orchestrator` keeps everything that works today working.
- [ ] Verification is emitted output on a real topology: heartbeat before, and after, including the resolved destination.
- [ ] State explicitly whether any surface **other than** heartbeat resolves a role by a hardcoded literal. If others do, say whether they are in scope; if none do, say that too, so the next reader knows the sweep happened.

## Verification

- Paste heartbeat output on an `orchestrator`-named team before and after.
- Paste the failure text for a topology that genuinely has no coordinating seat.
- Paste the rendered text that states the naming expectation.

## Related Links

- Same class, prior units: G714 (a duty assigned to a role that could not perform it), G719 (a diagnostic that named the wrong act)

## Knowledge Maintenance

Node 08 records that a safety net which cannot address its recipient is not a degraded safety net but an absent one — and that it fails in the shape hardest to notice, because detection keeps working while delivery is dead.

## Base Branch Policy

Do not fix this by telling operators to rename their roles. The reporter's topology was generated by our own commands and works everywhere else; requiring a rename would move our defect onto every existing install.

Judge the failure message by what a reader would **do**. "does not contain logical role 'orchestration'" is true and leaves them nowhere — it does not say that the name is required, that theirs differs, or what to change. That is the same shape as G719, and this unit should not reproduce it while fixing its sibling.

The docs chapter recommending this watchdog uses "orchestrator" throughout while the watchdog rejects that name. Whichever way the code goes, the chapter and the code must agree afterwards.

tomohisa:
Closes #1569

## Summary

- Centralized recorded-role resolution in the topology store, delegating alias normalization to the existing `GuideRoleContractGuidance.Normalize` source.
- Updated heartbeat, notify delivery, pending delegation metadata, inline payload warnings, and task-envelope selection to use the shared resolver.
- Added explicit heartbeat evidence for the recorded role and delivery destination, plus actionable remediation when the coordinating seat is genuinely missing.
- Documented the `orchestration` / `orchestrator` contract in mirrored English and Japanese guidance.

## Emitted evidence

- Baseline CLI against a topology recorded with `roles.orchestrator`: `cannot-determine`; it reported that logical role `orchestration` was absent.
- Branch build against that same topology: `actionable-stall`, `target_role=orchestration`, `resolved_recorded_role=orchestrator`, `resolved_destination=pane:wG723:p1`, and canonical notify command `--to orchestration`.
- Branch build against the same topology with the coordinating seat removed: `cannot-determine`; both `reason` and `suggested_action` name the roster, `session-layer topology record --role <orchestration|orchestrator> ... --write`, the no-rename rule, and rerunning heartbeat/notify.

## Verification

- `dotnet test ... --filter FullyQualifiedName~AutomationHeartbeatDecisionG597Tests --no-restore` — 11 passed.
- Focused affected command suites — 41 passed.
- Full repository test project with an isolated writable NuGet cache — 5,148 passed, 1 skipped, 0 failed.
- Role-consumer sweep: no remaining direct topology role lookup outside the shared resolver (the remaining structural comparison is not a consumer); the only alias mapping remains `GuideRoleContractGuidance.Normalize`.

### DIFF
--- a/docs/en/12-agent-message-orchestration.md
+++ b/docs/en/12-agent-message-orchestration.md
@@ -3139,6 +3139,8 @@ visible on the operator's screen the moment it breaks.
   only an `actionable-stall` verdict with its returned canonical notify command
   ever produces a sent nudge. Do not use `stale` or `message_body` to infer the
   action; the closed verdict is the sole decision trigger.
+- **Recorded coordinating-seat name** — `session-layer topology record --role <name>` accepts operator-supplied logical names. Heartbeat treats `orchestration` as the canonical coordinating role and `orchestrator` as its accepted recorded alias through the shared role-contract normalization used by topology delivery. A topology recorded as `orchestrator` therefore works without renaming or migration; when recording a new or corrected coordinating seat, choose either spelling and keep it in the team record. If no coordinating seat is genuinely present, `cannot-determine` names the source, missing role, recorded roster, and the `topology record --write` action that clears it.
+- **Role-consumer sweep** — `notify` delivery, pending-recipient metadata, task-envelope selection, recipient-delivery judgment, topology show, and heartbeat all use the same resolver. Semantic actor labels such as `action_owner: orchestration` are not topology lookups and remain canonical; no second alias table is maintained.
 - **Failure visibility** — silence is reserved for `healthy-active-wait`
   ONLY. A heartbeat command execution failure or
   malformed/non-object output must be surfaced **visibly** in the watchdog's
--- a/docs/ja/12-agent-message-orchestration.md
+++ b/docs/ja/12-agent-message-orchestration.md
@@ -2731,6 +2731,8 @@ credential/keychain セットアップも不要で(セッションの他の部�
   nudge を捏造・送信することは決してありません。送信する nudge は
   `actionable-stall` verdict とともに返された canonical notify command の場合だけです。
   `stale` や `message_body` から action を推論せず、closed verdict だけを decision trigger にします。
+- **記録する coordinating seat の名前** — `session-layer topology record --role <name>` は operator が logical role 名を指定する surface です。heartbeat は `orchestration` を canonical な coordinating role とし、topology delivery と共有する role-contract normalization によって `orchestrator` を記録済み alias として受け入れます。したがって `orchestrator` で記録済みの topology を rename や migration なしで使えます。新規または修正の coordinating seat を記録するときは、どちらかの spelling を team record に残してください。本当に seat がない場合の `cannot-determine` は source、missing role、記録済み roster、それを解消する `topology record --write` action を出力します。
+- **role consumer の sweep** — `notify` delivery、pending recipient metadata、task-envelope 選択、recipient-delivery judgment、topology show、heartbeat はすべて同じ resolver を使います。`action_owner: orchestration` のような semantic actor label は topology lookup ではないため canonical のままです。alias table を別に持ちません。
 - **failure visibility** — 沈黙は `healthy-active-wait` の heartbeat 結果にのみ
   許されます。heartbeat コマンドの実行失敗や不正な/オブジェクトでない出力は、
   この wake の watchdog 自身の turn 出力で **可視的に** 表面化させなければ
--- a/src/IntentSystem.Cli/Commands/AutomationHeartbeatCommand.cs
+++ b/src/IntentSystem.Cli/Commands/AutomationHeartbeatCommand.cs
@@ -110,6 +110,8 @@ internal static class AutomationHeartbeatCommand
             DedupeKey = decision.DedupeKey,
             ActionOwner = decision.ActionOwner,
             TargetRole = decision.TargetRole,
+            ResolvedRecordedRole = decision.ResolvedRecordedRole,
+            ResolvedDestination = decision.ResolvedDestination,
             CanonicalNotifyCommand = decision.CanonicalNotifyCommand,
             WaitCondition = decision.WaitCondition,
             WaitEndSignal = decision.WaitEndSignal,
@@ -368,7 +370,11 @@ internal static class AutomationHeartbeatCommand
         var route = ResolveOrchestratorRoute(context, domain, team);
         if (!route.Resolved)
         {
-            return CannotDetermine(route.Reason!, stalledWork.Items.FirstOrDefault(), "recorded topology");
+            return CannotDetermine(
+                route.Reason!,
+                stalledWork.Items.FirstOrDefault(),
+                "recorded topology",
+                route.Reason);
         }
 
         // A concrete actionable finding always wins over a legitimate active
@@ -400,7 +406,9 @@ internal static class AutomationHeartbeatCommand
                 WaitCondition: $"CI for PR #{ciPending.Pr?.Number} head {ciPending.PrHeadSha} to reach a terminal outcome",
                 WaitEndSignal: "the mode-specific CI-completion wake followed by an exact-head GitHub re-check",
                 WaitBoundMinutes: staleMinutes,
-                SuggestedAction: "Wait for the named CI completion signal, then re-evaluate this heartbeat decision.");
+                SuggestedAction: "Wait for the named CI completion signal, then re-evaluate this heartbeat decision.",
+                ResolvedRecordedRole: route.RecordedRole,
+                ResolvedDestination: route.Destination);
         }
 
         if (ciPending is not null)
@@ -430,7 +438,9 @@ internal static class AutomationHeartbeatCommand
             WaitCondition: null,
             WaitEndSignal: null,
             WaitBoundMinutes: null,
-            SuggestedAction: "Run the canonical notify command once for this no-pending-transition dedupe key.");
+            SuggestedAction: "Run the canonical notify command once for this no-pending-transition dedupe key.",
+            ResolvedRecordedRole: route.RecordedRole,
+            ResolvedDestination: route.Destination);
     }
 
     private static HeartbeatDecision AttentionRequired(StalledWorkItem record, string owner, HeartbeatRoute? route)
@@ -453,7 +463,9 @@ internal static class AutomationHeartbeatCommand
             WaitCondition: null,
             WaitEndSignal: null,
             WaitBoundMinutes: null,
-            SuggestedAction: action);
+            SuggestedAction: action,
+            ResolvedRecordedRole: route?.RecordedRole,
+            ResolvedDestination: route?.Destination);
     }
 
     private static HeartbeatDecision ActionableStall(
@@ -479,10 +491,16 @@ internal static class AutomationHeartbeatCommand
             WaitCondition: null,
             WaitEndSignal: null,
             WaitBoundMinutes: null,
-            SuggestedAction: $"Run the canonical notify command once for dedupe key '{dedupeKey}'.");
+            SuggestedAction: $"Run the canonical notify command once for dedupe key '{dedupeKey}'.",
+            ResolvedRecordedRole: route.RecordedRole,
+            ResolvedDestination: route.Destination);
     }
 
-    private static HeartbeatDecision CannotDetermine(string reason, StalledWorkItem? item, string source) => new(
+    private static HeartbeatDecision CannotDetermine(
+        string reason,
+        StalledWorkItem? item,
+        string source,
+        string? suggestedAction = null) => new(
         Verdict: "cannot-determine",
         Reason: reason,
         LastProgressBasis: item is null ? source : ProgressBasis(item),
@@ -495,7 +513,8 @@ internal static class AutomationHeartbeatCommand
         WaitCondition: null,
         WaitEndSignal: null,
         WaitBoundMinutes: null,
-        SuggestedAction: $"Repair or inspect the named {source} failure before treating this heartbeat as healthy.");
+        SuggestedAction: suggestedAction
+            ?? $"Repair or inspect the named {source} failure before treating this heartbeat as healthy.");
 
     private static HeartbeatRoute ResolveOrchestratorRoute(
         CliContext context,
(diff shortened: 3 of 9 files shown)


tomohisa (commented):
# Semantic review — G723 / PR #1571 (head `93ce59aa`)

**Verdict: approve-equivalent (no implementation findings).** Submitted as `COMMENTED` because `gh pr review --approve` is rejected on a same-account PR (G696). The PR is draft; this review does not merge, approve, or promote it — draft resolution belongs to the host closeout path.

Execution unit **G723**, packet `.intent-cli/issues/G723` (`packet.yaml` / `implementation.md` / `review-context.md` / `github-body.md`), closeout-plan `ready:true`, linked issue **#1569**, closing reference `Closes #1569` present and matching (G311).

I did not take the PR body's evidence on trust. I built the CLI at PR head in an isolated worktree and re-ran the reported scenario end-to-end against a topology I recorded as `orchestrator`. The pasted output below is from that run, not from the PR description.

---

## AC-by-AC mapping

### AC1 — a team recorded under another name gets a working heartbeat, demonstrated on a topology recorded the way the reporter's was

Verified on a topology whose only coordinating seat is `roles.orchestrator` (never `orchestration`).

**Before** — installed released `intent-cli 0.23.1-d49984d-G721`, same fixture:

```
"verdict": "cannot-determine",
"reason": "Recorded routing for team 'g723demo' is unresolved: Resolved herdr logical role 'design'
 to recorded pane 'wG723:p3' in workspace 'wG723'. Recorded role topology '.../g723demo.json'
 does not contain logical role 'orchestration'.",
"action_owner": "monitor",
"suggested_action": "Repair or inspect the named recorded topology failure before treating this heartbeat as healthy."
```

That reproduces the reported defect exactly: detection ran, addressing was dead, no canonical notify command was produced.

**After** — build of this PR head (`93ce59aa`), identical fixture:

```
"verdict": "actionable-stall",
"action_owner": "orchestration",
"target_role": "orchestration",
"resolved_recorded_role": "orchestrator",
"resolved_destination": "pane:wG723:p1",
"canonical_notify_command": "intent-cli notify report --domain intent-cli --team g723demo --from design
 --to orchestration --task-id <heartbeat-dedupe-key> --status question ... --write --format json"
```

The failing case is the one that was demonstrated. `AutomationHeartbeatDecisionG597Tests.Execute_RecordedOrchestratorAliasResolvesAndEmitsDestination_G723` pins the same shape at unit level, and its fixture writes `roles.orchestrator` — `WriteTopology(string? coordinatingRole)` was parameterised precisely so the test cannot silently fall back to the name that always worked.

### AC2 — the failure stops being silent

For a topology that genuinely has no coordinating seat, the emitted text (PR head build) is:

```
Recorded role topology '.../g723demo.json' for team 'g723demo' workspace 'wG723' does not contain
logical role 'orchestration' (found in that team scope: design). Record that role for this team
before retrying notify. The coordinating seat is accepted as canonical 'orchestration' or accepted
recorded alias 'orchestrator'; an existing record under either name does not need renaming.
Record it with `intent-cli session-layer topology record --domain intent-cli --team g723demo
--role <orchestration|orchestrator> ... --write`; do not rename an existing accepted alias.
Then rerun the heartbeat/notify command.
```

Judged by the standard `implementation.md` set — what would a reader *do* — this names the source file, the missing logical role, the actual roster, the accepted spellings, the exact recording command with the team already substituted, the explicit no-rename rule, and the rerun. It does not reproduce the G719 shape it was warned against.

### AC3 — one alias table, not two

`GuideRoleContractGuidance.Normalize` remains the sole mapping. `NotifyRoleTopologyStore.ResolveRecordedRole` (`NotifyRoleTopology.cs:466`) calls into it rather than restating it, and every consumer reaches that one function: `NotifyTransport.cs:322,331`, `NotifyCommand.cs:1776` (delivery) and `:1999` (inline payload warning), `NotifyTaskEnvelopeStore.cs:92`, `NotifyRecipientDeliveryJudgment.cs:40` and `SessionLayerTopologyCommand.cs:235` via `ResolveDeliveryTarget`, and heartbeat via the same. No second alias table was introduced anywhere in the diff.

Also good: `NotifyTransport`'s private `FormatRoles` was deleted rather than left beside the new central one — the duplicate-drift risk was removed on both axes, not just the alias map.

### AC4 — the naming expectation stops being implicit

Both directions are covered. Resolution no longer requires a match, *and* the surface where the name is chosen now says it. Verified from the PR-head build:

```
$ intent-cli session-layer topology record
Usage: ... --role <name> ...
   Heartbeat coordination accepts recorded role 'orchestration' or the alias 'orchestrator';
   existing records under either name need no rename or migration.

$ intent-cli session-layer topology record --domain intent-cli --team g723fresh --role orchestrator ... --dry-run
"summary": "Dry-run: would record operator-supplied role 'orchestrator' for team 'g723fresh'.
 Heartbeat coordination accepts recorded role 'orchestration' or the alias 'orchestrator';
 existing records under either name need no rename or migration."
```

### AC5 — existing topologies are not invalidated, no operator migration

Verified on the three topology shapes that actually exist in the field:

| recorded roster | heartbeat at PR head |
|---|---|
| `orchestration` only (the case that always worked) | unchanged — covered green by the existing G597 suite |
| `orchestrator` only (the reporter's) | `actionable-stall`, `resolved_recorded_role=orchestrator` |
| **both** — i.e. a team that already applied the reporter's alias workaround | `actionable-stall`, `resolved_recorded_role=orchestration` |

The third row matters and I checked it deliberately: teams that took the published workaround now have two entries that both normalize to `orchestration`. The exact-match fast path fires first, so they resolve cleanly instead of tripping the new `ambiguous-role` branch. Nobody has to undo the workaround, and nobody has to rename anything. The defect was not pushed onto installs.

### AC6 — verification from emitted output on a real topology

Satisfied. The before/after and the genuine-failure text above are emitted output from a real CLI run on a topology recorded as `orchestrator`, reproduced independently at this head. The PR description reported these values but did not paste the output; pasting it here puts the durable emitted evidence on the PR itself.

### AC7 — the hardcoded-role sweep, negatives included

I re-ran the sweep rather than accepting the summary. At `93ce59aa` exactly two direct `topology.Roles` lookups remain in `src/`:

- `NotifyRoleTopology.cs:477` — the exact-match fast path *inside* `ResolveRecordedRole` itself. Correct by construction.
- `NotifyRoleTopology.cs:1062` — `Equivalent(...)`, a structural topology-to-topology comparison. Not a role consumer, so alias normalization would be wrong there.

The remaining `"orchestration"` string literals are semantic actor labels and canonical constants, not topology lookups: `action_owner` / `ActionOwner`, `AutomationStalledWorkCommand.RequiredActor`, `BranchLaneDecisionRecord`/`Command` actor checks, `GuideBootstrapCommand.ExpectedRoles`, `GuideWorkspaceLayoutCommand` labels, `HerdrStandardLayoutRegistry`, `PromptAdjudication.OrchestrationRole`, `CloseoutRecordRole.Orchestration`. Heartbeat's own `ResolveDeliveryTarget(..., "orchestration")` at `:531`/`:561` is now the *canonical logical* role handed to the shared resolver — that is the intended shape, not a residual hardcode. The PR states this sweep including the surfaces that were already correct, which is what the packet asked for.

---

## Out-of-scope boundaries — not crossed

- `stalled-work` detection is untouched; `AutomationStalledWorkCommand.cs` is not in the diff.
- No rename or migration is required or suggested anywhere in the change.
- Role topology as a concept is not reworked.
- The reporter's alias workaround is not shipped as the mechanism; it is only acknowledged as a diagnosis.

## Design alignment — `design_alignment_checked=true`

Consistent with the primary intent reference `intents/intent-cli/intent-tree/means/08-agent-message-orchestration.md`, which holds that a role's delivery is determined solely by how that role is recorded in the topology, and that a herdr recipient's identity is the recorded **workspace + pane**, not an agent name, with `--to` taking a *logical* role. This change keeps `--to` logical, resolves the recorded spelling centrally, and now emits both halves (`target_role` logical, `resolved_recorded_role` recorded, `resolved_destination` workspace+pane) — that is the node's model made observable rather than an exception to it. Supporting reference `means/04-worker-interface-strategy.md` owns the watchdog contract that was silently disabled; it is restored, not widened.

## Tests

- Focused suites at PR head, in-tree: **223 passed / 0 failed** (`AutomationHeartbeat*`, `Notify*`, `SessionLayerTopology*`, `RoleTopology*`).
- CI at `93ce59aa`: `Build and test (source contract)` pass, `npm package dry-run (no publish)` pass.
- `git diff --check` against the merge result: clean.
- Head SHA confirmed `93ce59aa8a9e5a52be394746cb43e7be361827f0` before and after this pass.

Passing tests are treated as **necessary, not sufficient**. The approval rests on the packet-contract and emitted-output evidence above, not on the green run.

---

## Informational notes (no change requested, no follow-up required)

1. In the `cannot-determine` route, `suggested_action` is now byte-identical to `reason`, and the whole remedial paragraph is also embedded in `dedupe_key`. Behaviour is correct and the key stays stable across polls, but the two fields no longer carry distinct information and the dedupe key is now a multi-sentence string containing an absolute path. Worth a thought if that key is ever surfaced to an operator.
2. The centralized `NotifyRoleTopologyStore.FormatRoles` dropped the `"none"` empty-roster fallback that the deleted `NotifyTransport.FormatRoles` had. Unreachable today — the empty-roster guard fires first and emits `contains no roles` (I confirmed this on an empty-roles fixture) — but a future refactor that removes that guard would render `(found in that team scope: )`.

Neither affects an acceptance criterion.

<sub>Review evidence collected at PR head `93ce59aa` with a build of this branch; before-state captured with released `intent-cli 0.23.1-d49984d-G721`.</sub>

PR #1572, opened 2026-08-20
fix: isolate multi-domain session-layer worker identity

### LINKED ISSUE #1570: G724: the CLAUDE.md session-layer marker is a single slot, so on a multi-domain host one team's claim blocks another team's worker complete
# G724: the CLAUDE.md session-layer marker is a single slot, so on a multi-domain host one team's claim blocks another team's worker complete

Resolves the consumer report in #1568.

## Goal

On a host carrying several domains, each domain's session-layer marker coexists, and a worker whose domain does not currently own the marker can still complete its canonical closeout.

## Why This Slice Exists Now

`SekibanDcbTsHost` carries five domains together: `sekiban-dcb-ts`, `sekiban-as-a-service`, `sekiban-wasm-runtime`, `idd-com`, `intent-cli`. That is a real, supported operating shape — not an abuse.

The reporter's timeline:

1. **2026-08-08** — the `sekiban-as-a-service` team generated the marker (commit `dad745b5`). The CLAUDE.md block became that domain's claim.
2. The `sekiban-dcb-ts` team then closed out **twelve packets, G13 through G24, without incident** — the design-side `closeout pr` path does not check the marker.
3. **SDT-G25** — an implementation worker ran canonical `worker complete` and hit it for the first time:

   > canonical worker complete is blocked by host domain-identity conflict and manual workflow labels are prohibited.

4. The worker behaved correctly: it did not hand-edit labels, it reported blocked, and it stopped. Orchestration recovered via a PR-linkage recover path and closeout finished through the design-side `closeout pr`. **That recovery is not a documented escape route.** It worked, which is not the same as being sanctioned.
5. Separately, ordinary intent-cli invocations from the `sekiban-dcb-ts` side **silently rewrote the working-tree CLAUDE.md marker** to `sekiban-dcb-ts`. Committing that would have destroyed the other domain's claim. The reporter reverted it by hand. Left alone, two teams tug the marker back and forth on `main`.

The seventeen-day gap between step 1 and step 3 is the point. Nothing surfaced the collision until an unrelated command failed closed for a reason it did not name.

## Current Observed State

Verified on current `main`, after 0.23.1: `SessionLayerMarkerCommand.cs` declares one marker prefix, `intent-cli:session-layer-marker:`. **One slot per host, not one per domain.**

The durable record is already correct: `.intent-cli/session-layer-mode.json` is separated per domain and per team. Only the CLAUDE.md block — the display and claim surface — is single-slot.

- Multi-domain hosts are supported. Do not resolve this by declaring the shape unsupported.
- The worker was right to fail closed rather than hand-edit workflow labels. That behaviour stays.
- `.intent-cli/session-layer-mode.json` is already per domain and per team, and is not the problem.
- The reporter's PR-linkage recovery is an observation, not a sanctioned path. Do not enshrine it just because it worked.

## Target Repo / Path / Part

- Target paths: `src/IntentSystem.Cli/Commands/SessionLayerMarkerCommand.cs`, the `worker complete` domain-identity check, the multi-domain host documentation in both language mirrors, and their tests
- Part: how the managed session-layer marker is written, and which record `worker complete` checks domain identity against

## In Scope

- Making the marker coexist across domains on one host, or demoting it so identity is checked against the record that is already per-domain.
- Stopping ordinary commands from silently rewriting a marker they do not own.
- Naming the sanctioned recovery if domain-identity conflict can still legitimately block a worker.

## Out Of Scope

- Declaring multi-domain hosts unsupported.
- Removing the worker's fail-closed behaviour.
- Changing `.intent-cli/session-layer-mode.json`, which is already correct.
- Requiring operators to hand-edit CLAUDE.md.

Self-contained. The five-domain host, the dated timeline, the twelve packets that passed before the collision surfaced, the silent rewrite, and the separation between the durable record and the display block are all above. #1568 carries the reporter's original wording.

## Acceptance Criteria

- [ ] Each domain's marker coexists with the others on one host. **Demonstrated on a host with at least two domains recorded**, not on a single-domain fixture.
- [ ] A worker for a domain that does not currently own the marker can complete its canonical closeout. The reported blocked state stops occurring for that reason.
- [ ] An ordinary command never silently rewrites another domain's claim in the working tree. If a command would change a marker it does not own, it either does not, or it says so **at the time**, in terms an operator notices before committing.
- [ ] If domain-identity conflict can still legitimately block a worker, the sanctioned recovery is written down and reachable **from the worker's own surface**. A path that happens to work is not an escape route until it is documented.
- [ ] Existing single-domain hosts are unaffected and need no migration — byte-identical where nothing changed.
- [ ] Existing multi-domain hosts already carrying one domain's marker are handled **without** asking the operator to hand-edit CLAUDE.md. State what happens to the marker already there.
- [ ] The reported scenario is walked end to end on a two-domain host: domain A generates the marker, domain B runs `worker complete`, with what each step emits shown.

## Verification

- Paste the CLAUDE.md marker region on a two-domain host after both have been recorded.
- Paste `worker complete` output for the domain that does not own the marker, before and after.
- Show that a command run from domain B leaves domain A's claim untouched, or announces the change.

## Related Links

- Consumer report: https://github.com/J-Tech-Japan/intent-system/issues/1568
- Same host shape, different collision: #1555, resolved as G717

## Knowledge Maintenance

Node 08 records that a single-slot claim written into a shared file assumes one occupant — on a host that genuinely carries several domains the last writer wins, and the loss is silent until an unrelated command fails closed for a reason it does not name.

An ADR is expected for which record is authoritative for host session-layer domain identity, because the alternatives trade differently and a later reader will otherwise relitigate it.

## Base Branch Policy

Decide first which record is authoritative — the CLAUDE.md block or `.intent-cli/session-layer-mode.json`. The rest follows from that answer, and doing the mechanical parts first will produce a marker that coexists while the identity check still consults the wrong thing.

Note the shape this shares with G717: two representations of one fact, no stated precedence, and the collision only visible when something unrelated fails closed. If the answer here is "the durable per-domain record wins and the block is display", say it in those words so the next reader does not have to infer it from behaviour.

Do not enshrine the reporter's PR-linkage recovery just because it worked. Judge it on whether it is the right sanctioned path; if it is, document it as one, and if it is not, provide the path that is.

tomohisa:
Closes #1570

## Decision before mechanics

Added [ADR 0009](https://github.com/J-Tech-Japan/intent-system/blob/claude/g724-session-identity-authority/docs/adr/0009-session-layer-domain-identity-authority.md) before the marker or worker changes. The authoritative host session-layer identity/mode is the domain/team-scoped `.intent-cli/session-layer-mode.json` record; durable queue/packet domain metadata is authoritative for an execution unit. The generated `AGENTS.md`/`CLAUDE.md` block is display/verification evidence only and never selects a worker domain. The ADR also defines additive migration, legacy domain-less recovery, and explicitly rejects PR-linkage recovery as a domain identity workaround.

## Implementation

- Marker generation appends a missing recorded domain/team block when another valid managed block exists, reports `marker_action: appended`, and preserves existing blocks byte-for-byte. A truly unmanaged first marker still requires the documented placeholder; no hand edit is requested for an existing multi-domain marker.
- Host `worker complete` resolves the execution-unit domain from the durable queue/packet record, validates explicit `--domain`, uses the authoritative session-layer record for legacy domain-less migration, and fails closed with an exact worker-surface re-invocation for missing, ambiguous, contradictory, or unreadable evidence.
- `linked_pr` projection follows the resolved execution unit, preventing colliding issue numbers from selecting another domain's queue row.
- Child `--github-only` remains metadata-free.
- Updated the English/Japanese getting-started, implementation-loop, recovery, and orchestration documentation.

## Acceptance evidence

The focused G724 test uses two recorded domains, generates A first, then appends B:

```text
marker_generate_domain_b: marker_action=appended, preserved_existing_blocks=1
marker_after_b:
<!-- intent-cli:session-layer-marker:start domain="domain-a" team="alpha" -->
<!-- ... mode="agmsg" ... -->
<!-- intent-cli:session-layer-marker:end -->
<!-- intent-cli:session-layer-marker:start domain="domain-b" team="alpha" -->
<!-- ... mode="herdr-only" ... -->
<!-- intent-cli:session-layer-marker:end -->
```

The end-to-end host test then runs domain B while A owns the visible marker:

```text
marker_before:
<!-- intent-cli:session-layer-marker:start domain="intent-cli" team="alpha" -->
<!-- ... mode="agmsg" ... -->
<!-- intent-cli:session-layer-marker:end -->

worker_complete_after:
"proceed": true,
"applied": true,
"linked_pr_synced": true,
"domain": "sekiban-as-a-service",
"domain_source": "queue-record",
"execution_unit": "G724"

marker_after: byte-identical to marker_before; no sekiban-as-a-service block was written by worker complete.
```

For a genuine missing domain on a two-domain host, the worker stops before labels or queue writes and emits the worker recovery, including `--domain intent-cli` (and the alternate recorded domain). The output says the startup marker and PR-linkage recovery cannot supply the missing durable identity.

## Verification

- Focused CLI tests: `SessionLayerMarkerG601Tests`, `WorkerCompleteCommandTests`, and `GuideWorkerIssueToPrCommandTests` — passing.
- Full solution: `5154 passed, 1 skipped` under SDK `10.0.100`; packaged invocation smoke is `11 passed`. The sandbox required a temporary local NuGet audit override and writable CLI/NuGet homes for nested pack/tool execution; no verification override is committed.
- No manual GitHub label mutation or PR linkage recovery was used. The issue-side lifecycle transition was performed through `intent-cli worker claim`; completion will be performed through the canonical worker command after PR creation.

### DIFF
--- a/docs/adr/0009-session-layer-domain-identity-authority.md
+++ b/docs/adr/0009-session-layer-domain-identity-authority.md
@@ -21,7 +21,10 @@ causes silent rewrites and can make a worker for another domain fail closed.
    specific than a domain-wide entry, exactly as the existing mode resolver
    defines. Durable execution-unit or packet domain metadata is authoritative
    for the execution unit itself. An explicit worker `--domain` is accepted
-   only when it agrees with that durable domain metadata.
+   only when it agrees with the durable queue/packet domain or, for a legacy
+   domain-less queue row, with the authoritative session-layer record. It is
+   a migration binding for that legacy shape, never an override of declared
+   durable data.
 2. The generated startup-file marker is display and verification evidence
    only. It carries the domain, team, mode, and record hash so an operator can
    see which durable record a block represents, but it never establishes
--- a/docs/adr/0009-session-layer-domain-identity-authority.md
+++ b/docs/adr/0009-session-layer-domain-identity-authority.md
@@ -1,0 +1,70 @@
+# ADR 0009: Durable session-layer records own domain identity
+
+- Status: Accepted (preview-through-1.x)
+- Date: 2026-08-20
+- Deciders: Operator and implementation seat, recorded by G724
+- Related: G601 session-layer markers, G717 claim authority, G724 / #1570
+
+## Context
+
+The host carries more than one supported domain. The durable
+`.intent-cli/session-layer-mode.json` record is already scoped by domain and,
+when applicable, team. The generated `AGENTS.md` / `CLAUDE.md` marker is a
+single working-tree display surface, so treating that block as the host's
+domain identity makes the last marker writer appear to own every domain. That
+causes silent rewrites and can make a worker for another domain fail closed.
+
+## Decision
+
+1. `.intent-cli/session-layer-mode.json` is authoritative for host
+   session-layer domain identity and mode. A team-scoped entry is more
+   specific than a domain-wide entry, exactly as the existing mode resolver
+   defines. Durable execution-unit or packet domain metadata is authoritative
+   for the execution unit itself. An explicit worker `--domain` is accepted
+   only when it agrees with that durable domain metadata.
+2. The generated startup-file marker is display and verification evidence
+   only. It carries the domain, team, mode, and record hash so an operator can
+   see which durable record a block represents, but it never establishes
+   identity, selects a worker domain, or overrides the durable record. Marker
+   precedence is therefore below the durable record: a marker for domain A
+   does not block or rewrite domain B.
+3. Marker generation is domain-scoped and additive. It replaces only the
+   requested domain/team block and preserves every other block byte-for-byte.
+   When a recorded domain/team has no block, the canonical writer appends its
+   block under `--write`; an operator must not hand-edit `CLAUDE.md` or
+   `AGENTS.md`. A command that cannot prove it will preserve another domain's
+   block refuses with an explanation before writing.
+4. Worker completion resolves the execution-unit domain from the durable
+   queue/packet record, with an explicit matching `--domain` available from
+   the worker surface. It never reads the startup marker as an identity claim.
+   If durable domain evidence is absent or contradictory, the worker remains
+   fail-closed and emits the exact domain-scoped re-invocation needed to
+   repair or select the durable record. This is the sanctioned recovery and
+   preserves queue linkage; the reporter's accidental PR-linkage recovery is
+   not a sanctioned substitute.
+
+## Migration and compatibility
+
+- A valid existing single-domain marker remains valid and is unchanged when
+  its own domain is regenerated; no migration is required and the unchanged
+  path remains byte-identical.
+- A multi-domain host carrying only domain A's existing marker is not asked to
+  edit the file. The next canonical `marker generate --domain B ... --write`
+  appends B's generated block while retaining A. Existing A content is never
+  silently replaced by B.
+- Existing durable mode records are read in their current schema and remain
+  the source of truth. No conversion of `.intent-cli/session-layer-mode.json`
+  is required.
+
+## Consequences
+
+- Multiple domain/team markers can coexist in one startup file without a
+  last-writer-wins claim.
+- A domain-B worker can complete while domain A is the visible marker owner,
+  because worker identity follows durable execution-unit evidence rather than
+  the display file.
+- Stale or missing display evidence is visible and repairable, but cannot
+  silently change routing or ownership.
+- A genuine durable-domain contradiction still stops the worker; the worker
+  itself names the sanctioned explicit-domain recovery instead of inviting a
+  manual label edit or an undocumented PR-linkage workaround.
--- a/docs/en/02a-getting-started-orchestration.md
+++ b/docs/en/02a-getting-started-orchestration.md
@@ -136,14 +136,15 @@ in [doc 12](12-agent-message-orchestration.md).
 
 ### 2.3 Generate the visible marker
 
-Before generating, add the one empty managed marker block for this domain/team
-to `AGENTS.md` or `CLAUDE.md` exactly as [doc 12's generated-marker
-section](12-agent-message-orchestration.md#visible-generated-mode-markers)
-specifies. Then paste this prompt:
+If `AGENTS.md` or `CLAUDE.md` already has another valid managed marker block,
+do not edit it or add a placeholder by hand. The canonical writer will append
+the missing `(domain, team)` block. Only a startup file with no managed block
+at all still needs the one empty placeholder described by [doc 12's
+generated-marker section](12-agent-message-orchestration.md#visible-generated-mode-markers).
+Then paste this prompt:
 
 > Generate the marker for the recorded domain `<domain>` and team `<team>`.
-> Run `intent-cli session-layer marker generate --domain <domain> --team <team> --file AGENTS.md --write --format json`, then show the JSON. Change only the
-> managed marker block.
+> Run `intent-cli session-layer marker generate --domain <domain> --team <team> --file AGENTS.md --write --format json`, then show the JSON. The command may append this domain's block when another managed block exists; it must preserve every other block.
 
 The scratch-host run returned this success shape (the record hash is dynamic):
 
@@ -155,7 +156,9 @@ The scratch-host run returned this success shape (the record hash is dynamic):
   "team": "docs-team",
   "mode": "herdr-only",
   "verify_command": "intent-cli session-layer show --domain onboarding --team docs-team",
-  "summary": "Generated the managed session-layer marker for team 'docs-team' in 'AGENTS.md'."
+  "marker_action": "appended",
+  "preserved_existing_blocks": 1,
+  "summary": "Appended the managed session-layer marker for domain 'onboarding' and team 'docs-team' in 'AGENTS.md'; preserved 1 existing managed block(s)."
 }
 ```
 
--- a/docs/en/05-implementation-loop.md
+++ b/docs/en/05-implementation-loop.md
@@ -58,6 +58,25 @@ copy a long loop body from this document.
 - A child agent never applies `intent-target` (host-owned) or `intent-pr-created` (issue-side marker) to a PR
 - `linked_pr_synced: false` from `worker complete` is the expected child-cwd warning — record it and move on
 
+## G724: worker domain identity on a multi-domain host
+
+The startup marker is display evidence, not a worker binding. In host context,
+`worker complete --kind issue --outcome pr-created` resolves the execution-unit
+domain from the durable queue/packet record. A domain-B worker therefore remains
+eligible even when the shared `CLAUDE.md` currently displays domain A. The JSON
+result reports the selected `domain`, `domain_source` (normally
+`queue-record`), and `execution_unit`; worker completion does not rewrite the
+marker.
+
+If a host invocation supplies `--domain`, it must match the durable queue
+domain, or the authoritative session-layer record for a legacy domain-less
+queue row. Missing, contradictory, unreadable, or ambiguous durable identity
+is a fail-closed result with an exact re-invocation using `--domain <name>`.
+Use that worker-surface recovery after repairing/selecting the canonical queue
+or packet record. Do not hand-edit the marker, apply labels manually, or use
+PR-linkage recovery as a domain-identity workaround. Child `--github-only`
+remains metadata-free and does not read host queue state.
+
 ## Preview: Git-backed cross-clone scope claims (G679)
 
 The decision and its boundaries are recorded in
@@ -147,6 +166,11 @@ intent-cli worker claim --kind issue --number <n> --repo <owner>/<repo> --github
 intent-cli worker result-summary --kind issue-to-pr --repo <owner>/<repo> --issue <n> --pr <pr> --outcome <outcome> --format json
 intent-cli worker complete --kind issue --number <n> --repo <owner>/<repo> --github-only --outcome <outcome> --pr <pr> --write --format json
 ```
+
+For a host-side completion that needs explicit domain selection, add
+`--domain <durable-domain>` to the same `worker complete` command. It must
+agree with the queue/packet record; the startup marker never supplies or
+overrides it.
 
 ## Next
 
--- a/docs/en/07-recovery.md
+++ b/docs/en/07-recovery.md
@@ -89,20 +89,26 @@ you whether a child-loop-owned repair exists. Host-owned categories surface as
 > `--issue <n>`. Child `--github-only` loops never write `linked_pr`; this recovery
 > belongs to the host closeout surface.
 
-> **Repository-qualified linkage and completed repairs (G603).** A GitHub
+> **Repository-qualified linkage and completed repairs (G603/G724).** A GitHub
 > issue or PR number is an identity only together with `owner/repo`. Worker
 > completion, closeout, review planning, and stalled-work therefore reject a
 > bare or wrong-repository linkage rather than letting a colliding number pick
-> a foreign queue item. In host context, `worker complete` also refuses a
-> queue write when the selected unit's declared domain disagrees with the
-> command domain. For a completed unit whose `linked_pr` points at the wrong
-> repository, use `automation host-queue-item-recovery --repo <owner/repo>
-> --unit <unit> --issue <n> --pr <n> --write`: it requires GitHub evidence
-> that the proposed PR exists, is merged, and closes that unit's own issue,
-> then appends a `completed-linkage-repair` run event containing the evidence.
-> A legacy packet without a readable `publish.yaml` identity stops with
-> `legacy-publish-identity-missing` and names the missing evidence; it never
-> guesses or changes a completed linkage.
+> a foreign queue item. In host context, `worker complete` resolves the
+> execution-unit domain from the durable queue/packet record; the visible
+> `AGENTS.md`/`CLAUDE.md` marker and host default domain are display/default
+> context only. A supplied `--domain` must agree with the durable domain. If
+> the durable identity is missing, contradictory, unreadable, or ambiguous,
+> the worker fails closed and prints the exact `worker complete ... --domain
+> <name>` re-invocation after the canonical queue/packet record is repaired.
+> This worker-surface recovery is the sanctioned domain recovery; PR-linkage
+> recovery is not a substitute. For a completed unit whose `linked_pr` points
+> at the wrong repository, use `automation host-queue-item-recovery --repo
+> <owner/repo> --unit <unit> --issue <n> --pr <n> --write`: it requires GitHub
+> evidence that the proposed PR exists, is merged, and closes that unit's own
+> issue, then appends a `completed-linkage-repair` run event containing the
+> evidence. A legacy packet without a readable `publish.yaml` identity stops
+> with `legacy-publish-identity-missing` and names the missing evidence; it
+> never guesses or changes a completed linkage.
 
 ### Repeated-stall recovery (G408)
 
--- a/docs/en/12-agent-message-orchestration.md
+++ b/docs/en/12-agent-message-orchestration.md
@@ -1640,10 +1640,11 @@ shape still refuses its conflict and has no force flag.
 
 #### Visible, generated mode markers
 
-The mode record remains the only source of truth, but a team should not have to
-remember to query it before noticing which transport it uses. Put one explicit
-empty managed block for each `(domain, team)` in an agent-startup file
-(`AGENTS.md` or `CLAUDE.md`), then generate its display from the recorded mode:
+The mode record remains the only source of truth for host session-layer domain
+identity and mode. A team should not have to remember to query it before
+noticing which transport it uses. Put one explicit empty managed block for each
+`(domain, team)` in an agent-startup file (`AGENTS.md` or `CLAUDE.md`), then
+generate its display from the recorded mode:
 
 ```text
 <!-- intent-cli:session-layer-marker:start domain="<domain>" team="<team>" -->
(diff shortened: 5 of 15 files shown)


tomohisa (commented):
# Semantic review — G724 / PR #1572 (head `1ad6d6b6`)

**Verdict: request-update — one implementation finding.** Everything else in this unit is verified and passing; the scope of the requested change is a single branch of `ResolveWorkerDomain`. Submitted as `COMMENTED` because `gh pr review --approve` is rejected on a same-account PR (G696). The PR is draft; this review does not merge, approve, or promote it.

Execution unit **G724**, packet `.intent-cli/issues/G724`, closeout-plan `ready:true`, linked issue **#1570**, `Closes #1570` present and matching (G311).

I did not take the PR body's evidence on trust. I built the CLI at this head and walked the reported two-domain scenario against a host with two genuinely recorded domains, with released `0.23.1-d49984d-G721` as the before-state. Every transcript below is from that run.

---

## The finding

**Category: implementation-finding. Fails AC4 (sanctioned recovery reachable from the worker's own surface), and contradicts this PR's own ADR 0009 decision #1.**

`WorkerCompleteCommand.ResolveWorkerDomain` — the branch where exactly one **legacy domain-less** queue row matches and the host records **more than one** session-layer domain — builds its suggested re-invocation from `context.Config.Project.Domain`, the host default:

```csharp
var suggestedDomain = context.Config.Project.Domain;
return WorkerDomainResolution.Error(
    $"refused worker complete: queue row '{...}' ... has no declared domain. Re-invoke from the worker "
    + $"surface with the domain recorded for its packet: `{BuildDomainReinvocation(..., suggestedDomain, mode)}`. ...");
```

On a two-domain host where domain A owns the marker and the host default, and the unit belongs to domain B — the exact shape this unit exists to fix — the worker emits:

```text
refused worker complete: queue row 'SDT-G25' for J-Tech-Japan/intent-system#1570 has no declared domain.
Re-invoke from the worker surface with the domain recorded for its packet:
`intent-cli worker complete ... --domain sekiban-as-a-service --outcome pr-created --pr 1572 --write --format json`.
A startup marker or PR-linkage recovery cannot supply missing durable identity.
```

`sekiban-as-a-service` is domain **A**. The unit is domain **B**'s. Running the emitted command verbatim is **accepted**:

```text
domain = sekiban-as-a-service
domain_source = explicit+session-layer-record+legacy-queue
execution_unit = SDT-G25
```

So a worker that follows the instruction the worker itself printed binds the unit to the wrong domain. It stopped here only on an unrelated stale-label check, not on anything domain-related.

**Why this is a real gap and not a nit.** ADR 0009 decision #1, landed by this PR, says the host default is a default binding and not identity evidence; `docs/en/07-recovery.md` as edited by this PR says the marker "and host default domain are display/default context only". This branch elevates the host default back into the suggested identity, on the one host shape where it is most likely to be wrong. The whole unit is about not letting a display/default surface answer a per-domain question.

**The correct pattern is already three branches away in the same function.** When no queue row matches and several domains are recorded, the code enumerates candidates instead of guessing:

```text
refused worker complete: J-Tech-Japan/intent-system#1570 has no queue domain and this host records multiple
session-layer domains (sekiban-as-a-service, sekiban-dcb-ts). Choose the packet domain explicitly:
`... --domain sekiban-as-a-service ...` or `... --domain sekiban-dcb-ts ...`. The startup marker is display-only.
```

**Required unblock action.** In the matched-legacy-row branch, when `recordedDomains.Length != 1`, enumerate `recordedDomains` the way the sibling branch does (falling back to the `<domain>` placeholder when none are recorded) rather than emitting `context.Config.Project.Domain` as the suggestion. A regression test on a two-domain host asserting the emitted recovery names both recorded domains — and does not name only the host default — would pin it.

**What I did not verify and am not claiming.** I did not run this against a real five-domain host, and I did not exercise `--write` past the refusal. The wrong-domain acceptance above is from a `--dry-run` invocation of the emitted command; I have not shown durable corruption, and `linked_pr` projection is unit-scoped so it still lands on the right row. The defect I am asserting is misleading recovery guidance on the reported host shape, nothing wider.

---

## Everything else — verified

### ADR-first is real, not retrospective

This is the check I expected to fail and it does not. The branch has two commits and the ADR arrived alone, 42 minutes ahead of the mechanics:

```text
6ebd6629  2026-08-20 13:06:29  docs: decide session-layer domain identity authority
          docs/adr/0009-session-layer-domain-identity-authority.md | 70 ++++++  (1 file changed)
1ad6d6b6  2026-08-20 13:48:15  fix: isolate multi-domain session-layer worker identity
          15 files changed, 744 insertions(+), 92 deletions(-)
```

ADR 0009 states the decision in words rather than leaving it inferable: `.intent-cli/session-layer-mode.json` is authoritative for host session-layer identity/mode (team-scoped beats domain-wide), durable queue/packet metadata is authoritative for the execution unit, and the generated block "is display and verification evidence only … it never establishes identity, selects a worker domain, or overrides the durable record". It covers additive migration, states what happens to an existing single-domain and existing multi-domain marker, and explicitly rejects the reporter's PR-linkage recovery as a substitute rather than enshrining it because it worked — which is what review-context #5 asked for.

The mechanical commit's only ADR edit narrows decision #1 for the legacy domain-less shape and labels it "a migration binding for that legacy shape, never an override of declared durable data". That is a refinement discovered in implementation, not a decision retrofitted to match code.

### AC1 — markers coexist, demonstrated on a genuine two-domain host

Two domains recorded (`sekiban-as-a-service`/alpha agmsg, `sekiban-dcb-ts`/alpha herdr-only), A generated first.

**Before** — released `0.23.1` asked for a hand edit rather than coexisting:

```json
{"written": false, "cause": "marker-absent",
 "summary": "Refusing to touch 'CLAUDE.md' because it has no explicit managed marker block for domain 'sekiban-dcb-ts' and team 'alpha'. Add the delimited placeholder from the guide first."}
```

**After** — this head appends and preserves:

```json
{"written": true, "marker_action": "appended", "preserved_existing_blocks": 1,
 "summary": "Appended the managed session-layer marker for domain 'sekiban-dcb-ts' and team 'alpha' in 'CLAUDE.md'; preserved 1 existing managed block(s)."}
```

```text
<!-- intent-cli:session-layer-marker:start domain="sekiban-as-a-service" team="alpha" -->
<!-- ... mode="agmsg" record-hash="sha256:4c6d9dfe…" -->
<!-- intent-cli:session-layer-marker:end -->
<!-- intent-cli:session-layer-marker:start domain="sekiban-dcb-ts" team="alpha" -->
<!-- ... mode="herdr-only" record-hash="sha256:c4ae70d5…" -->
<!-- intent-cli:session-layer-marker:end -->
```

Domain A's region is byte-identical and still a literal prefix of the file after B's write (verified by byte comparison, not by reading the diff).

### AC2 — the domain-B worker completes

Host default `sekiban-as-a-service` (domain A), queue row `SDT-G25` declared under `intents/sekiban-dcb-ts/…`.

**Before** — the reported block reproduces exactly:

```text
refused queue write: command identity 'sekiban-as-a-service/J-Tech-Japan/intent-system#1570' conflicts with
queue identity 'sekiban-dcb-ts/J-Tech-Japan/intent-system#1570' on unit 'SDT-G25'.
```

**After** — the domain refusal is gone and identity is resolved from the durable record:

```json
"domain": "sekiban-dcb-ts",
"domain_source": "queue-record",
"execution_unit": "SDT-G25",
"errors": ["complete.stale.not-claimed: …", "complete.stale.already-completed: …"]
```

The only remaining refusal is the ordinary lifecycle check against the real issue's real labels — not a domain conflict. That is the reported failure stopping for the reason it was supposed to stop being.

### AC3 — no silent rewrite of another domain's claim

Structurally eliminated: generation now either replaces **only** the requested domain/team block or appends a new one, so a block it does not own is never in the write path. `worker complete` never writes the marker at all — I compared the file bytes before and after the domain-B completion and they are identical, and `sekiban-as-a-service` remains the sole content of A's block.

One honest correction to the record, since the packet asserts it as measured baseline: **the silent-rewrite symptom did not reproduce on `0.23.1`** in my fixture. With a domain-qualified block present, released marker generation for another domain refuses with `marker-absent` (transcript above) rather than overwriting. On current released code the multi-domain failure mode is a refusal that demands a hand edit — which is what AC6 forbids — not a silent overwrite. That does not weaken the PR; it means the change fixes the reachable defect and makes the reported one unreachable by construction. I flag it only so the closeout record is not built on a symptom I could not observe.

### AC4 — recovery reachable from the worker's surface

Reachable, and correct in every branch except the one in the finding. On genuinely missing durable identity the worker fails closed with exit `1`, before any label or queue write, and names the act:

```text
refused worker complete: … has no declared domain. Re-invoke from the worker surface with the domain
recorded for its packet: `intent-cli worker complete … --domain <name> …`.
A startup marker or PR-linkage recovery cannot supply missing durable identity.
```

`guide worker issue-to-pr` carries the same rule as a hard rule, so a worker reading its own operating guide finds it without being told. PR-linkage recovery is named and refused in the ADR, the docs, and the runtime message — three places, consistently.

### AC5 — single-domain hosts unaffected, byte-identical

A marker written by released `0.23.1` on a single-domain host, then regenerated by this head:

```text
0.23.1 output   sha256 7c7080806b54647e954f11a3af20a31b39c6f8b1d3ad41da608eed58b5840f7d
this head       written = false, marker_action = "replaced"
                sha256 7c7080806b54647e954f11a3af20a31b39c6f8b1d3ad41da608eed58b5840f7d
```

Identical hash, no write, no migration.

### AC6 — existing multi-domain hosts need no hand edit

This is the case every affected install is in today, and the PR states what happens to the marker already present: it stays, and the next canonical `marker generate --domain B … --write` appends B beside it. Verified above — the `marker-absent` "add the placeholder from the guide first" demand is what disappears. The first-time refusal for a file with **no** managed block is deliberately retained, and `candidates.Length > 1` still refuses; the comment in `SessionLayerMarkerCommand` says why.

### AC7 — the walk is on a two-domain host

Both halves of the reported scenario were walked end to end on a host with two recorded domains, before and after, with each step's emitted output shown. No step of my verification used a single-domain fixture.

### Worker does not infer identity from PR linkage

Confirmed in the direction of the data flow: `TryPatchQueueStateLinkedPr` now takes the **resolved** execution unit and domain and filters candidate rows by them, so linkage follows identity. Identity never reads linkage. The colliding-issue-number case the comment describes is closed by the same filter.

### Out-of-scope boundaries hold

- `.intent-cli/session-layer-mode.json` and its store are **not** in the diff — review-context #7 satisfied.
- No rename or migration is demanded of any install.
- The reporter's PR-linkage recovery is rejected, not shipped as the mechanism.
- Child `--github-only` remains metadata-free and still does not read host queue state.

### Design alignment — `design_alignment_checked=true`

Consistent with the primary intent reference `intents/intent-cli/intent-tree/means/08-agent-message-orchestration.md`, which owns the session layer and holds that the durable per-domain record is the machine's truth while the visible surface is display. This change makes the display surface additive and demotes it out of the identity path, which is that node's model applied rather than excepted. Supporting reference `means/04-worker-interface-strategy.md` owns the worker completion contract that was failing closed with no forward path; the forward path now exists on the worker's own surface — modulo the finding above, which is precisely about the quality of that path.

### Docs mirrored

`docs/en` and `docs/ja` 02a / 05 / 07 / 12 all carry the same G724 statements; I read the ja `05` section and it states the same rule as the en one, including that the marker never supplies or overrides the domain. No mirror drift.

### Tests

- Focused suites at this head, in-tree: **321 passed / 0 failed** (`SessionLayerMarker*`, `WorkerComplete*`, `GuideWorkerIssueToPr*`, `SessionLayer*`).
- CI at `1ad6d6b6`: `Build and test (source contract)` pass, `npm package dry-run (no publish)` pass.
- `git diff --check`: clean.
- Head SHA `1ad6d6b6ce2b8767e6b50d7554977af6619a2cf1` confirmed before and after this pass.

The two new marker tests use two genuinely recorded domains and the end-to-end worker test asserts domain A's marker bytes are unchanged, so the suite tests the failing case rather than the one that always worked. Passing tests are treated as **necessary, not sufficient** — the verdict above rests on the packet/ADR conformance evidence, not the green run.

---

## Informational (no change required)

The appended block leaves the file without a trailing newline (`…:end -->` then EOF). Subsequent appends still separate correctly because the writer inserts the leading newline itself, but the file loses its final newline on first append, which some diff and lint tooling will flag.

<sub>Evidence collected at head `1ad6d6b6` with a build of this branch; before-state captured with released `intent-cli 0.23.1-d49984d-G721` on a fixture with two recorded domains.</sub>

tomohisa (commented):
# Re-review — G724 / PR #1572 (head `14b8953b`)

**Verdict: approve-equivalent. The single finding from [review 4987327560](https://github.com/J-Tech-Japan/intent-system/pull/1572#pullrequestreview-4987327560) is fixed; no remaining findings.** Submitted as `COMMENTED` because `gh pr review --approve` is rejected on a same-account PR (G696). The PR is still draft; this re-review does not merge, approve, or promote it.

Re-reviewed against the prior reviewed head `1ad6d6b6`. I rebuilt the CLI at the new head and re-ran the previously failing scenario myself rather than reading the diff and assuming.

---

## The finding is closed

The repair removes `context.Config.Project.Domain` from that branch entirely — the host default is no longer reachable as a suggested identity — and splits it in two: enumerate when the host records several domains, placeholder when it records none.

**Scenario 1 — two recorded domains + matched legacy domain-less row.** This is the exact invocation that previously emitted `--domain sekiban-as-a-service` for a domain-B unit. At `14b8953b`, on a host with `sekiban-as-a-service` and `sekiban-dcb-ts` recorded, host default `sekiban-as-a-service`, and legacy domain-less row `SDT-G25`:

```text
refused worker complete: queue row 'SDT-G25' for J-Tech-Japan/intent-system#1570 has no declared domain
and this host records multiple session-layer domains (sekiban-as-a-service, sekiban-dcb-ts).
Choose the packet domain explicitly:
`intent-cli worker complete … --domain sekiban-as-a-service --outcome pr-created --pr 1572 --write --format json` or
`intent-cli worker complete … --domain sekiban-dcb-ts     --outcome pr-created --pr 1572 --write --format json`.
A startup marker or PR-linkage recovery cannot supply missing durable identity.
```

Checked mechanically against that output:

```text
names sekiban-as-a-service                     : True
names sekiban-dcb-ts                           : True
enumerates with ' or '                         : True
says the startup marker cannot supply identity : True
OLD host-default-only suggestion recurs        : False
```

Exit `1`, `queue-state.json` unchanged. The old wording is structurally unreachable here now: the enumerating branch returns before the single-suggestion path, and that path no longer has a host-default value to emit.

**Scenario 2 — no recorded session-layer domains + matched legacy domain-less row.** Same fixture with the recorded-domain file absent, host default still `sekiban-as-a-service`:

```text
refused worker complete: queue row 'SDT-G25' … has no declared domain. Re-invoke from the worker surface
with the domain recorded for its packet:
`intent-cli worker complete … --domain <domain> --outcome pr-created --pr 1572 --write --format json`.
A startup marker or PR-linkage recovery cannot supply missing durable identity.
```

```text
emits <domain> placeholder                : True
host default leaked (sekiban-as-a-service): False
```

Exit `1`, no write. The placeholder is now the honest answer — the worker says it does not know rather than naming a value that merely happens to be configured. That is the behaviour the finding asked for, and it is consistent with ADR 0009 decision #1 and with `docs/en/07-recovery.md` as this PR wrote them: the host default is default context, never identity evidence.

The single-recorded-domain case does not reach either message — it resolves earlier as `session-layer-record+legacy-queue` — so the placeholder branch is correctly the zero-record case only.

## Repair scope is limited as described

One commit, two files:

```text
14b8953b  2026-08-20 14:14:31  fix: enumerate legacy domain recovery candidates
  src/IntentSystem.Cli/Commands/WorkerCompleteCommand.cs   | 15 ++++++++--
  tests/IntentSystem.Cli.Tests/WorkerCompleteCommandTests.cs | 33 ++++++++++++++++++++--
```

No production file outside `WorkerCompleteCommand.cs` was touched, no ADR or docs edit was made to paper over the finding, and the marker command, guide surface, and mirrored docs are untouched from the head I already reviewed.

The regression coverage pins both halves rather than only the one that changed: the existing multi-domain test now asserts **both** `--domain intent-cli` and `--domain sekiban-as-a-service` plus the `" or "` separator, and a new test asserts `--domain <domain>` **and** `DoesNotContain("--domain intent-cli")` for the zero-record case. That negative assertion is what stops the host default creeping back.

## Prior evidence re-verified at the new head

I did not carry the earlier verdict forward on trust; `WorkerCompleteCommand` changed, so I re-ran the load-bearing checks against `14b8953b`:

- **AC1** — domain B appended beside domain A: `marker_action = appended`, `preserved_existing_blocks = 1`, and domain A's region still a byte-for-byte prefix of the file.
- **AC2** — domain-B worker completion while A owns the visible marker: `domain = sekiban-dcb-ts`, `domain_source = queue-record`, `execution_unit = SDT-G25`; the marker file is byte-identical afterwards and still holds both managed blocks.
- **AC5** — single-domain compatibility: regenerating a marker originally produced by released `0.23.1` gives `written = false` and the same `sha256 7c7080806b54647e954f11a3af20a31b39c6f8b1d3ad41da608eed58b5840f7d` before and after.
- **AC3 / AC4 / AC6 / AC7**, the ADR-first commit ordering (`6ebd6629` ADR-only, then the mechanical commit), the PR-linkage rejection, the `session-layer-mode.json` out-of-scope boundary, and the en/ja doc mirrors are unchanged by this commit and remain as verified in the prior review.

## Design alignment — `design_alignment_checked=true`

Unchanged and reconfirmed against `intents/intent-cli/intent-tree/means/08-agent-message-orchestration.md` (durable per-domain record is the truth; the visible surface is display) and `means/04-worker-interface-strategy.md` (worker completion contract needs a forward path). The repair strengthens the alignment rather than merely satisfying a reviewer: the last place where a default/display value could answer a per-domain question is gone.

## Tests and state

- Focused suites at `14b8953b`, in-tree: **322 passed / 0 failed** (`SessionLayerMarker*`, `WorkerComplete*`, `GuideWorkerIssueToPr*`, `SessionLayer*`) — one more than the 321 at the prior head, matching the added regression test.
- CI at `14b8953b`: `Build and test (source contract)` pass, `npm package dry-run (no publish)` pass.
- `git diff --check` across the full PR range `0252948e..14b8953b`: clean.
- Head SHA `14b8953ba4f557014521e57e0354bd653951ed23` confirmed unchanged before and after this pass; PR state `OPEN`, `MERGEABLE`, still draft.

Passing tests remain **necessary, not sufficient** — this verdict rests on the two reproduced recovery transcripts and the re-verified AC evidence above.

The informational note from the prior review still stands and still requires nothing: an appended block leaves the file without a trailing newline.

<sub>Re-review evidence collected at head `14b8953b` with a build of this branch, on a fixture with two recorded domains and a second fixture with none.</sub>

PR #1578, opened 2026-08-21
G727: report stalled-work checkout freshness

### LINKED ISSUE #1575: G727: host-state commands answer from whatever the local checkout happens to be, so a stale clone reports findings that are already resolved
# G727: host-state commands answer from whatever the local checkout happens to be, so a stale clone reports findings that are already resolved

## Goal

An operator can tell whether a host-state answer was computed from a current view, without having to remember to check.

## Why This Slice Exists Now

Measured on this host today.

`automation stalled-work` reported `knowledge-writeback-pending` for G717 and G718. The design thread read that as a real regression and started re-recording evidence that was already correct.

It was not a regression. The checkout was **seven commits behind** `origin/main`, and the records existed on the remote. Syncing and re-running gave a different, correct answer.

**Nothing in the output indicated it had been produced from a stale tree.** A finding computed from a seven-commit-old view looks exactly like one computed from the tip.

This host is written to by five domains — `sekiban-dcb-ts`, `sekiban-as-a-service`, `sekiban-wasm-runtime`, `idd-com`, `intent-cli`. A local checkout goes stale within minutes here, not days. **Staleness is the normal condition, not an unusual one.**

The failure direction is the dangerous one. A stale tree **over-reports**: it shows work as outstanding that is already done, which invites redoing durable evidence. On 2026-08-19 the same class of mistake produced knowledge-writeback records written against a commit that a later rebase orphaned.

This is not a defect in the finding logic. **Every finding was correct for the tree it was given.** The gap is that the answer carries no statement about what it was computed from.

## Current Observed State

- `stalled-work` reports against the local working tree with no freshness statement.
- On a multi-domain host the tree is behind most of the time.
- The over-report direction leads to redoing durable records rather than skipping work.

- The finding logic is correct and not in scope.
- Read-only commands must stay read-only. Fetching or resetting as a side effect would be a larger defect than this one.
- Operators sync before acting when they remember. The unit exists because remembering is not a mechanism.

## Target Repo / Path / Part

- Target paths: `src/IntentSystem.Cli/Commands/` (the `automation stalled-work` and sibling host-state reporting surfaces), `docs/en/`, `docs/ja/`, and their tests
- Part: what a host-state reporting command tells the operator about the freshness of the checkout it answered from

## In Scope

- Making checkout freshness visible in host-state reporting output.
- Documenting what these answers are computed from.
- Surveying which sibling surfaces share the property.

## Out Of Scope

- Changing any finding logic.
- Auto-syncing, auto-fetching, or mutating the checkout.
- Requiring network access for these commands to function.

Self-contained. The measured incident, the seven-commit gap, the multi-domain reason staleness is normal, the over-report direction, and the boundary against auto-syncing are all above.

## Acceptance Criteria

- [ ] A host-state reporting command makes the freshness of the checkout it answered from visible in its own output. An operator must not have to remember to sync before believing a finding.
- [ ] It does not require network access to be useful. A command that becomes unusable offline is worse than the condition being fixed — if freshness cannot be determined, **say that** rather than implying currency.
- [ ] Nothing is auto-synced. Fetching or resetting a host checkout as a side effect of a read-only report would be a larger defect than this one.
- [ ] The ordinary case — a current checkout — stays quiet. A notice on every invocation becomes invisible within a day, and this unit would have added noise instead of signal.
- [ ] Demonstrated on a genuinely stale checkout against a moved remote, showing output before and after syncing; and on a current checkout, showing it stays quiet.
- [ ] State whether other read-only host-state surfaces share this property. Report what was surveyed **including the ones that do not**, so the next reader knows the sweep was real.

## Verification

- Paste `stalled-work` from a stale checkout, then after syncing.
- Paste it from a current checkout, showing silence.
- Paste the survey of sibling surfaces.

## Related Links

- Same-day consequence of acting on a stale read: knowledge-writeback records written against a commit a later rebase orphaned

## Knowledge Maintenance

Node 03 records that an answer computed from a stale view is not wrong — it is answering a different question than the one asked — and that on a host several domains write to, the view is stale by default.

## Base Branch Policy

Do not make the command sync. It is read-only, several domains write to this host, and a read that mutates the tree would be a far worse defect than the one being fixed.

Do not print a freshness banner unconditionally. The value of this notice is entirely in its rarity; on a current checkout it must say nothing. A line that appears every time will be filtered out by the reader within a day, and then the next stale answer will be believed exactly as this one was.

If freshness genuinely cannot be determined — no remote configured, offline, whatever — say so plainly. "Unknown" is honest and actionable. Implied currency is what caused this.

tomohisa:
Closes #1575

## G727 repair

This update addresses exactly the two semantic findings from the review at
`ced5d13429794e887d973b2ef52f70e93d8b5642`: the unbounded remote probe (AC2)
and the over-broad sibling survey (AC6). The finding logic, stale/current
behavior, heartbeat inheritance, and scope remain otherwise unchanged.

## Finding 1 — bounded, noninteractive freshness probe (AC2)

The shared `GitRemoteCommandRunner` keeps its historical unbounded behavior
for issue/status/publish callers. `automation stalled-work` now uses the
freshness-only `CheckoutFreshnessGitCommandRunner` with a three-second bound.
That deliberate separation avoids silently imposing this policy on unrelated
Git operations.

The scoped runner:

- reads stdout and stderr concurrently and puts both reads plus process exit
  under one `Task.WhenAll(...).WaitAsync(timeout)` bound;
- closes redirected stdin and sets `GIT_TERMINAL_PROMPT=0` plus SSH
  `BatchMode=yes`, so credential and SSH prompts cannot block the report;
- kills the entire Git process tree when the bound expires; and
- returns `checkout_freshness: unknown` with a reason containing the timeout,
  allowing the existing report and findings to continue.

### Real blackhole transcript

The built CLI ran from a real Git clone whose `origin` was changed to a local
TCP listener on `127.0.0.1` that accepted the connection and never replied:

```text
BLACKHOLE_EXIT=0
checkout_freshness: {
  status: unknown,
  local_head: c4641e8a1519bfe8883a2ed3cc674b1d4c406f75,
  reason: "git ls-remote --symref origin HEAD timed out after 3 seconds"
}
warning: ... Do not treat this report as evidence that the checkout is current ...
  This command is read-only and does not fetch, pull, reset, or sync.
HEAD after: c4641e8a1519bfe8883a2ed3cc674b1d4c406f75
refs/remotes/origin/main after: c4641e8a1519bfe8883a2ed3cc674b1d4c406f75
worktree: unchanged
```

The full command returned in 9.744s because it continued through the normal
GitHub candidate reads after the freshness probe timed out; the freshness
probe itself returned its explicit three-second timeout result and did not
silence the report. The deterministic regression uses the same real listener
with a 250ms freshness-only bound, asserts completion within three seconds,
asserts the preserved `published-not-delegated` finding, and verifies HEAD,
the tracking ref, and worktree status are unchanged.

The existing fast-failure unknown case and moved-remote stale → external-sync
→ current-silence cases remain covered.

## Finding 2 — honest sibling survey (AC6)

The survey no longer calls all siblings unaffected.

- `intent status` was independently demonstrated on the same stale clone to
  return stale local queue state without any checkout-provenance statement.
- Source inspection shows `automation summary`, `automation state-doctor`,
  `host-loop-next-action`, and `automation heartbeat` read `context.RepoRoot`
  and therefore share the unstated-checkout provenance property. Heartbeat
  inherits the new warning in this slice; the other RepoRoot-reading surfaces
  remain a named follow-up.
- `status brief` and host-review diagnostics were inspected and do not read
  `RepoRoot` for these answers. They are unaffected by this specific
  unstated-checkout path, not globally certified as current.

The remaining sibling follow-up is to define and test an explicit
freshness/provenance contract for the RepoRoot-reading surfaces. This PR stays
scoped to `stalled-work` plus heartbeat inheritance. Both English and Japanese
mirrored docs carry the same qualification.

## Verification

- Focused `AutomationStalledWorkCommandTests`: 119 passed.
- Targeted stale/current/fast-failure/blackhole tests: 3 passed.
- Full solution: 5,172 total, 5,171 passed, 1 intentional skip, 0 failed.
- `git diff --check`: clean.
- CI at `8ef637f5fd4946024ce762a9bef6f2597a00a5e3`: `Build and test (source contract)` passed; `npm package dry-run (no publish)` passed.

## Canonical rereview-ready transition

`intent-cli worker result-summary --kind pr-comment-fix --repo J-Tech-Japan/intent-system --pr 1578 --outcome repair-pushed --format json` returned `status: completed`. The canonical GitHub-only worker claim and `intent-cli worker complete --kind pr --number 1578 --outcome repair-pushed --write --format json` returned `proceed: true`, `applied: true`, added `intent-pr-rereview-ready`, and removed `intent-pr-update-in-progress` plus `intent-pr-request-update`. No host claim or host metadata was acquired or changed.

No host claim was acquired or released from the child, and no host state,
intent tree, tag, release, merge, or closeout was changed.

### DIFF
--- a/docs/en/08-command-reference.md
+++ b/docs/en/08-command-reference.md
@@ -365,6 +365,29 @@ when quota makes GitHub-consulting surfaces inoperable. Callers decide whether
 to wait for reset; no command retries, sleeps, schedules a reset, budgets
 requests, changes transport, caches, or batches in G673.
 
+### Checkout freshness in host-state reports (G727)
+
+`automation stalled-work` also reports what checkout its answer came from when
+that fact is not safely current. It compares the local `HEAD` with the actual
+default-branch `HEAD` returned by `git ls-remote --symref origin HEAD`:
+
+- `checkout_freshness: stale` names the local and remote commit IDs and tells
+  the operator to sync and rerun the report.
+- A genuinely current checkout omits `checkout_freshness` and emits no
+  freshness banner. This keeps the notice rare enough to remain useful.
+- If the remote cannot be queried (offline, missing remote, or an incomplete
+  response), `checkout_freshness: unknown` is emitted with a reason. Unknown
+  must not be read as current.
+
+The probe is read-only: it uses no `fetch`, `pull`, `reset`, or other sync
+operation, and it does not change the existing stalled-work finding logic.
+`automation heartbeat` wraps `stalled-work` and therefore carries the same
+warning. The sibling read-only surfaces were surveyed: `automation summary`,
+`intent status`, `automation state-doctor`, `status brief`,
+`host-loop-next-action`, and host-review diagnostics do not make this
+stalled-work checkout-freshness claim and remain unchanged. The survey is not
+a reason to widen G727 to those surfaces.
+
 G672 adds an optional invoking-role pointer (preview-through-1.x):
 
 ```bash
--- a/docs/en/08-command-reference.md
+++ b/docs/en/08-command-reference.md
@@ -378,15 +378,27 @@ default-branch `HEAD` returned by `git ls-remote --symref origin HEAD`:
 - If the remote cannot be queried (offline, missing remote, or an incomplete
   response), `checkout_freshness: unknown` is emitted with a reason. Unknown
   must not be read as current.
+- The remote probe is bounded to three seconds, reads stdout and stderr under
+  that same bound, kills the Git process tree on expiry, closes stdin, and
+  disables terminal/SSH prompts. A timeout is therefore an actionable
+  `unknown`, not a stalled wake.
 
 The probe is read-only: it uses no `fetch`, `pull`, `reset`, or other sync
 operation, and it does not change the existing stalled-work finding logic.
 `automation heartbeat` wraps `stalled-work` and therefore carries the same
-warning. The sibling read-only surfaces were surveyed: `automation summary`,
-`intent status`, `automation state-doctor`, `status brief`,
-`host-loop-next-action`, and host-review diagnostics do not make this
-stalled-work checkout-freshness claim and remain unchanged. The survey is not
-a reason to widen G727 to those surfaces.
+warning. The survey does **not** classify every sibling as unaffected:
+`intent status` was independently demonstrated on the same stale clone to
+return stale local queue state without checkout provenance. Source inspection
+also shows `automation summary`, `automation state-doctor`,
+`host-loop-next-action`, and `automation heartbeat` read `context.RepoRoot`, so
+they share the unstated-checkout provenance property (heartbeat inherits the
+new warning here, while the others remain follow-up work). This slice is
+deliberately scoped to `stalled-work` plus heartbeat inheritance. A follow-up
+must extend and test an explicit freshness/provenance contract for those
+RepoRoot-reading siblings. `status brief` and host-review diagnostics were
+also inspected and do not read `RepoRoot` for these answers; they are
+unaffected by this specific unstated-checkout path, not globally certified as
+current. The survey is not a reason to widen G727 here.
 
 G672 adds an optional invoking-role pointer (preview-through-1.x):
 
--- a/docs/en/12-agent-message-orchestration.md
+++ b/docs/en/12-agent-message-orchestration.md
@@ -2666,6 +2666,18 @@ timer.
 intent-cli automation stalled-work --domain <domain> --repo <owner/repo> --format json
 ```
 
+G727 makes the input view explicit without adding noise to healthy wakes.
+`stalled-work` emits `checkout_freshness: stale` with the local `HEAD`, the
+actual `origin` default-branch `HEAD`, and a sync-and-rerun instruction when
+the remote has moved. A genuinely current checkout emits no freshness notice.
+When the remote cannot be checked, it emits `checkout_freshness: unknown`
+with the reason instead of implying currency. The probe uses only
+`git rev-parse` and `git ls-remote --symref`; it never fetches, pulls, resets,
+or otherwise syncs the shared host checkout. `automation heartbeat` carries
+the same warning because it wraps this scan. Other read-only host-state
+surfaces remain outside this slice because they do not make the same
+stalled-work answer-from-checkout claim.
+
 - **Never defer** — process every actionable item the check reports in THIS
   wake (delegate, repair, or route to closeout) before sleeping. Do not
   announce work for a future wake unless an explicit fallback/legacy timer
--- a/docs/en/12-agent-message-orchestration.md
+++ b/docs/en/12-agent-message-orchestration.md
@@ -2673,10 +2673,17 @@ the remote has moved. A genuinely current checkout emits no freshness notice.
 When the remote cannot be checked, it emits `checkout_freshness: unknown`
 with the reason instead of implying currency. The probe uses only
 `git rev-parse` and `git ls-remote --symref`; it never fetches, pulls, resets,
-or otherwise syncs the shared host checkout. `automation heartbeat` carries
-the same warning because it wraps this scan. Other read-only host-state
-surfaces remain outside this slice because they do not make the same
-stalled-work answer-from-checkout claim.
+or otherwise syncs the shared host checkout. The remote probe is bounded to
+three seconds across process exit and both output reads, kills the process
+tree on expiry, closes stdin, and disables terminal/SSH prompts; expiry is an
+explicit `unknown` result. `automation heartbeat` carries the same warning
+because it wraps this scan. The sibling survey found that `intent status` is
+independently stale-provenance-bearing on the same clone, and that
+`automation summary`, `automation state-doctor`, and
+`host-loop-next-action` also read `context.RepoRoot`; they need a follow-up
+freshness/provenance contract. This slice remains scoped to stalled-work plus
+heartbeat inheritance. `status brief` and host-review diagnostics do not read
+`RepoRoot` for these answers and are unaffected by this specific path.
 
 - **Never defer** — process every actionable item the check reports in THIS
   wake (delegate, repair, or route to closeout) before sleeping. Do not
--- a/docs/ja/08-command-reference.md
+++ b/docs/ja/08-command-reference.md
@@ -342,6 +342,27 @@ quota exhaustion では machine-readable な `cause: github-api-quota-exhausted`
 `ok` 以外の verdict を返します。reset を待つかどうかは caller が判断します。G673 は retry、sleep、
 reset scheduling、request budgeting、transport change、cache、batching を追加しません。
 
+### host-state report の checkout freshness（G727）
+
+`automation stalled-work` は、answer を計算した checkout が current であると
+安全に言えない場合、その事実も報告します。local `HEAD` と、
+`git ls-remote --symref origin HEAD` が返す実際の default branch の `HEAD` を比較します。
+
+- `checkout_freshness: stale` は local と remote の commit ID を示し、sync して
+  report を再実行するよう案内します。
+- 本当に current な checkout では `checkout_freshness` を省略し、freshness banner も
+  出しません。notice を稀に保つことで signal としての意味を維持します。
+- remote を問い合わせられない場合（offline、remote 不在、応答不完全）は理由つきで
+  `checkout_freshness: unknown` を出します。unknown を current と解釈してはいけません。
+
+この probe は read-only です。`fetch`、`pull`、`reset`、その他の sync operation を
+実行せず、既存の stalled-work の finding logic も変更しません。`automation heartbeat` は
+`stalled-work` を wrapper するため同じ warning を運びます。兄弟の read-only surface も
+survey しました: `automation summary`、`intent status`、`automation state-doctor`、
+`status brief`、`host-loop-next-action`、host-review diagnostics はこの
+stalled-work checkout-freshness claim を行わず、変更していません。この survey を理由に
+G727 の scope を広げません。
+
 G672 は invoking role の pointer を optional に追加します（preview-through-1.x）。
 
 ```bash
--- a/docs/ja/08-command-reference.md
+++ b/docs/ja/08-command-reference.md
@@ -354,14 +354,23 @@ reset scheduling、request budgeting、transport change、cache、batching を�
   出しません。notice を稀に保つことで signal としての意味を維持します。
 - remote を問い合わせられない場合（offline、remote 不在、応答不完全）は理由つきで
   `checkout_freshness: unknown` を出します。unknown を current と解釈してはいけません。
+- remote probe は 3 秒で bounded です。stdout と stderr の read も同じ bound に含め、
+  expiry では Git process tree を終了させ、stdin を閉じ、terminal/SSH prompt を無効にします。
+  timeout は wake を止めず、理由つきの actionable な `unknown` になります。
 
 この probe は read-only です。`fetch`、`pull`、`reset`、その他の sync operation を
 実行せず、既存の stalled-work の finding logic も変更しません。`automation heartbeat` は
-`stalled-work` を wrapper するため同じ warning を運びます。兄弟の read-only surface も
-survey しました: `automation summary`、`intent status`、`automation state-doctor`、
-`status brief`、`host-loop-next-action`、host-review diagnostics はこの
-stalled-work checkout-freshness claim を行わず、変更していません。この survey を理由に
-G727 の scope を広げません。
+`stalled-work` を wrapper するため同じ warning を運びます。兄弟をすべて unaffected とは
+分類していません。同じ stale clone で `intent status` が stale な local queue state を
+checkout provenance なしに返すことを独立に実証しました。source survey でも
+`automation summary`、`automation state-doctor`、`host-loop-next-action`、
+`automation heartbeat` は `context.RepoRoot` を読むため、unstated checkout provenance の
+property を共有します（heartbeat はこの slice の warning を継承し、他は follow-up です）。
+この slice の scope は `stalled-work` と heartbeat inheritance に限定します。これらの
+RepoRoot-reading sibling には、freshness/provenance contract と test を追加する follow-up
+が必要です。`status brief` と host-review diagnostics はこの answer で `RepoRoot` を読まず、
+この特定の unstated-checkout path には unaffected です。ただし全体が current だと証明した
+わけではありません。この survey を理由に G727 の scope をここで広げません。
 
 G672 は invoking role の pointer を optional に追加します（preview-through-1.x）。
 
--- a/docs/ja/12-agent-message-orchestration.md
+++ b/docs/ja/12-agent-message-orchestration.md
@@ -2325,6 +2325,16 @@ publish-then-sleep と silent-completion の stall class を、タイマーを�
 intent-cli automation stalled-work --domain <domain> --repo <owner/repo> --format json
 ```
 
+G727 は healthy な wake に noise を追加せず、answer の input view を明示します。
+remote が動いている場合、`stalled-work` は local `HEAD`、実際の `origin` default branch の
+`HEAD`、sync 後に再実行する案内とともに `checkout_freshness: stale` を出します。
+本当に current な checkout では freshness notice を出しません。remote を確認できない場合は
+current を暗黙に意味させず、理由つきで `checkout_freshness: unknown` を出します。probe が使うのは
+`git rev-parse` と `git ls-remote --symref` だけで、共有 host checkout を fetch、pull、reset、
+その他の sync で変更しません。`automation heartbeat` はこの scan を wrapper するため同じ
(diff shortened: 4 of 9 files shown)


tomohisa (commented):
# Semantic review — G727 / PR #1578 (head `ced5d134`)

**Verdict: request-update — two findings, tied to AC2 and AC6.** The core mechanism is right and I verified it end to end on a genuinely moved remote, including the hard part (silence on a current checkout). Both findings are about honesty at the edges: what happens when the remote does not answer, and what the survey says about the siblings. Submitted as `COMMENTED` because `gh pr review --approve` is rejected on a same-account PR (G696). Draft; this review does not promote, transition labels, merge, close out, release, tag, or begin release prep.

Execution unit **G727**, packet `.intent-cli/issues/G727`, linked issue **#1575**, `Closes #1575` present and matching (G311).

I built the CLI at this head in an isolated worktree and produced every transcript below against a real bare `origin` I created for the purpose. Nothing is copied from the PR body.

---

## Finding 1 — the command hangs indefinitely on an unresponsive remote (AC2)

**AC2: "The behaviour does not require network access to be useful. A command that becomes unusable offline is worse than the condition being fixed; if freshness cannot be determined, say that rather than implying currency."**

`CheckoutFreshnessProbe.Capture` calls `git ls-remote --symref origin HEAD` through `GitRemoteCommandRunner`, which is:

```csharp
using var process = Process.Start(startInfo) ?? throw new InvalidOperationException(...);
var stdOut = process.StandardOutput.ReadToEnd();
var stdErr = process.StandardError.ReadToEnd();
process.WaitForExit();          // no timeout
```

No timeout, no cancellation, and stdin is not redirected (`RedirectStandardInput` is not set and `GIT_TERMINAL_PROMPT` is not forced to `0`).

The PR's offline evidence is a remote that **fails fast**. I reproduced that case and it is fine — a missing local path returns in ~6s with an honest result:

```json
{"status": "unknown", "local_head": "fa2c5610…",
 "reason": "fatal: '…/does-not-exist.git' does not appear to be a git repository\nfatal: Could not read from remote repository. …"}
```

But "offline" is not only fast failure. I pointed `origin` at a **local** TCP listener that accepts the connection and never replies — no external traffic, entirely on 127.0.0.1 — which is the ordinary shape of a hung VPN, a dropped route, or a wedged proxy. Control first, to prove the fixture bites:

```text
plain `git ls-remote --symref origin HEAD` against the blackhole: still running when killed at 20s
```

Then the actual command:

```text
intent-cli automation stalled-work --domain intent-cli --repo … --format json
  → still running when SIGKILLed at the 45s cap; exit 137; ZERO bytes of output
```

It did not return `unknown`. It did not return the findings. It returned **nothing**, and would have waited indefinitely. That inverts the criterion: before this change `stalled-work` always answered (sometimes from a stale tree); after it, a wedged remote can stop the answer from arriving at all. On a host where `stalled-work` feeds `automation heartbeat` and the G539 watchdog, a probe that never returns silences the watchdog rather than informing it.

**Required unblock action.** Bound the probe: run the remote query with a short timeout (a few seconds) and treat expiry as `unknown` with a reason naming the timeout, and make it non-interactive — redirect stdin and/or set `GIT_TERMINAL_PROMPT=0` (and `GIT_SSH_COMMAND` batch mode where applicable) so a credential-requiring remote cannot block on a prompt. A regression test against an unresponsive local listener would pin it; the fast-failure test that exists today cannot.

**What I am not claiming.** I did not reproduce an interactive credential prompt — that needs a real authenticating remote. The prompt risk is a code-level reading of the missing stdin redirect; the hang is measured.

## Finding 2 — the sibling survey's "unaffected" claim is not accurate (AC6)

**AC6: "State whether other read-only host-state surfaces share this property. Report what was surveyed including the ones that do not, so the next reader knows the sweep was real."**

The survey happened and names negatives, which is the right shape. But the claim itself is the circular one:

> Unaffected negatives: `automation summary`, `intent status`, `automation state-doctor`, `status brief`, `host-loop-next-action`, and host-review diagnostics **do not make the same stalled-work answer-from-checkout claim** and remain unchanged.

"Does not make the same *stalled-work* claim" is true by construction of the words and says nothing about whether they answer from an unstated checkout. I tested it. On the same genuinely stale clone — a host queue-state where `Z1` is `completed` on `origin` and still `queued` locally:

```text
intent-cli intent status --domain intent-cli --format json
  queue_state_path: …/clone/.intent-cli/queue-state.json
  queued: [ { "execution_unit": "Z1", "state": "queued" } ]        ← the stale answer
  checkout_freshness: absent, no freshness warning
```

That is precisely the reported failure shape: an answer computed from a stale local view, over-reporting work as outstanding that is already done, with nothing saying which tree it came from. A provenance sweep of the sources agrees — `AutomationSummaryAnalyzer` (7 `RepoRoot` references), `IntentStatusCommand` (3), `AutomationStateDoctorCommand` (4), `AutomationHostLoopNextActionCommand` (2), `AutomationHeartbeatCommand` (8) all answer from `context.RepoRoot`; only `StatusBriefCommand` and the host-review diagnostics have none.

**Required unblock action.** This is a reporting fix, not a widening. The packet asks for an honest survey, not for the siblings to be fixed here. Change the survey wording — in the PR body and both doc mirrors — to say that `intent status` (and the other `RepoRoot`-reading surfaces) share the unstated-checkout provenance property, that this slice deliberately scopes the freshness statement to `stalled-work` and its heartbeat consumer, and name the follow-up that would extend it. As written, the next reader is told those surfaces are fine when at least one demonstrably is not.

---

## What is verified and correct

### AC1 and AC5 — the stale case, on a genuinely moved remote

Real bare `origin`, a clone, then a commit pushed to `origin` so the clone is truly behind — not a fixture flag:

```json
"checkout_freshness": { "status": "stale",
  "local_head": "8b8a2d3004c6d44b2999bb6065e4da8545732e35",
  "default_branch": "main",
  "remote_head": "fa2c5610a9c5be31e76d3baafc9732998c9ce794" }
```
```text
warning: checkout freshness: stale; this answer was computed from local HEAD 8b8a2d30… but
origin/main is fa2c5610…. Sync the checkout and re-run the report; this command is read-only
and does not fetch, pull, reset, or sync.
```

Both commit IDs and the default branch are named, and the remedy is stated. A reader can act without consulting anything else.

### AC3 — nothing is auto-synced, proven by refs

I captured `HEAD` and the tracking ref around the stale report:

```text
before: HEAD 8b8a2d30…   refs/remotes/origin/main 8b8a2d30…
after : HEAD 8b8a2d30…   refs/remotes/origin/main 8b8a2d30…   working tree clean
```

Unchanged, including the tracking ref — `ls-remote` reads without updating refs, unlike `fetch`. The diff introduces no `fetch`/`pull`/`reset` call; every match for those words is documentation prose stating the absence.

### AC4 — silence on a current checkout

The criterion review-context calls the most important. After an **external** operator sync (`git fetch origin main && git merge --ff-only origin/main`, run by me, not by the report):

```text
checkout_freshness present in output: False
freshness warnings: 0
SILENT on a current checkout: True
```

The field is omitted entirely, not set to `"current"` — so nothing new appears in the ordinary case and there is nothing for a reader to learn to filter out.

### Finding logic untouched

The `AutomationStalledWorkCommand` diff is **purely additive** — zero removed lines. No collector, no kind, no threshold was altered, which is what review-context #4 asks for. The probe returns `null` for a non-Git root, so unit workspaces and non-Git host directories keep their historical output instead of gaining an `unknown` banner on every invocation — a deliberate touch I would otherwise have flagged as noise.

### Heartbeat inheritance verified positively

The PR claims heartbeat inherits the warning through the shared analyzer. It does:

```text
heartbeat carries the freshness warning: True
  - checkout freshness: stale; this answer was computed from local HEAD 6e0454d5… but origin/main is 1093e76d…
```

Note the asymmetry, offered as an observation rather than a finding: heartbeat surfaces the *warning* but not the structured `checkout_freshness` field, so a machine consumer of heartbeat JSON sees the prose and not the status.

### Tests, docs, scope

New tests build a **real** bare remote and clone (`git init --bare`, `RunGit(root, "clone", remote, …)`, `MoveRemoteForward()`, `SyncToOrigin()`) rather than stubbing a boolean, and assert stale→sync→silent plus the offline unknown path. Docs are mirrored across `docs/{en,ja}/08-command-reference.md` and `12-agent-message-orchestration.md`. The diff is 7 files, 552 insertions, 0 deletions, confined to the contract's target paths.

### Design alignment — `design_alignment_checked=true`

Consistent with the declared reference `intents/intent-cli/intent-tree/means/03-state-and-audit-strategy.md`, which owns state and audit reporting — an answer that does not state what it was computed from is exactly that node's concern, and this change adds the statement without touching what is computed. Node 03's standard also underwrites Finding 2: a survey that reports a property as absent when it is present is a claim stronger than its evidence.

### CI, head, whitespace

- CI at `ced5d134`: `Build and test (source contract)` pass, `npm package dry-run (no publish)` pass.
- `git diff --check` across `728989c6..ced5d134`: clean.
- Head SHA `ced5d13429794e887d973b2ef52f70e93d8b5642` confirmed before and after this pass.

Passing tests are **necessary, not sufficient** — the two findings above are both invisible to the current suite, which is the point.

<sub>Evidence collected at head `ced5d134` with a build of this branch. The stale/current cases used a bare `origin` and clone created under my scratch directory; the unresponsive-remote case used a listener bound to 127.0.0.1 that accepts and never replies, stopped afterwards. No real repository, ref, or remote was modified.</sub>

tomohisa:
Repaired both semantic findings at `8ef637f5fd4946024ce762a9bef6f2597a00a5e3`.

- Freshness uses a scoped three-second noninteractive Git runner; concurrent stdout/stderr reads share the bound, expiry kills the process tree, and timeout returns `checkout_freshness=unknown` while preserving findings. The real 127.0.0.1 blackhole transcript and HEAD/ref/worktree invariants are in the body.
- The EN/JA sibling survey now records the independently demonstrated stale `intent status` path, the other `RepoRoot`-reading follow-ups, and only the supported unaffected negatives.
- Focused: 119 passed. Full: 5,172 total / 5,171 passed / 1 skip / 0 failed. CI checks pass.

Canonical worker completion transitioned the PR to `intent-pr-rereview-ready`; no host claim or host metadata was changed.

tomohisa (commented):
# Re-review — G727 / PR #1578 (head `8ef637f5`)

**Verdict: approve-equivalent. Both findings from [review 4989221426](https://github.com/J-Tech-Japan/intent-system/pull/1578#pullrequestreview-4989221426) are fixed; no new findings.** Submitted as `COMMENTED` because `gh pr review --approve` is rejected on a same-account PR (G696). Still draft; this re-review does not promote, transition labels, merge, close out, release, tag, or begin release prep.

I rebuilt at the new head and re-ran both failures against the same fixtures that produced them, rather than reading the diff and inferring.

---

## Finding 1 — bounded, non-interactive probe. Fixed.

Same local blackhole listener as before — accepts on 127.0.0.1 and never replies, no external traffic — with the actual command and **no external kill**:

```text
intent-cli automation stalled-work --domain intent-cli --repo … --format json
  exit=0   elapsed=10.0s   (returned on its own; no SIGKILL was issued)

"checkout_freshness": {
  "status": "unknown",
  "local_head": "6e0454d5d1585a4fe10d02c44769a332f1de2503",
  "reason": "git ls-remote --symref origin HEAD timed out after 3 seconds"
}
warning: checkout freshness: unknown; git ls-remote --symref origin HEAD timed out after 3 seconds.
Do not treat this report as evidence that the checkout is current; verify the checkout and re-run.
This command is read-only and does not fetch, pull, reset, or sync.
```

At the prior head this same invocation produced zero bytes and had to be SIGKILLed at 45s. Now it returns inside the declared bound, the reason **names the timeout** rather than hiding behind a generic message, and — the part that matters for the watchdog — **the report itself still arrives**: the response carried `stalled`, `items`, and `warnings`, so a wedged remote degrades one field instead of silencing the answer.

Checked the rest of what a timeout must not leave behind:

```text
lingering git children after expiry : none (pgrep -fl ls-remote → no rows)
refs after                          : HEAD 6e0454d5…  refs/remotes/origin/main 6e0454d5…  (identical to before)
worktree                            : clean
stdin closed / no tty               : exit=0 in 8.7s — no prompt, no block
```

**Implementation inspected as the envelope asked.** The concurrency hazard I was worried about is handled explicitly — both pipes are read asynchronously and **both reads plus process exit are inside the same bound**:

```csharp
var stdOutTask = process.StandardOutput.ReadToEndAsync();
var stdErrTask = process.StandardError.ReadToEndAsync();
var completion = Task.WhenAll(process.WaitForExitAsync(), stdOutTask, stdErrTask);
completion.WaitAsync(boundedTimeout).GetAwaiter().GetResult();
```

with a comment naming the failure mode it avoids (a synchronous `ReadToEnd` on one pipe blocking before `WaitForExit` when the other fills). On expiry it calls `Kill(entireProcessTree: true)`, guarded against the exited/reaped races, and returns `TimedOut = true` which the probe maps to `unknown`.

**Unrelated Git callers were not silently changed.** The policy lives in a separate `CheckoutFreshnessGitCommandRunner`; `GitRemoteCommandRunner` delegates with `timeout: null, nonInteractive: false`, so `BugImplementationIssueCommand`, `IssueCreateCommand`, `IssuePublishCommand`, `IssueStatusCommand`, and `QueueDispatchCommand` each still construct the historical unbounded runner. Non-interactivity is scoped the same way: stdin redirected **and closed**, `GIT_TERMINAL_PROMPT=0`, and `GIT_SSH_COMMAND` set to batch mode — appending `-o BatchMode=yes` to an operator's configured command rather than discarding it.

The 3-second default is a judgement call I am content with: long enough for an ordinary remote, short enough that a stalled host wake still completes, and the code says why.

### The other paths still behave

```text
fast-failure remote (missing local path) : status=unknown, reason "fatal: '…does-not-exist.git' does not appear to be…"
genuinely moved remote                   : status=stale, local 6e0454d5d158, remote 1093e76d7186, branch main
after external operator sync             : checkout_freshness present=False, freshness warnings=0 → SILENT
```

The current-checkout case — the one review-context calls decisive — still omits the field entirely rather than emitting `"current"`.

## Finding 2 — honest sibling survey. Fixed.

The circular "do not make this stalled-work claim" wording is gone. Both mirrors and the PR body now say the same thing, and they say it in the right direction:

> The survey does **not** classify every sibling as unaffected: `intent status` was independently demonstrated on the same stale clone to return stale local queue state without checkout provenance. Source inspection also shows `automation summary`, `automation state-doctor`, `host-loop-next-action`, and `automation heartbeat` read `context.RepoRoot`, so they share the unstated-checkout provenance property (heartbeat inherits the new warning here, while the others remain follow-up work). This slice is deliberately scoped to `stalled-work` plus heartbeat inheritance. A follow-up must extend and test an explicit freshness/provenance contract for those RepoRoot-reading siblings. `status brief` and host-review diagnostics were also inspected and do not read `RepoRoot` for these answers; they are **unaffected by this specific unstated-checkout path, not globally certified as current**.

That matches what I measured and swept myself: `intent status` demonstrated stale, the four `RepoRoot`-readers classified on source evidence, and the two non-readers reserved narrowly rather than blessed. The Japanese mirror carries the same content including the "全体が current だと証明したわけではありません" reservation. Scope is stated as deliberate with a named follow-up, so the correction is reporting-only — nothing was widened into the siblings.

## Rechecked at this head

- **Blackhole test quality**: the new `Analyze_BlackholeRemoteTimesOutAndLeavesCheckoutUntouched_G727` binds a **real** `TcpListener(IPAddress.Loopback, 0)`, accepts a connection and never answers, points `origin` at it, and drives the real `CheckoutFreshnessGitCommandRunner` with a 250 ms bound — a genuine unresponsive transport, not a boolean stub, and it asserts the checkout is left untouched.
- **Finding logic untouched**: the `AutomationStalledWorkCommand` change against the merge base remains additive apart from the single runner-factory line; no collector, kind, or threshold moved.
- **Scope**: 9 files, 828 insertions, 5 deletions vs merge base `728989c6`, confined to the contract's target paths, docs mirrored across all four files.
- **Tests**: focused `AutomationStalledWork*` at this head — **119 passed / 0 failed** (up from 118, matching the added blackhole test).
- **CI** at `8ef637f5`: `Build and test (source contract)` pass, `npm package dry-run (no publish)` pass.
- **`git diff --check`** across `728989c6..8ef637f5`: clean.
- **Head** `8ef637f5fd4946024ce762a9bef6f2597a00a5e3` confirmed unchanged before and after this pass; `Closes #1575` still present and matching.
- **`design_alignment_checked=true`** against `intents/intent-cli/intent-tree/means/03-state-and-audit-strategy.md`. Both repairs move in that node's direction: the answer now states what it was computed from *and* what it could not establish, and the survey no longer claims more than its evidence supports.

Passing tests remain **necessary, not sufficient** — both prior findings were invisible to the suite as it stood, and the fix for each brought its own test with it.

<sub>Re-review evidence collected at head `8ef637f5` with a build of this branch. The unresponsive-remote case used a listener bound to 127.0.0.1 that accepts and never replies, stopped and verified gone afterwards; the stale/current cases used a bare `origin` and clone under my scratch directory. No real repository, ref, or remote was modified.</sub>

PR #1589, opened 2026-08-25
G731: recover undelivered sender-local reports

### LINKED ISSUE #1586: G731: an undelivered sender-local report has no working recovery, because the recovery commands ask whether two roots differ instead of whether the write would succeed
# G731: an undelivered sender-local report has no working recovery, because the recovery commands ask whether two roots differ instead of whether the write would succeed

Resolves the consumer report in #1585.

## Goal

An undelivered sender-local report can be recovered from the host, by a command that is named.

## Why This Slice Exists Now

A report stuck as `undelivered` with `report-routing-root-write-required` has no working recovery path. `notify reconcile` declines it and points at an unnamed "delivery-level recovery path". `notify collect`, run on the host where the events file is demonstrably writable, replays the sender-side sandbox error instead of performing the append.

The reporter's control from the same host shell:

```
touch <host-root>/.intent-cli/events/<d>/.write-probe     # succeeds
grep -c "<task>" <host-root>/.intent-cli/events/<d>/<t>.jsonl   # 0 — nothing appended
```

Their reports have been stuck since the G26 era of that domain.

## Current Observed State

Located on current `main`, not taken on trust:

```csharp
// NotifyCommand.cs:1292
var senderLocalReport = isReport && !PathsEqual(resolvedReportRoot, options.RoutingRoot!);

// NotifyCommand.cs:1504
if (delivery.ReaderPath is not null && senderLocalReport && options.Write)
    →  "report-routing-root-write-required"
       "...cannot be written from this sandboxed seat..."
```

**The gate is structural, not empirical.** It asks whether the report root differs from the routing root. It never attempts the write and never probes writability — it asserts a sandbox constraint it has not measured.

So when the host runs the documented recovery:

```
notify collect --report-root <child> --routing-root <host> --write
```

those two roots differ *by definition*, `senderLocalReport` is true, and the same refusal fires on a machine where the write would succeed. **The recovery command can never succeed for exactly the entries it exists to recover.**

The other half, `NotifyCommand.cs:614`:

> Sender-local report for task '…' is 'undelivered', not delivered; the local handoff remains available for its delivery-level recovery path.

No command is named for that path.

## This is the unbuilt half of G719

G719 shipped sender-local reporting so a sandboxed child could write locally and defer host synchronisation. The child half works — the dry-run even says *"Sender-local report handoff persisted …; no host-root write was required"*, which reads as the intended design. Nothing on the host completes it for `undelivered` entries, so delegation reports from sandboxed seats are one-way.

The design thread has been hand-carrying implementation-seat reports for weeks because of this. That is why it went unreported as a defect for so long — a human was absorbing it.

- The child-side refusal is correct **when a sandbox genuinely blocks the write**. This unit completes the host half; it does not relax the child half.
- `.intent-cli` sender-local storage is correct and is not the problem.
- The reporter's environment is a supported shape: separate host routing root and child report root, herdr-only, preflight ready.
- Corroboration on the intent-cli host: `sekiban-as-a-service` carries three outbox entries with this exact `delivery_error`, all undelivered. Their layout differs from the reporter's, so that is evidence the state occurs, **not** a reproduction of their case.

## Target Repo / Path / Part

- Target paths: `src/IntentSystem.Cli/Commands/NotifyCommand.cs`, the notify documentation in `docs/en/` and `docs/ja/`, and the notify tests under `tests/`
- Part: what `notify collect` and `notify reconcile` do with a sender-local report whose `delivery_state` is `undelivered`

## In Scope

- Making the host-side recovery complete for undelivered entries.
- Deciding the refusal by whether the write can succeed, not by whether two paths differ.
- Naming the recovery command wherever the current text gestures at one.

## Out Of Scope

- Relaxing the child-side refusal when a sandbox genuinely blocks the write.
- Changing sender-local storage.
- Auto-recovering entries without an explicit command.

Self-contained. The two code sites, the structural-versus-empirical distinction, the reporter's writability control, and the relationship to G719 are all above. #1585 carries the reporter's full reproduction.

## Acceptance Criteria

- [ ] An undelivered sender-local report can be recovered from the host by a **named** command that completes and marks the entry delivered. Demonstrated on an entry that is genuinely `undelivered` with `report-routing-root-write-required` — not a fresh one.
- [ ] The refusal is decided by whether the write **can succeed in the current execution context**, not by whether two paths differ. A host that can write must not be told it cannot.
- [ ] The sandbox message is emitted only when a sandbox constraint was actually encountered. Asserting an unmeasured constraint is the specific defect here and must not survive in a narrower form.
- [ ] If `reconcile` still declines undelivered entries, it **names** the command that recovers them.
- [ ] Child-side behaviour is unchanged: a seat that genuinely cannot write the host root still retains its report locally and still reports the routing fault.
- [ ] Verification is emitted output from **both** sides: the child producing an undelivered entry, and the host recovering it, with the host events file shown gaining the line.
- [ ] State whether any other command decides a sandbox refusal by comparing paths rather than attempting the operation. Report what was surveyed, **including the surfaces found correct**.

## Verification

- Paste the child-side `notify report` producing the undelivered entry.
- Paste the host-side recovery completing, and the events file before and after.
- Paste `reconcile` on an undelivered entry showing it names a command.

## Related Links

- Consumer report: https://github.com/J-Tech-Japan/intent-system/issues/1585
- The unit that shipped the child half: G719, PR #1562

## Knowledge Maintenance

Node 08 records that a refusal which infers a constraint from the shape of its arguments will fire in contexts where the constraint does not hold — here it fired on the host, the one place the recovery was supposed to work.

## Base Branch Policy

Do not fix this by special-casing `collect` to skip the check. The defect is that the check answers a different question than the one being asked; narrowing its blast radius leaves the same false assertion in place for the next caller.

Attempt the operation, or probe the specific path, and say what was actually observed. The current message tells an operator on an unsandboxed host that they are in a sandbox. Whatever replaces it must be a statement about something measured.

Note what this unit inherits: G719's review proved the child-side boundary by revoking write access and probing, which was the right question for the child and left the host path unexercised. Do not accept this one on child-side evidence.

tomohisa:
## Summary

- Closes #1586.
- Make sender-local report delivery capability/outcome-based for both `notify report` and `notify collect`: differing report and routing roots no longer refuse delivery before the append is attempted.
- Preserve the sender-local outbox only after an observed I/O/access failure, and expose the exact delivery-level `notify collect --routing-root <host> --report-root <role-work-root>` command from `notify reconcile` as `recovery_command` and in the summary.
- Add a test-owned split-root regression that makes the host root read-only, creates a genuinely undelivered sender-local entry, then proves host collection increases the destination events file by one line and marks the outbox delivered.
- Update English and Japanese orchestration guidance.

## Verification

- `NotifyG719Tests`: 4 passed.
- notify-focused tests: 166 passed.
- docs-focused tests: 248 passed.
- Full `IntentSystem.Cli.Tests`: 5,184 passed, 1 skipped; two unrelated `PackagedInvocationSmokeTests` dotnet-tool smoke tests still fail in the packaged invocation surface.
- A source survey found no other report-delivery refusal based on comparing roots. Remaining root comparisons are descriptive storage/sync metadata, reconciliation-command selection, topology identity validation, or launch-envelope comparison; the delivery capability surface is the actual `NotifyEventWriter.Append` outcome.


tomohisa (commented):
# Semantic review — G731 / PR #1589 @ `197e135416a54a00285f81fca8440d8b49853846`

**Verdict: PASS** (delivered as `COMMENTED`; `--approve` is rejected on a same-account-owned repo — the notify report carries the workflow verdict).
Head verified `197e1354` before and after the pass. Base `main` @ `3debf8ee`. `git diff --check` clean. `Closes #1586` present and matches the linked source issue (G311).

Execution unit **G731**, packet `.intent-cli/issues/G731` (`packet.yaml`, `implementation.md`, `review-context.md`, `github-body.md`); `review closeout-plan` → `ready`, `gaps: []`.

`design_alignment_checked=true`.

---

## 1. The two facts this review was told to press on

### (a) Real append proof, from a genuinely stuck entry — **satisfied**

`CollectRecoversGenuinelyUndeliveredSenderLocalReportByObservedHostWrite_G731` does not simulate the stall. `MakeHostReadOnly()` drops the write bit on `HostRoot` and every directory beneath it (real `File.SetUnixFileMode`), and the child's report is launched as a **separate process** running the exact generated command (`RunExactCommand` → seat shell). The resulting entry is genuinely `undelivered` with `report-routing-root-write-required`, produced by a real `UnauthorizedAccessException`, not by a structural predicate.

The destination is the canonical host events file, not a synthetic reader: the topology is written with `NotifyEventWriter.RelativePathFor(Domain, Team)` = `.intent-cli/events/<domain>/<team>.jsonl` — the same file the reporter measured with `grep -c` returning 0. The previous G719 test pointed at `external-reader.jsonl`; this is a real strengthening.

Append proof is a before/after count on that file:

| stage | `.intent-cli/events/intent-cli/intent-cli-dev.jsonl` |
|---|---|
| after the stuck child report (host read-only) | `Assert.Empty(...)` → **0 lines** |
| after `notify collect --routing-root <host> --report-root <seat> --write` | `Assert.Equal(beforeLines + 1, ...)` → **1 line** |

and the single line is parsed and asserted (`kind: completion`, `unit: G731-recovery`), the outbox flips to `delivered`, and a follow-up `reconcile` returns `reconciled: true`. That is the item the packet said "cannot be argued with", and it is present.

I captured the two summaries the run actually emits (instrumented in a throwaway worktree, then reverted):

```
Could not append notification to external role 'orchestration' through recorded reader
'…/.intent-cli/events/intent-cli/intent-cli-dev.jsonl' in the current execution context:
Access to the path '…/.intent-cli/events/intent-cli/intent-cli-dev.jsonl' is denied.
The attempted host-root write failed, so the sender-local report handoff is retained at '…/report-outbox.jsonl'.
This measured write failure is a delegation-level routing fault, not an implementation-seat stall.
```

```
Delivered collect to external logical role 'orchestration' … Sender-local report handoff persisted under
'<seat>'; the host routing event was appended at '<host>'. Host pending and continuation state at '<host>'
remains for orchestration reconciliation.
```

The refusal now quotes the OS error and the path it actually tried. It reports an observation, not a deduction — AC #3 satisfied on emitted output, and the "sandboxed seat" claim is gone rather than narrowed.

### (b) No collect-only special case — **satisfied**

The refusal at old `NotifyCommand.cs:1504` is **deleted outright**, not conditioned. `ExecuteDelivery` is shared: `isReport` is `report || collect` (`:1295-1296`), so both callers now reach `NotifyEventWriter.Append` and are decided by its outcome. `senderLocalReport` survives only as (i) advisory text, (ii) a label on an *already-measured* failure (`routingWriteFailure` sits inside `catch (IOException or UnauthorizedAccessException)`), and (iii) the pre-existing `!senderLocalReport` boundary on host pending writes — which is ADR 0008's authority split, not a sandbox claim.

I did not take that from reading. Mutation matrix, run in a throwaway worktree pinned at `197e1354`, restored clean afterwards:

| # | mutation of the real artifact | expected | result |
|---|---|---|---|
| C | control — untouched tree | pass | **4/4 pass** |
| M1 | re-insert the pre-fix structural refusal verbatim | fail | **FAIL** (G731 test) |
| M2 | re-insert it as a **collect-only special case** (`&& !OperationCollect`) | fail | **FAIL** — trips at `NotifyG719Tests.cs:215`, the report-side measured-message assertion, i.e. the guard catches exactly the forbidden fix |
| M3 | keep `eventAppended = true` but drop the actual `Append` (successful-looking output, nothing written) | fail | **FAIL** (G731 **and** the G719 child-half test) |
| M4 | revert `reconcile` to the unnamed "delivery-level recovery path" text | fail | **FAIL** |
| M5 | drop `--report-root` from the emitted recovery command | fail | **FAIL** |

M2 and M3 are the two failure modes the packet named in advance. Both are caught.

---

## 2. Remaining acceptance criteria

- **AC1 — named command completes and marks delivered, on a genuinely undelivered entry.** Test above; `delivered: true`, `event_appended: true`, outbox `delivered`.
- **AC2 — decided by whether the write can succeed.** Deletion of the pre-append gate; verified by M1/M2.
- **AC4 — `reconcile` names the command.** Verified against **live stuck production data**, not only the fixture. This machine carries seven real entries `undelivered` / `report-routing-root-write-required` (sekiban-as-a-service, 2026-08-21/22 — the same corroboration class the packet cites; the packet recorded three, seven are present today). Read-only `--dry-run`, PR-head binary vs. installed pre-fix binary, same entry:

  ```
  PR head   recovery_command: intent-cli notify collect --domain sekiban-as-a-service \
              --team sekiban-as-a-service-orch --task-id design-1019-gap-audit-20260822 \
              --write --routing-root <host> --report-root <seat>
  installed recovery_command: null
            summary: "... the local handoff remains available for its delivery-level recovery path."
  ```

  The sentence now ends in something runnable, on the exact class of entry that has been stuck for weeks. No live state was mutated: `--dry-run`, and the undelivered branch returns before any write.
- **AC5 — child half unchanged.** `ExternalReaderRetainsSenderLocalHandoffWhenHostRootIsDenied_G719` is untouched and still passes: real chmod denial → `report-routing-root-write-required`, outbox `undelivered`, reader file byte-identical, host pending still open. M3 also breaks it, so it is not vacuous.
- **AC6 — both sides emitted.** Child (stuck report) and host (recovery) are both in the one regression, with the events file before and after.
- **AC7 — sibling survey, including surfaces found correct.** I re-ran the survey rather than accepting it. Every surviving root comparison in `NotifyCommand.cs`: `:1741 ReportStorageMode` and `:1748 HostStateSync` (descriptive metadata), `:1983 BuildReconciliationCommand` (command *selection*), `:1298 senderLocalReport` (advisory + post-failure labelling + the ADR 0008 pending-write boundary). Outside notify, the only other match is `QueueStateForwardDeltaAnalyzer.PacketPathsEqual` (packet-path equality, unrelated). **No other command decides a refusal by comparing roots.** The PR body's claim is accurate.

  I also checked the three call sites that still build the collect command without `--report-root` (`NotifyMeasuredSupervisor.cs:1435`, `NotifyReportOutboxStore.cs:50`, `NotifyCommand.cs:754`). Those read the outbox at `ResolvePath(routingRoot, …)`, i.e. report root == routing root there, and `collect` defaults its report root to the invoking repo root — so the two-argument form stays correct and is not a half-named command. `reconcile` is the one surface holding both roots, and it is the one that now names both.

---

## 3. Design alignment (sources)

- `intents/intent-cli/intent-tree/means/04-worker-interface-strategy.md:225` — G719's recorded decision: `report-routing-root-write-required` is surfaced **as a delegation-level routing fault, not a seat pane failure**, and host sync is deferred to a host-owned idempotent reconcile path. The new failure summary keeps that framing verbatim, and the `!senderLocalReport` gate keeps host pending/continuation on the reconcile path. Honoured.
- `intents/intent-cli/intent-tree/means/04-…:227` — the record that G719's review proved the boundary from the negative side only. That is exactly the gap this unit closes, and this review deliberately did not accept child-side evidence.
- `intents/intent-cli/intent-tree/means/08-agent-message-orchestration.md:275` — "notify は signal surface であり、merge・label・publish・queue transition を行わない". The change appends one event and updates the sender-local outbox; no lifecycle authority is added.
- `docs/adr/0008-sender-local-report-reconciliation.md` — "Using `notify collect` as reconciliation" is a *rejected alternative*: collect is **delivery recovery for an undelivered entry**, reconcile is host-state closure for a delivered one. The PR implements precisely that split — reconcile still declines undelivered entries and hands off to collect. ADR text at `:84` remains accurate under the new behaviour (the outcome is still emitted, now only when measured).
- `docs/en/12-…:557ff` / `docs/ja/12-…:472ff` — updated in both mirrors, and the EN/JA pair now agrees on the same three facts (outcome-based decision across both callers, the named collect recovery, reconcile returning `recovery_command`). The stale clause "fails before appending … and retains the local handoff" was removed from both, not just EN.

Out-of-scope boundaries from the issue body were **not** crossed: the child-side refusal is intact, sender-local storage is untouched, and nothing auto-recovers without an explicit command.

Node 08's knowledge write-back (`knowledge_updates.intent_tree`) targets a **host-owned** path outside this repository — correctly absent from the implementation PR (G287); it belongs to closeout.

---

## 4. Tests — necessary, not sufficient

Executed at the exact head, in a dedicated worktree, in-tree (no `--artifacts-path`):

- `NotifyG719Tests` — **4/4 passed**.
- `PackagedInvocationSmokeTests` — **11/11 passed, 0 skipped**.
- Full `IntentSystem.Cli.Tests` — **5,186 passed, 1 skipped, 0 failed** (1m46s).

On the two packaged smoke failures reported locally in the PR body: **they do not reproduce here**, and they are not accepted as unexplained. `PackagedInvocationSmokeTests` is fully green in this checkout and the required CI check "Build and test (source contract)" is green at this head. The residual signal is environment-local to the author's seat (this machine runs concurrent agent sessions against shared MSBuild/VBCSCompiler daemons; `obj` contention in the packaged-invocation surface is a known, previously reported symptom there). Nothing in this diff touches the packaged-tool surface. Not a regression.

Per policy, the green suite is paired with the packet evidence above rather than standing on its own.

---

## 5. Non-blocking notes (no change required for this merge)

1. `NotifyReportOutboxStore.BuildCollectCommand` interpolates `routingRoot` / `reportRoot` unquoted, while its sibling `BuildReconciliationCommand` uses `ShellQuote`. A root containing a space would emit a command that cannot be pasted. Pre-existing pattern, extended; ordinary roots are unaffected.
2. On the host, a measured collect failure is still labelled `report-routing-root-write-required`. Literally accurate (a routing-root write was required and failed) and the prose no longer claims a sandbox, so this is cosmetic only.

Neither is an implementation finding against a G731 acceptance criterion.

PR #1598, opened 2026-08-26
G734: add live-safe supervision state shrink

### LINKED ISSUE #1592: G734: supervision jsonl grows toward the push limit with no sanctioned way to shrink it, and each record is far larger than the event it records
# G734: supervision jsonl grows toward the push limit with no sanctioned way to shrink it, and each record is far larger than the event it records

Resolves the consumer report in #1590.

## Goal

Supervision state stops heading for the push limit, and an operator whose state has already grown has a sanctioned command to shrink it.

## Why This Slice Exists Now

The reporter's file reached **52,604,011 bytes**. GitHub warns at 50 MB and rejects at 100 MB, so continued growth ends in a host repo that cannot be pushed at all.

Measured on the intent-cli design host rather than taken on trust — **73.7 MB across 8 files**:

```
47.7 MB  sekiban-dcb-ts/sekiban-dcb-ts-orch/stalls.jsonl
12.7 MB  sekiban-dcb-ts/sekiban-dcb-ts-orch/cycles.jsonl
 6.6 MB  intent-cli/intent-cli-dev/stalls.jsonl
 3.9 MB  sekiban-wasm-runtime/sekiban-wasm-runtime-orch/cycles.jsonl
```

**The size is not a long history.** That 47.7 MB file holds only **10,063 records** — an average of **4,970 bytes each**, the largest **37,539 bytes**.

Rotation alone would start a fresh file that fills at the same rate.

## Where the bytes actually are

Breakdown of the largest record:

```
stall                    32,610 of 37,539 bytes
  ├─ evidence                    15,684
  ├─ registration_lookup          9,672
  ├─ consulted_observations       5,659
  └─ summary                        380
```

`evidence` carries prose that **does not vary between records**. The full sentence

> a recorded herdr seat is registered only when the matching agent-list entry is running at the recorded workspace and pane

is re-embedded in every record that cites it. An invariant definition is being stored per occurrence rather than referenced.

Density varies with which findings fire: `intent-cli-dev` stalls averages 1,638 bytes over 4,221 records. So this is not uniform overhead — it is specific findings carrying large fixed payloads.

## And there is no sanctioned action

`intent-cli notify supervise --help` exposes no rotate, compact, prune or gc surface. Repository policy correctly forbids hand-editing supervision state when a CLI transition exists — but here none exists.

**An operator facing a 50 MB warning has nothing they are allowed to do.**

## Current Observed State

- Supervision state: 73.7 MB across 8 files on this host; 52.6 MB in one file on the reporter's.
- 10,063 records producing 47.7 MB.
- No rotate/compact surface anywhere in the CLI.
- `cycles.jsonl` has the same append-only shape, already 12.7 MB for one domain.

- The policy forbidding hand-edits of supervision state is correct and stays. It is why this needs a command.
- Supervision records are evidence. Shrinking must not mean quietly losing the audit trail.
- `.intent-cli/runs/G361.provider.jsonl` and siblings (14.1, 12.3, 8.3 MB) share the append-only shape but are **not** supervision state and not this unit's target.

## Target Repo / Path / Part

- Target paths: `src/IntentSystem.Cli/Commands/` (the `notify supervise` surface and the stall-record writer), `docs/en/`, `docs/ja/`, and their tests
- Part: what the supervise daemon writes per stall record, and the sanctioned way to shrink supervision state without corrupting it

## In Scope

- A sanctioned way to shrink supervision state.
- Reducing per-record size, specifically invariant text stored per occurrence.
- Documenting the growth behaviour and the shrink path.

## Out Of Scope

- `.intent-cli/runs/*.provider.jsonl`.
- Changing what supervision detects.
- Relaxing the no-hand-edit policy.

Self-contained. The reporter's size, this host's independent measurements, the per-record breakdown, the invariant-prose finding, and the absent command surface are all above.

## Acceptance Criteria

- [ ] An operator whose supervision state has grown large has a **sanctioned command** to shrink it. Guidance to edit the file by hand is not acceptable — the policy forbidding that is correct.
- [ ] Whatever is archived or discarded, the audit trail says what happened. Supervision state is evidence; silently dropping records trades a size problem for an evidence problem.
- [ ] **The per-record size is addressed, not only the file size.** 10,063 records producing 47.7 MB is not a history problem, and rotation alone refills at the same rate. Show the new average bytes per record against the measured **4,970**.
- [ ] Invariant text is not re-embedded per record. Where a record cites a fixed definition, it references it. State how much of the reduction came from this versus anything else — **measured, not estimated**.
- [ ] The supervisor keeps working across a shrink. Demonstrate a rotation or compaction **while a supervise process is running** and show its next cycle records correctly. That is the state every affected host is actually in.
- [ ] `cycles.jsonl` is covered by the same mechanism or explicitly excluded with a reason. Same shape, already 12.7 MB for one domain.
- [ ] State whether existing oversized files on a live host can be brought down by the new mechanism, or need a one-off path. **The reporter has a 52 MB file today and needs to know which.**

## Verification

- Paste before/after sizes and the new average bytes per record.
- Paste the shrink performed while a supervisor is running, and the next cycle recorded.
- Paste what the audit trail says about the shrink.

## Related Links

- Consumer report: https://github.com/J-Tech-Japan/intent-system/issues/1590

## Knowledge Maintenance

Node 03 records that an append-only audit log with no sanctioned way to shrink it is a deadline rather than a design — and that here the size came from record density, not history, so the obvious remedy would not have held.

## Base Branch Policy

The reporter asked for rotation. That is the obvious remedy and, on its own, it would not hold — the file would refill at the same rate, and the next report would arrive from the same host in a few weeks. Measure before choosing.

Be careful with the evidence field. It exists so a stall finding can be understood after the fact, and shrinking it into an opaque code would trade one unusable state for another. Referencing a fixed definition is different from deleting it; a reader must still be able to resolve what a record means.

If some records genuinely need to be large, say which and why rather than compressing everything uniformly.

tomohisa:
Closes #1592

## Summary

- Keep the sanctioned `intent-cli notify supervise shrink --domain <domain> --team <team> --write --format json` path live-safe and readable.
- Validate every retained `evidence_ref` and its manifest dependency under the shared supervision lock before staging or writing any target.
- Replace the manifest/stalls/cycles sequence with a durable staged transaction journal containing before/after hashes; restart recovery records `recovered-completed`, and unexpected target state records `aborted` without overwriting it.
- Preserve every stall and cycle record, retain the human-readable evidence manifest, and append an audit naming completed/recovered/aborted outcomes plus archive/discard/rotation counts.

## Accepted review findings addressed

1. Constructed unknown-reference, missing-definition, and unreadable-manifest cases fail with the named `shrink-validation-failed` error before any manifest, `stalls.jsonl`, `cycles.jsonl`, or audit write. The tests assert byte-identical state and no audit line or transaction journal.
2. Fault injection after manifest, stalls, and cycles replacement, and immediately before final audit append, leaves a staged journal. A fresh canonical shrink invocation verifies and completes the missing work, proves readable state and unchanged record counts, and appends `recovered-completed`; a target-conflict test proves `aborted` accounting without silent overwrite.
3. The committed verification document is now a stable renderer/contract. The focused tests emit `density.json` and `live.json`; `eng/emit-g734-verification.sh` binds those values to the PR head SHA and CI run ID, validates the numeric/live invariants, counts the TRX results, and uploads one `g734-verification.json` source-of-truth artifact plus a portable SHA-256 sidecar naming only `g734-verification.json`. CI copies both files into an unrelated consumer directory and runs the standard checksum checker there. Timestamp-dependent live bytes are taken from that exact artifact rather than copied into a self-invalidating document.

## Exact-head verification

The final branch head is `1dd4456a457bc4ace786f42efd4273844df9457d`. GitHub Actions run [32932228146](https://github.com/J-Tech-Japan/intent-system/actions/runs/32932228146) was green at that exact head. Download its `g734-verification` artifact; its `source.head_sha` is the branch head above and its `source.ci_run_id` is `32932228146`.

- Source-contract tests: `5,538` passed, `1` not-executed, `0` failed; the focused `NotifySupervisionShrinkG734Tests` class contributed `15` passing tests. The npm package dry-run also passed.
- Density: `50,735,552 -> 48,561,944` bytes for `10,063` records; `5,041.79 -> 4,825.79` bytes/record versus the `4,970` baseline.
- Direct invariant rewrite saving: `2,173,608` bytes, with `other_record_bytes_saved=0`; diagnostic counters report `2,435,246` literal bytes removed and `322,016` reference-key bytes added. Those counters are occurrence diagnostics, while the direct before/after byte delta is the saving claim.
- Live proof from the emitted artifact: `supervisor_state=running`, `cycles=1->2`, total `10,986,486 -> 8,812,878` bytes, and `20,127 -> 20,127` records. `stalls.jsonl` was `10,985,655 -> 8,812,047` bytes; `cycles.jsonl` was `831 -> 831` bytes with its record retained. The next cycle appended after the live replacement.
- The completed audit reports `0` archived, `0` discarded, and `0` rotated records, with both files retained in place. `.intent-cli/runs/*.provider.jsonl` remains out of scope.
- Checksum portability: the uploaded sidecar contains `ee82e1a3780a1e5ebc8b0b95f1045d2021c0a61917a7ad99ccc1b5c398104723  g734-verification.json`. CI copied the pair into a fresh consumer directory and passed `sha256sum --check`; an independent download of the same pair also passed `shasum -a 256 -c` and `sha256sum --check` outside the CI workspace.
- `git diff --check`: clean.


tomohisa (commented):
## Request update — G734

Reviewed commit `54a0aa44c915f6bcde81b63773fe78f297b088f0`.

The sanctioned `notify supervise shrink` surface, existing-file rewrite, shared directory lock, readable manifest, `cycles.jsonl` coverage, and the genuine external-supervisor test are all present. In particular, the exact-head CI run's `ShrinkWrite_ReportsRunningExternalSupervisorAndNextCycleAppends` passed with `supervisor_state=running` and `cycles=1->2`. The command is therefore not rotation-only. `git diff --check` is clean and both required CI checks are green.

However, the following are blocking acceptance gaps.

1. **[P1] Fail closed before shrinking unreadable evidence references.** `NotifySuperviseShrinkCommand.Execute` first detects an unreadable state through `Read`, but still calls `Shrink`. Inside `Shrink`, the code only parses the definition manifest and then reserializes raw stall events; it never resolves every existing `evidence_ref` before replacing files and appending the audit. Thus a legacy record with `evidence_ref: "unknown"` (or the shipped key with a missing manifest) makes the normal reader reject the state, but `shrink --write` can still return success, rewrite `cycles.jsonl`, and append an audit saying all records were retained. That does not satisfy the readable-evidence contract. Validate all retained records/references under the lock and return a named failure with **no writes** when any cannot be resolved. Add the required constructed counterexample for an unreadable/missing reference, asserting nonzero exit, byte-identical state, and no audit append.

2. **[P1] Make a crash between data replacement and audit recoverable/accounted for.** The write ordering is manifest, `stalls.jsonl` replacement, `cycles.jsonl` replacement, then `shrink-audit.jsonl` append. Each individual JSONL replacement is atomic, but the operation is not: a crash or I/O failure after the first replacement can leave a compacted data file with no durable shrink outcome, while the command merely returns an error. The task requires every compaction outcome to be accounted for and explicitly calls for crash/no-silent-loss coverage. Introduce a recoverable transaction/audit protocol (or equivalent) and a fault-injected test covering failures between replacements and before audit append, then prove restart leaves readable state and durable accounting for the completed/recovered outcome.

3. **[P1] Correct the committed density evidence to the exact-head measurement.** `docs/g734-supervision-shrink-verification.md` and the PR body claim `before_bytes=50,804,891`, `after_bytes=48,631,283`, and averages `5,048.68 -> 4,832.68`. The uploaded TRX for exact-head CI run `32925101531` records the executed density test as `50,814,959 -> 48,641,351` bytes and `5,049.68 -> 4,833.68` bytes/record (the 2,173,608-byte invariant saving is unchanged). The document calls itself the emitted verification record, so these are not harmless rounded values; the mandatory measured before/after evidence is stale. Update the durable verification/PR summary from the final exact-head run and ensure its calculation is reproducible. The same artifact reports 5,200 passed / 1 skipped, rather than the PR's 5,198 claim.

### Acceptance mapping

- Sanctioned command and discoverable English/Japanese/operator guidance: implemented.
- Existing `stalls.jsonl` and `cycles.jsonl`: intended in-place, same-lock coverage is implemented, but unreadable input and crash recovery are not yet safe enough to pass.
- Density and invariant-text split: implementation measures the requested fields, but the published exact-head proof is inaccurate.
- Live running supervisor and subsequent append: independently confirmed in exact-head CI (`running`, `cycles=1->2`).
- Audit/no silent loss: normal-path audit is present; the unresolved-reference and interrupted-operation paths above prevent approval.

tomohisa (commented):
## Request update — G734 re-review

Reviewed commit `444aaee3219f9d2fd097246c79ed0d955b9aaf5c` against the G734 packet, issue #1592, and the exact-head CI artifact for run `32929396702`.

### Closed findings independently verified

- **Unreadable references:** the shared-lock preflight resolves retained references before staging. The exact-head CI passed all three constructed no-write cases: unknown reference, missing canonical definition, and unsupported/unreadable manifest. Each asserts exit 1 with `shrink-validation-failed`, byte-identical stalls/cycles/manifest state, no audit line, and no transaction journal.
- **Replacement/audit crashes:** the staged journal carries before/after hashes. Exact-head CI passed each injected boundary — after manifest, after stalls, after cycles, and before audit — followed by readable recovery, retained records, `recovered-completed` accounting, and journal removal. The external-target conflict case passed with a durable `aborted` outcome and no overwrite.
- **Density and live safety:** independently recalculated from final-head TRX: `50,735,552 / 10,063 = 5,041.7919` and `48,561,944 / 10,063 = 4,825.7919` bytes/record; the latter is 144.2081 below the 4,970 baseline. The 2,173,608-byte net saving exactly equals invariant-text savings (2,435,246 literal bytes removed minus 322,016 reference bytes) with 0 other record bytes saved. The real external-supervisor test passed with `supervisor_state=running` and `cycles=1->2`.

The sanctioned command, existing-file path, readable manifest, `cycles.jsonl` boundary, English/Japanese guidance, and provider-log exclusion are present. `git diff --check` is clean; exact-head CI has 5,538 passed / 1 skipped across projects (CLI: 5,208 passed / 1 skipped), npm dry-run is green, and no automated-review signals are pending. Tests are necessary but not sufficient; the packet's state-and-audit contract and its referenced `03-state-and-audit-strategy.md` are otherwise consistent with the implementation. `Closes #1592` is present.

### Blocking implementation finding

1. **[P1] Correct the claimed final-head evidence chain.** The required exact-head evidence has not yet been made durable. `docs/g734-supervision-shrink-verification.md` says its “final implementation” TRX is run `32929074724`, but that run is for `70c4a16dfe01c32aba3fc19edb4cc0b379e71ca7`, not this reviewed head. The PR body names the correct final run `32929396702`, but both it and the document quote the prior run's live totals (`10,986,486 -> 8,812,878`, cycles `831` bytes); the actual assigned-head TRX is `10,986,485 -> 8,812,877`, cycles `830` bytes. This directly fails the task's required reproducible exact-head verification across docs, PR body, artifact, byte counts, averages, savings, and test count. Regenerate or restructure the durable live transcript so it identifies the final-head run and cannot drift with timestamp-dependent fixture bytes; do not claim “exact-head” evidence until all quoted values come from that run.

This is an **implementation-finding** against the issue Verification requirement and the packet/review-context requirement for measured, durable before/after and live-supervisor evidence. It is the sole remaining blocker in this re-review.

tomohisa (commented):
## Request update — G734 final re-review

Reviewed commit `1d31db6490e6d4336f257a6b1048e59c367135f4` and independently downloaded the `g734-verification` artifact from exact-head CI run `32931300767`.

### Exact-head and functional evidence confirmed

- The artifact binds `source.head_sha=1d31db6490e6d4336f257a6b1048e59c367135f4` and `source.ci_run_id=32931300767`, matching the reviewed PR head and run.
- It agrees with the exact TRX: 5,538 passed / 1 skipped / 0 failed across the source-contract projects, 15 focused G734 tests, density `50,735,552 -> 48,561,944` for 10,063 records, and a running external supervisor with `cycles=1->2` and 20,127 retained records. Independent arithmetic gives 4,825.791910960946 bytes/record after shrink, 144.208089039054 below the 4,970 baseline; the 2,173,608-byte direct delta matches `net_record_bytes_saved`, with 0 other record saving.
- The live artifact supplies the intentionally variable transcript rather than copying it into the committed document: running supervisor, total `10,986,486 -> 8,812,878`, stalls `10,985,655 -> 8,812,047`, cycles `831 -> 831`, zero archived/discarded/rotated, and the subsequent cycle append.
- The prior safety findings remain closed in the exact-head evidence: unknown/missing/unreadable references fail named/nonzero before writes; all manifest/stalls/cycles/audit fault boundaries recover with durable recovered/aborted accounting; readable manifest resolution, existing-state compaction, `cycles.jsonl` coverage, shared locking, English/Japanese guidance, and provider-log exclusion remain present. `git diff --check` is clean, both CI jobs are green, no reviewer signals are pending, and `Closes #1592` is present.

### Blocking implementation finding

1. **[P1] Make the uploaded SHA-256 sidecar directly verifiable after download.** The artifact's digest bytes are correct, but its sidecar is not valid in its delivered layout. It contains:

   ```text
   d667bca9…  /home/runner/work/intent-system/intent-system/.artifacts/g734-verification/g734-verification.json
   ```

   After downloading the published artifact, `shasum -a 256 -c g734-verification.json.sha256` fails because that runner-only absolute path does not exist. The producer (`eng/emit-g734-verification.sh`) calls `shasum` on an absolute output path, which produces this non-portable sidecar. Manually comparing the first field to the downloaded JSON's digest is not the promised sidecar verification contract.

   This is an **implementation-finding** against the task's required exact-head artifact integrity proof: emit a relative `g734-verification.json` entry (or equivalent portable checksum manifest) and add a test/CI check that verifies it from the uploaded-artifact directory layout with the standard checksum command. The full packet and issue evidence otherwise pass, but this sole integrity-chain gap prevents a PASS.

tomohisa (commented):
## PASS — G734 final re-review

Reviewed exact commit `1dd4456a457bc4ace786f42efd4273844df9457d` against G734's `packet.yaml`, implementation notes, review context, issue #1592, and the referenced State and Audit Strategy intent. This is a same-account `COMMENTED` PASS; GitHub approval is not available to this author.

### Exact-head artifact and portable-integrity proof

- Exact-head CI run `32932228146` is green and binds the downloaded `g734-verification` artifact to `source.head_sha=1dd4456a457bc4ace786f42efd4273844df9457d` and `source.ci_run_id=32932228146`.
- I downloaded the artifact into an unrelated fresh consumer directory. Its sidecar is exactly a relative entry, `ee82e1a3780a1e5ebc8b0b95f1045d2021c0a61917a7ad99ccc1b5c398104723  g734-verification.json`; both `shasum -a 256 -c g734-verification.json.sha256` and `sha256sum --check g734-verification.json.sha256` returned `OK` without path rewriting.
- This is contractual rather than a manual check: `eng/emit-g734-verification.sh` emits the checksum only after changing into the artifact directory, and the versioned CI workflow copies the pair into `mktemp -d` before its standard checksum verification. That CI step passed in the reviewed exact-head job.

### Packet / issue acceptance mapping

- The sanctioned, discoverable `intent-cli notify supervise shrink --domain <domain> --team <team> --write` path, English/Japanese guidance, and existing-file in-place path satisfy the operator-action and 52-MB-live-host criteria.
- The shared lock, readable evidence manifest/reference resolution, and unknown/missing/malformed evidence no-write tests satisfy the evidence-preservation/fail-closed contract.
- Staged manifest/stalls/cycles replacement with durable hashes, recovery to `recovered-completed`, and conflict handling to `aborted` satisfy the audit and crash-boundary accounting requirement. `cycles.jsonl` is rewritten under the same transaction and retains its record.
- Density evidence is deterministic and agrees across document, PR body, artifact, and TRX: `50,735,552 -> 48,561,944` bytes for 10,063 records, `5,041.7919 -> 4,825.7919` bytes/record (144.2081 below the 4,970 baseline). The `2,173,608` direct saving reconciles to the measured invariant rewrite with `other_record_bytes_saved=0`.
- Exact-head live evidence proves an external supervisor was `running`, retained `20,127 -> 20,127` records, compacted `10,986,486 -> 8,812,878` bytes, and appended the next cycle (`1 -> 2`). The committed document correctly leaves timestamp-dependent live byte values to the exact artifact rather than copying a stale transcript.
- `.intent-cli/runs/*.provider.jsonl` was neither read nor changed, honoring the explicit out-of-scope boundary. `Closes #1592` is present.

### Independent verification

- Downloaded exact CI TRX independently: `5,538` passed, `1` not-executed, `0` failed; 15 focused `NotifySupervisionShrinkG734Tests` passed, including all three no-write inputs, four injected crash boundaries, conflict abort, density, and live external-supervisor cases.
- Locally executed `dotnet test tests/IntentSystem.Cli.Tests/IntentSystem.Cli.Tests.csproj --configuration Release --filter FullyQualifiedName~NotifySupervisionShrinkG734Tests`: 15 passed, 0 failed.
- `git diff --check` is clean; the npm dry run is green; no pending automated-review signals were found.

Tests are necessary but not sufficient: the packet's state-and-audit contract and the referenced intent's durable, traceable evidence principle are reflected in the source transaction/audit implementation and the exact-head artifact chain. All G734 issue acceptance criteria and stated out-of-scope boundaries are satisfied.

Issue #1593, opened 2026-08-25
topology move refuses a team whose orchestrator and orchestration share one pane - the standard four-thread layout cannot be moved after a reboot


tomohisa:
## Summary

`session-layer topology move` (G697) cannot express a team whose orchestrator and orchestration roles share one pane — which is the four-thread layout the orchestrator guide itself describes. The whole-team move it exists for is exactly the post-reboot rebuild, and it refuses it.

## Environment

`intent-cli 0.23.2-a6f8a9d-G728`, macOS, herdr-only team. Measured 2026-08-25 during a real post-reboot rebuild (old workspace destroyed by restart, fresh workspace created, same three panes).

## Reproduction

Recorded topology has `orchestrator` and `orchestration` both at `wS:p1` (one pane, two logical roles — the shape our team has run with since bootstrap).

```
$ intent-cli session-layer topology move --domain remote-herdr --team remote-herdr \
    --workspace-id w44 --pane-map wS:p1=w44:p1 --pane-map wS:p2=w44:p2 --pane-map wS:p3=w44:p3 --write
applied: false — "--pane-map maps more than one recorded role to new pane 'w44:p1'; refusing an ambiguous workspace move."
```

There is nothing ambiguous to us: both roles should follow their shared pane. Per-role `topology record` cannot be used instead — it refuses with "Team already records workspace_id 'wS' ... use topology move" — so the two surfaces refer to each other and neither can perform the move.

## What we did

Hand-edited the local gitignored topology JSON to the new workspace/panes, then proved it with `topology validate` (valid). Recorded here so the workaround is not mistaken for a supported path.

## Suggestion

Treat a many-roles-to-one-old-pane mapping as unambiguous when the old pane maps to exactly one new pane — the roles travel together. Ambiguity only genuinely arises if two *old* panes map to one new pane.

tomohisa:
Fixed and merged. Closing.

Shipped as G735, PR #1599, merged `2d77c557`.

## Your analysis was the fix

The guard was on the wrong side of the mapping. `Move` now maps `newPane -> oldPane` and refuses only a repeated new pane arriving from a **different** old pane — exactly the distinction you drew:

> Ambiguity only genuinely arises if two *old* panes map to one new pane.

The genuine convergence guard is retained, not removed.

## The dead end is resolved

[...]

PR #1599, opened 2026-08-26
G735: allow shared-pane topology moves

### LINKED ISSUE #1594: G735: topology move rejects roles that share one pane, and topology record points back at move, so the standard layout cannot be rebuilt after a reboot
# G735: topology move rejects roles that share one pane, and topology record points back at move, so the standard layout cannot be rebuilt after a reboot

Resolves the consumer report in #1593.

## Goal

A team whose roles share a pane can be moved to a new workspace by the canonical command.

## Why This Slice Exists Now

Measured by the reporter during a real post-reboot rebuild. Their recorded topology has `orchestrator` and `orchestration` both at `wS:p1` — one pane, two logical roles, the shape their team has run since bootstrap and the one the orchestrator guide describes.

```
$ intent-cli session-layer topology move --domain remote-herdr --team remote-herdr \
    --workspace-id w44 --pane-map wS:p1=w44:p1 --pane-map wS:p2=w44:p2 --pane-map wS:p3=w44:p3 --write
applied: false — "--pane-map maps more than one recorded role to new pane 'w44:p1';
                  refusing an ambiguous workspace move."
```

And `topology record` cannot substitute:

> Team already records workspace_id 'wS' … use topology move

**The two surfaces point at each other and neither performs the rebuild.** Following the guidance correctly leads nowhere.

`move` exists for exactly this — the post-reboot whole-team rebuild — and refuses it.

## Current Observed State

Located on current `main`, not taken on trust:

```csharp
// SessionLayerTopologyCommand.cs:1380
if (!mappedPanes.Add(newPane))
    → "maps more than one recorded role to new pane ...; refusing an ambiguous workspace move"

// SessionLayerTopologyCommand.cs:1183
    → "Team ... already records workspace_id ...; use topology move"
```

**The guard is keyed on the new pane.** Genuine ambiguity arises when two *different old* panes map to one new pane — there, the CLI cannot know which old pane's roles belong where. Several roles sitting on one old pane that moves to one new pane is not ambiguous: the roles travel with their pane.

The reporter reached the same conclusion and it is what the code shows.

They hand-edited the gitignored topology JSON, proved it with `topology validate`, and **recorded the hand edit explicitly so it would not be mistaken for a supported path**. There is currently no sanctioned way to rebuild that layout, so every reboot costs a hand edit.

## Not reproducible here

None of the five domains on the intent-cli design host records two roles on one pane. Evidence for this unit is the reporter's measurement plus the located source — **not a local reproduction**, and the implementer should construct the failing shape rather than assume it.

- The shared-pane layout is legitimate. The orchestrator guide documents it.
- The record-side refusal ("use topology move") is correct in intent — a whole-team move should not be done role by role.
- The reporter's hand edit was disclosed, not concealed. Treat it as a measurement, not as the fix.

## Target Repo / Path / Part

- Target paths: `src/IntentSystem.Cli/Commands/SessionLayerTopologyCommand.cs`, the session-layer documentation in `docs/en/` and `docs/ja/`, and their tests
- Part: what `topology move` treats as an ambiguous pane mapping, and what a team is told when neither surface can perform its rebuild

## In Scope

- Accepting a move where several roles share one old pane that maps to one new pane.
- Keeping genuinely ambiguous mappings refused.
- Resolving the mutual redirection between `move` and `record`.

## Out Of Scope

- Changing what a role topology means.
- Removing the record-side refusal.
- Auto-detecting the new workspace — the move stays an explicit operator-supplied transition.

Self-contained. The verbatim refusals from both surfaces, the two code sites, the correct distinction between the ambiguous and unambiguous cases, and the disclosed workaround are all above. #1593 carries the reporter's original wording.

## Acceptance Criteria

- [ ] A team whose roles share one pane can be moved by the canonical command. **Demonstrated on a topology that actually has two roles on one pane** — that is the case that fails today.
- [ ] Genuine ambiguity is still refused: two **different old** panes mapping to one new pane remains an error, and that refusal is shown alongside the success. A fix that accepts everything has removed a guard rather than corrected it.
- [ ] The mutual redirection is resolved. Today `move` refuses and `record` says to use `move`; a reader following either surface must reach a command that works.
- [ ] Existing recorded topologies are not invalidated and no migration is asked of anyone. Teams with distinct panes per role behave exactly as now.
- [ ] Verification is emitted output from **both** directions — the shared-pane move succeeding and the ambiguous mapping still refused — with the resulting topology validating afterwards.
- [ ] State whether hand-edited topologies already in the wild match what the fixed command would produce. **The reporter has one and needs to know whether to redo it.**

## Verification

- Paste the shared-pane move succeeding, and `topology validate` on the result.
- Paste a two-old-panes-to-one-new-pane mapping still refused.
- Paste whatever `record` now says when a team already records a workspace.

## Related Links

- Consumer report: https://github.com/J-Tech-Japan/intent-system/issues/1593

## Knowledge Maintenance

Node 08 records that a guard keyed on the wrong side of a mapping refuses the case it was meant to permit — and that here two surfaces each redirected to the other, so following the guidance correctly led nowhere.

## Base Branch Policy

Construct the failing shape. It does not exist on this host and cannot be observed by running the command against any local team — a test written against distinct panes per role will pass today and prove nothing.

Do not fix this by relaxing the guard generally. The many-old-panes-to-one-new-pane case is genuinely ambiguous and must keep failing; the PR should show both outcomes rather than only the one it set out to enable.

Answer the reporter's practical question. They are running on a hand-edited topology right now. Whether it matches what the fixed command produces decides whether they redo it, and they cannot tell from the outside.

tomohisa:
Closes #1594

## Summary

- Correct `session-layer topology move` to treat a pane, rather than each logical role, as the mapping unit. Multiple roles sharing one old pane now travel together to one new pane.
- Preserve the ambiguity guard: two distinct old panes converging on one new pane still fail with a named ambiguity and leave the topology unchanged.
- Keep `topology record` fail-closed for a per-role workspace mismatch, but point it to the now-working whole-team move path and explain the shared-pane rule instead of leaving record and move as a dead end.
- Update the installed recipe and English/Japanese command, orchestration, and compatibility guidance.

## Constructed verification evidence

The focused G735 fixture is intentionally independent of this host's topology. Its actual input is:

```text
workspace_id=wS
orchestrator=wS/wS:p1
orchestration=wS/wS:p1
implementation=wS/wS:p2
review=wS/wS:p3
```

The canonical write with `wS:p1=w44:p1`, `wS:p2=w44:p2`, and `wS:p3=w44:p3` returned `applied=true`, `conflict=false`, and `Moved recorded topology ... from workspace 'wS' to 'w44' with 3 pane mapping(s).`; both `orchestrator` and `orchestration` became `w44:p1`, and `topology validate` returned `valid=true` with no findings.

The constructed ambiguity `wS:p1=w44:p1`, `wS:p2=w44:p1` returned exit code 1, `applied=false`, `changed=false`, and:

```text
--pane-map maps distinct recorded old panes 'wS:p2' and 'wS:p1' to new pane 'w44:p1'; refusing an ambiguous workspace move.
```

The ordinary distinct-pane mapping `wS:p1=w44:p1`, `wS:p2=w44:p2` remains successful and validates.

## Record recovery and hand-edit compatibility

When a record attempt requests a new workspace while the team already records `wS`, the command still refuses per-role repair and names the complete `topology move` command. The fixed move now works for the disclosed reporter shape, so following either surface reaches a usable recovery path.

The reporter’s hand edit in #1593 is compatible with the fixed command: it changes the team workspace from `wS` to `w44` and each role’s workspace/pane fields according to the explicit map, with both shared-pane roles at `w44:p1`. The regression compares those produced fields and emits `redo_required=false`. No redo is needed for that disclosed state; future rebuilds should use the sanctioned dry-run/CAS-guarded move rather than hand-editing.

## Tests

- `dotnet test tests/IntentSystem.Cli.Tests/IntentSystem.Cli.Tests.csproj --configuration Release --filter 'FullyQualifiedName~TopologyWorkspaceMove'`: 9 passed, 0 failed.
- `dotnet test IntentSystem.sln --configuration Release`: CLI 5,212 passed / 1 skipped; all other solution test projects passed, 0 failed.
- `git diff --check`: clean.


tomohisa (commented):
## PASS — G735 review

Reviewed exact commit `5257e59689103b6c3acce0570d10da821e1a391f` against the G735 packet, implementation notes, review context, issue #1594, and `08-agent-message-orchestration.md`. This is a same-account `COMMENTED` PASS; GitHub approval is unavailable to this author.

### Constructed topology evidence

- `TopologyWorkspaceMoveG735Tests.Move_AllowsRolesSharingOneOldPane_AndValidationSucceeds_G735` constructs the required failed-before shape: `orchestrator=wS/wS:p1` and `orchestration=wS/wS:p1`, with implementation/review at `wS:p2`/`wS:p3`. Its write map `wS:p1=w44:p1`, `wS:p2=w44:p2`, `wS:p3=w44:p3` succeeds (`applied=true`, `conflict=false`); both shared roles become `w44:p1`, and `topology validate` returns `valid=true`.
- `Move_RefusesDistinctOldPanesConvergingOnOneNewPane_G735` constructs `wS:p1=w44:p1` plus `wS:p2=w44:p1`, returns exit 1 with the named “distinct recorded old panes” ambiguity, and asserts byte-identical topology. This retains the genuine convergence guard.
- `Move_DistinctOldPanesToDistinctNewPanesRemainSupported_G735` keeps the ordinary distinct-pane path successful and valid.

### Source / recovery / compatibility audit

- `SessionLayerTopologyCommand.Move` now maps `newPane -> oldPane` and rejects only a repeated new pane with a *different* old pane; it does not delete the guard or allow arbitrary convergence. It changes only team/role workspace and pane identities. Existing G697 coverage proves `cwd`, kind, delivery method, reader/frontend, and custom team/role fields remain intact.
- The per-role `topology record` workspace-mismatch refusal remains fail-closed but names the working whole-team move command and the shared-pane/different-old-pane distinction. English and Japanese command, orchestration, and compatibility guidance all state that recovery path; neither weakens record-side safety.
- `Record_MismatchPointsToWorkingWholeTeamMove_AndMatchesDisclosedHandEdit_G735` compares the disclosed semantic output field-by-field: team workspace `w44`, each role workspace `w44`, and panes `{orchestrator, orchestration}=w44:p1`, `implementation=w44:p2`, `review=w44:p3`. Exact-head TRX emits that topology and `redo_required=false`; no redo is required for the reporter's disclosed hand edit.

### Contract conformance and verification

- This satisfies each issue AC: same-old-pane canonical move, retained distinct-old-pane refusal, resolved record/move dead end, compatibility/no migration, both emitted outcomes plus validation, and the hand-edit ruling. `Closes #1594` is present.
- The packet's explicit old-side ambiguity contract and the referenced intent’s topology-move atomic/preservation principle are reflected in both the guard and tests. Out of scope remains respected: role-topology semantics, record-side refusal, and automatic workspace discovery were not changed.
- Independently ran `dotnet test tests/IntentSystem.Cli.Tests/IntentSystem.Cli.Tests.csproj --configuration Release --filter 'FullyQualifiedName~TopologyWorkspaceMove'`: 9 passed, 0 failed. Exact-head CI run `32936046075` is green; downloaded TRX reports 5,542 passed, 1 not-executed, 0 failed, including all four focused G735 tests. `git diff --check` is clean and the npm dry run passed.

Tests are necessary but not sufficient; the source guard, constructed output, recovery guidance, preservation behavior, packet/intent mapping, and issue boundaries were independently checked as well.

PR #1604, opened 2026-08-27
[eval-localllm] G735: implemented by Qwen3.8-Flash-Next Q3_K_XL via OpenCode

tomohisa:
## What this changes

G735: `session-layer topology move` refused any pane map that placed more than one recorded role on one new pane, so a team whose roles legitimately share one pane could never be rebuilt — while `topology record` told that same team to use `topology move`. The guard was keyed on the wrong side of the mapping.

- `src/IntentSystem.Cli/Commands/SessionLayerTopologyCommand.cs`: the move guard now tracks a new-pane → old-pane map instead of a bare new-pane set. Several roles sharing one recorded pane travel together under that pane's single mapping; only two *different* recorded panes collapsing onto one new pane is still refused as genuinely ambiguous, and the refusal now names both old panes. The per-role workspace-mismatch refusal from `record` now explicitly states that roles sharing one recorded pane travel together, so a reader arriving at `record` on a shared-pane team lands on a route that completes.
- `src/IntentSystem.Cli/Commands/TopologyWorkspaceMoveGuidance.cs`: the guide's `pane_map_contract` now states the pane-keyed contract and the preserved ambiguity refusal.
- Docs (`docs/en`/`docs/ja` `08-command-reference.md`, `12-agent-message-orchestration.md`): the topology workspace move sections describe the pane-keyed map and the shared-pane behavior.

## Verification

- `dotnet build src/IntentSystem.Cli/IntentSystem.Cli.csproj -c Release` — succeeded, 0 warnings.
- `dotnet test tests/IntentSystem.Cli.Tests/IntentSystem.Cli.Tests.csproj -c Release` — 5211 passed, 0 failed, 1 skipped. The skip is the pre-existing `TaskingHandoffCommandTests.Execute_TolerantToSubAnalyzerFailure_...` skip, unrelated to this slice.
- New `tests/IntentSystem.Cli.Tests/TopologySharedPaneMoveG735Tests.cs` (3 tests, all pass):
  - Two roles recorded on one pane (`w2X:p1`) plus a second pane move cleanly via dry-run → write; post-move `topology validate` is valid and delivery targets resolve to the shared new pane.
  - Mapping two *different* recorded panes to one new pane is still refused, names both panes, and leaves the record byte-identical.
  - On a shared-pane team, the `record` workspace-mismatch refusal names `topology move` and the travel-together behavior, and the named move then completes against the same fixture — the mutual redirection is closed.

## Answer to the reporter's practical question (hand-edited topology)

**You do not need to redo the hand edit.** The fixed `topology move` changes only the team `workspace_id` and the recorded roles' `workspace_id`/`pane_id` fields and preserves every other field. So a hand-edited topology that passes `topology validate` (`valid: true`) already has exactly the shape the fixed command would produce; the difference was only that the command refused to *produce* that shape before. Two optional checks: run `topology move --dry-run` mapping each recorded pane to itself and confirm `changed: false` (proves the record is move-consistent), and from now on do rebuilds with `intent-cli guide topology-workspace-move` + `topology move` (dry-run first) so the edit is one atomic, CAS-guarded, auditable operation instead of a hand edit.

## What I did not do / could not do

- Did not run any live herdr check or reboot a real standard layout; verification is via the in-process CLI fixtures above, matching the existing G697 test approach.
- Did not touch `.intent-cli/` host state, `intents/`, or run `intent-cli` commands (host constraint for this run).
- Did not add release notes; the packet's target paths were the command source, session-layer docs, and tests.
- Untracked `.artifacts/` in the worktree is generated output and was intentionally left out of the commit.


tomohisa (commented):
# Review — PR #1604 (G735)

**Outcome: request-update.**

The fix is correct and the evidence holds. I re-ran the full suite on your head
and got exactly what you reported: `Failed: 0, Passed: 5211, Skipped: 1`. The
skip is the pre-existing one and you named it correctly.

The guard change is right, including the part that is easy to get wrong: you
kept the refusal for distinct old panes converging on one new pane, rather than
removing the check. The refusal now names both old panes. Your test asserting
that the record-side refusal and the move command form a route that actually
completes is the one that matters most, and it is the one I would have asked
for.

Answering the reporter's hand-edit question directly, in the PR body, is
correct and is what the release notes will need.

One blocker, and it is a coverage gap rather than a defect in what you wrote.

---

## Blocker — two surfaces still describe the old contract

You updated `TopologyWorkspaceMoveGuidance.PaneMapContract`, which is what
`guide topology-workspace-move` renders. Two other surfaces state the same
contract and still carry the pre-fix version.

**1. `src/IntentSystem.Cli/Commands/HerdrOnlyOperatingGuide.cs`**

This feeds `guide orchestrator-thread` through `SessionLayerSections`, so it is
what a seat reads when it is deciding how to perform a move. Two places:

- The `## Topology workspace move (G697)` prose still says only "include one
  mapping for every recorded herdr pane" with no statement of the shared-pane
  rule. Grep confirms: that sentence is present once, and no shared-pane
  sentence exists in the file.
- The `["dispatch"]["topology"]` string has the same omission.

A seat that reads the operating guide, hits the record-side refusal, and
follows the guide will still not learn that roles sharing a pane travel
together. That is the exact confusion this unit exists to remove, so leaving
the guide stale keeps the reported problem alive on the surface most likely to
be consulted.

**2. `docs/en/1.0-compatibility-ledger.md` and `docs/ja/1.0-compatibility-ledger.md`**

Line 231 (EN) still reads `G697` and describes the move without the shared-pane
behavior. The ledger is the compatibility record for `preview-through-1.x`
surfaces; a ledger row that describes behavior the code no longer has is worse
than a missing row, because it is consulted as authoritative.

Update both, and make the ledger row cite `G697/G735`.

Once `HerdrOnlyOperatingGuide` carries the rule, add the corresponding
assertion to `TopologyWorkspaceMoveG697Tests` so the guide text cannot silently
regress — the existing test already asserts other clauses of that contract, so
this is one more `Assert.Contains` in the same place.

---

## Not defects — do not change these

- Your new tests living in `TopologySharedPaneMoveG735Tests.cs`. Three tests,
  correctly scoped, no existing test file modified.
- The refusal wording. Naming both old panes is an improvement over naming
  none.
- Not adding release notes. Correct — that is release-prep's unit, not yours.
- Leaving `.artifacts/` uncommitted. Correct.
- Not running a live herdr check. Correct, and saying so explicitly in the PR
  is the right way to handle it.

## What the next review will check

- The two guide surfaces and both ledger files.
- That the EN and JA text say the same thing as each other and as
  `PaneMapContract`.
- The new `TopologyWorkspaceMoveG697Tests` assertion.
- The final line of a fresh full suite run.
- That the head SHA has not moved between your evidence and the PR.

Tests passing is necessary, not sufficient. State one packet contract clause
and one behavior the diff changed, in your own words, when you re-submit.

tomohisa:
## Review follow-up — blocker addressed on `364c7f0e`

### What changed (blocker only; not-defects list untouched)

1. **`HerdrOnlyOperatingGuide.cs`** — both named surfaces now carry the shared-pane rule in the same words as `PaneMapContract`:
   - `## Topology workspace move (G697)` prose: after "include one mapping for every recorded herdr pane" it now states — roles that share one recorded pane travel together to that pane's new pane, and mapping two different recorded panes to one new pane stays refused as genuinely ambiguous.
   - `["dispatch"]["topology"]`: the workspace-rebuild sentence now carries the same two clauses.
2. **`docs/en/1.0-compatibility-ledger.md` + `docs/ja/1.0-compatibility-ledger.md`** — the topology workspace move row (line 231 in both) now describes the one-mapping-per-recorded-pane contract, the shared-pane travel rule, and the preserved ambiguity refusal, and cites `G697/G735`. The JA row states the same contract as the EN row and `PaneMapContract`. I left the other G697 rows (`guide topology-workspace-move`, `session-layer topology` transport shape) alone: they describe surfaces whose behavior did not change.
3. **`TopologyWorkspaceMoveG697Tests`** — as requested: `AssertMoveGuide` now asserts both shared-pane clauses of `pane_map_contract` (so all four guide renderings that already pass through it are covered), and a new `OperatingGuide_StatesSharedPaneTravelRule_G735` asserts the two clauses in the `HerdrOnlyOperatingGuide` rendered markdown and in `dispatch.topology`, so the guide text cannot silently regress.

Not-defects list respected: `TopologySharedPaneMoveG735Tests.cs` unmodified, refusal wording unchanged, no release notes, `.artifacts/` still uncommitted, no live herdr run.

### Contract clause, in my own words

The pane map is keyed on recorded panes, not roles: one old-pane=new-pane pair per recorded pane, so several roles sharing one recorded pane always travel together to that pane's new pane; two *different* recorded panes mapped onto one new pane is still refused as genuinely ambiguous.

### Behavior the diff changed

Before, `topology move` refused any pane map that landed more than one recorded role on one new pane — even when both roles were recorded on the *same* old pane — so a shared-pane team could never be rebuilt even though `record` pointed it at `move`. The guard now tracks which old pane each new pane came from and refuses only when two distinct old panes converge on one new pane.

### Verification (all run locally against head `364c7f0ea7560c6120a06080d3342e5c01be02d9`)

- `dotnet build src/IntentSystem.Cli/IntentSystem.Cli.csproj -c Release` — 0 warnings, 0 errors.
- Targeted run (`TopologyWorkspaceMoveG697Tests|TopologySharedPaneMoveG735Tests|HerdrOnlyOperatingGuide|GuidePrimacyG701`): `Failed: 0, Passed: 31, Skipped: 0`.
- Fresh full suite (`dotnet test tests/IntentSystem.Cli.Tests/IntentSystem.Cli.Tests.csproj -c Release`) final line: `Passed!  - Failed: 0, Passed:  5212, Skipped: 1, Total:  5213` — the skip is the pre-existing `TaskingHandoffCommandTests.Execute_TolerantToSubAnalyzerFailure_...` skip.
- Note: this evidence is for head `364c7f0e`; the previous evidence (`5211/1/5212`) was for `8f51795b`. The +1 test is `OperatingGuide_StatesSharedPaneTravelRule_G735`. No live herdr check was run.

tomohisa:
ローカル LLM 評価(Qwen3.8 Flash Next Q3_K_XL / llama.cpp / OpenCode)の成果物です。マージはせずクローズします。

この PR は G735(実 PR #1599 で解決済み)を、本番と同じ base commit `4aea6b5e` から同一 packet で再実装させた評価用です。結果: 本物とほぼ同一のデータ構造変更(`HashSet` → old-pane を保持する `Dictionary`)に到達し、「拒否そのものは残して条件を狭める」判断も一致。初回 37 分で PR、自己申告のテスト結果は第三者実行と完全一致。レビュー 1 往復で網羅漏れ(運用ガイド・互換性台帳の未更新)を修正し通過。

評価記録: MyIntentHost の `intents/idd-com/research/local-llm-eval-2026-08/`(04 章)。base ブランチ `eval/base/G735` は差分参照のため残します。

PR #1620, opened 2026-08-27
G743: preserve committed claim result after push uncertainty

### LINKED ISSUE #1617: G743: claim commands report failure after the claim state has already been committed and pushed
# G743: claim commands report failure after the claim state has already been committed and pushed

## Goal

Restore the contract G738 established one release earlier: once the commit and push have succeeded, no teardown outcome may change the command's result. Today the shipped 0.25.0 reports the opposite.

## Why This Slice Exists Now

Field incident 2026-08-27 on `intent-cli 0.25.0-74a1c72-G741`, macOS. Three consecutive `claim acquire` runs each returned:

```
status: error
push_succeeded: false
detail: Could not clean up claim transaction temporary directory '...' before the claim state was committed.
```

All three claims are on `origin/main` anyway, with commits `e32643dfd`, `ef8130372`, `facba5811` at 23:01:17 / 23:01:43 / 23:02:07.

**A false negative here is worse than the Windows hang G738 fixed.** An operator who reads "the state was not committed" concludes no claim exists — and may re-acquire, or hand the scope to another actor, while the registry says they already own it. The whole point of the claim registry is that its answer can be trusted.

## Current Observed State

- `ClaimCommand.CleanupAttemptTimeout` is 250 ms per attempt, and a timed-out attempt breaks without retry.
- Deleting one transaction clone of this host repo measures **0.67 s** (`/usr/bin/time -p rm -rf`), because the clone carries a 1.3 GB `.git`. The bound is therefore always exceeded on this host; the cleanup path always fails.
- `CleanupTransactionRoot` throws only when `committed` is false; `committed` is set right after a zero-exit `git push origin HEAD`.
- Ten leftover clones accumulated to **1.2 GB** under the OS temp root before being removed by hand — one clone per claim operation.

- G738's contract text and its existing tests. This unit restores what they describe; it does not redefine it.
- The push is the transaction boundary. Everything after it is best-effort.
- The OS temp root is the correct location for a leftover clone.

## Target Repo / Path / Part

- Target paths: `src/IntentSystem.Cli/Commands/ClaimCommand.cs`, its tests, `docs/en/09-developer-reference.md`, `docs/ja/09-developer-reference.md`
- Part: the committed-flag propagation into teardown, and the cleanup bound.

## In Scope

- Finding why `committed` is false on a run whose push succeeded, and fixing it. **Do not assume the cause from this body — reproduce it first.**
- A bound that a realistically large clone can meet, or removing timeout-alone as a failure condition.
- Ensuring a genuine pre-commit failure reports its own cause rather than a cleanup message raised from `finally`.

## Out Of Scope

- The claim registry format, the push-retry semantics, and what the transaction commits.
- Reducing the host repo's `.git` size.

Self-contained: this body, `ClaimCommand.cs`, and G738's tests as the contract reference.

## Acceptance Criteria

- [ ] A claim whose commit and push succeeded reports `acquired`/`released` with `push_succeeded: true`, on a host where the cleanup bound is always exceeded.
- [ ] A regression test reproduces the reported shape deterministically and fails against current 0.25.0 behaviour.
- [ ] The bound accommodates a realistically large clone, or timeout alone stops being a failure. State the numbers and the reasoning.
- [ ] A genuine pre-commit failure reports its own cause; a cleanup exception from `finally` never masks it.
- [ ] Leftover clones are bounded in practice, or the command names the leftover path.
- [ ] Both mirrors state the corrected contract, including that a reported failure means the state is **not** durable.

## Verification

- Paste the regression test failing on the pre-fix code and passing after.
- Paste a real `claim acquire` run on a large-`.git` host showing success with a slow cleanup.

## Related Links

- G738 / PR #1609 — the contract this restores. Shipped in v0.25.0.

## Knowledge Maintenance

- Node 03 closeout records that the contract shipped with an inverted report, and what caught it.

## Base Branch Policy

tomohisa:
Closes #1617

## Summary

- Verify the transaction commit and resulting claim state on origin after a nonzero push process result. Only an exact remote commit and ownership-state match is reported as acquired, released, or taken-over with push_succeeded=true.
- Mark the transaction committed before bounded post-commit refresh and teardown, so cleanup or refresh failures cannot invite a duplicate acquire or change the result.
- Preserve the original pre-commit transaction exception when teardown also fails, with a separate leftover-path warning.
- Keep the existing 250 ms per cleanup attempt, three-attempt bound, 50 ms backoff, claim schema, and push retry semantics unchanged.

## Deterministic reproduction

Before the fix, ClaimCommandG743Tests.RemoteCommitAfterPushProcessFailure used a local bare origin and a git push shim that let the remote receive the commit before returning a nonzero process status. The command emitted status=error, push_succeeded=false, attempts=0, and the cleanup detail saying the claim state was not committed, while the remote branch and claim commit were already present. After the fix the same scenario passes with exit code 0, status=acquired, push_succeeded=true, the remote commit SHA, and the bounded cleanup warning naming the leftover OS-temp path.

## Verification

- ClaimCommandG743Tests: Failed: 0, Passed: 2, Skipped: 0, Total: 2.
- Adjacent claim/consumer/worker tests (G679, G680, WorkerClaimCommand): Failed: 0, Passed: 46, Skipped: 0, Total: 46.
- Full Release suite: Failed: 0, Passed: 5263, Skipped: 1, Total: 5264.
- `git diff --check`: clean.

## Workflow boundary

- Canonical child next-action and issue-preflight could not obtain fresh local claim evidence because `.git/FETCH_HEAD` was unavailable in the original child boundary; they reported wait/claim-unavailable for #1617. The supplied verified host authority for execution-unit:G743 held by implementation/team intent-cli-dev was consumed as authoritative, and no host repository was entered or read.
- Canonical issue claim was run with `--github-only --write` and added only `intent-issue-in-progress`.
- No work touched G744/#1618 or G745/#1619.

### DIFF
--- a/docs/en/09-developer-reference.md
+++ b/docs/en/09-developer-reference.md
@@ -2670,7 +2670,7 @@ incomplete packet through publication. The tolerance is scoped to `clarify open`
 
 ---
 
-## Claim transaction teardown (G738)
+## Claim transaction teardown (G738, G743)
 
 `claim acquire`, `claim release`, and `claim takeover` use a temporary clone
 under the OS temp root. The successful plain push of the claim state is the
@@ -2679,7 +2679,17 @@ boundary.
 
 - Before the claim state is committed and pushed, a transaction or teardown
   failure remains a command failure. Cleanup cannot turn an incomplete claim
-  into success.
+  into success. If the transaction itself fails before the boundary, its
+  original cause remains authoritative even when a teardown failure also
+  occurs in `finally`; the cleanup failure is reported separately and never
+  masks that cause.
+- A push process can return nonzero after the remote ref has accepted the
+  transaction. The command fetches `origin` and compares the transaction
+  commit and resulting claim state. Only an exact match is treated as the
+  committed ownership fact and reported as `acquired`, `released`, or
+  `taken-over` with `push_succeeded: true`; otherwise the existing rejected-
+  push/retry path remains in force. This prevents a durable claim from being
+  reported as absent and inviting a duplicate acquire.
 - After a successful push, cleanup is best-effort. A delete attempt is bounded
   to 250 ms, up to three attempts are allowed for ordinary failures, and the
   retry backoff is 50 ms. A timed-out attempt is not retried while its worker
--- a/docs/ja/09-developer-reference.md
+++ b/docs/ja/09-developer-reference.md
@@ -2736,7 +2736,7 @@ strict な projection serializer は**変更していません**。publish-flow 
 
 ---
 
-## claim transaction の teardown (G738)
+## claim transaction の teardown (G738, G743)
 
 `claim acquire`、`claim release`、`claim takeover` は OS の temp root 配下に一時 clone を
 作ります。claim state の plain push 成功が transaction boundary であり、その一時 clone の
@@ -2744,6 +2744,14 @@ cleanup は boundary の後に実行されます。
 
 - claim state が commit され push される前は、transaction または teardown の failure は
   command failure のままです。cleanup が未完了の claim を成功に変えることはありません。
+  transaction 自体が boundary 前に失敗した場合は元の原因を保持し、finally の teardown failure
+  がその原因を隠すことはありません。
+- push process が nonzero を返しても remote ref が transaction を受理している場合があります。
+  command は `origin` を fetch し、transaction commit と結果の claim state を照合します。
+  完全一致だけを ownership fact として `acquired`、`released`、`taken-over` と
+  `push_succeeded: true` で報告します。それ以外は既存の rejected-push/retry path のままで、
+  claim を absent と誤報して duplicate acquire を誘発しません。
+
 - push 成功後の cleanup は best-effort です。delete の 1 回ごとの待機は 250 ms に制限し、
   通常の failure には最大 3 回まで試行し、retry 間には 50 ms の backoff を置きます。
   timeout した試行は、worker が directory に触れている可能性がある間は retry しません。
--- a/src/IntentSystem.Cli/Commands/ClaimCommand.cs
+++ b/src/IntentSystem.Cli/Commands/ClaimCommand.cs
@@ -144,6 +144,7 @@ internal static class ClaimCommand
             var transactionRoot = Path.Combine(
                 Path.GetTempPath(), $"intent-cli-claim-{Guid.NewGuid():N}");
             var committed = false;
+            Exception? transactionFailure = null;
             try
             {
                 var clone = RunGit(Path.GetTempPath(),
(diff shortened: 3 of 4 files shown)


tomohisa (commented):
## Workflow verdict: APPROVED

Reviewed G743 / PR #1620 at exact head `2f372197d0e81a6560fa65c1230536605b83f19d` against `packet.yaml`, implementation notes, review context, issue #1617, and node 03 (`intents/intent-cli/intent-tree/means/03-state-and-audit-strategy.md`).

- I ran the new regression test against the pre-fix parent `74a1c722`: it failed as required with exit code 1, `status=error`, `push_succeeded=false`, and the false “before the claim state was committed” cleanup detail. The fixture's shim first runs the real `/usr/bin/git push`, and only after that succeeds returns exit 77, so this reproduces successful remote receipt plus a nonzero push process result rather than assuming the cause. The same two G743 tests pass at this head (2/2).
- The repair fetches origin and accepts a nonzero push process result only after both the remote branch tip equals the transaction commit and the resulting claim state matches the requested ownership transition. It then sets the committed boundary before best-effort refresh/cleanup. This restores the G738/node-03 contract: a durable claim is never reported as absent, while a genuine unverified/rejected push remains on the existing retry path.
- The separate pre-commit test creates a deterministic path blocker and injects a cleanup failure. `RunTransaction` preserves and rethrows the original transaction exception; the finally path records a separate warning and cannot replace it. The test confirms the thrown cause is neither the old not-committed cleanup text nor the injected cleanup exception, and that the leftover path is named.
- The 250 ms per-attempt bound, 3 attempts, and 50 ms backoff are unchanged (850 ms maximum added wait). That is deliberate: after the committed boundary, cleanup timeout is a named best-effort warning, not a result-changing failure; the tests assert the OS-temp leftover path and unchanged successful exit. This satisfies the packet's timeout-not-a-failure alternative without weakening the pre-commit path.
- Both EN/JA mirrors document the exact verified-remote-state rule, original pre-commit cause preservation, the `250 × 3 + 50 × 2` bound, and the named leftover-path warning. The PR changes only ClaimCommand, its G743 tests, and those two documentation mirrors: no claim schema or push-retry scope change. The final fixture-only commit changes repository-local Git user.name/user.email setup, with no product-code or global Git configuration change.

Focused G743 tests passed 2/2; adjacent claim/worker coverage passed 39/39. These are necessary but not sufficient; I also independently inspected the packet/issue ACs, code, docs, scope diff, and node-03 contract. `git diff --check` is clean, `Closes #1617` is present, no automated review comments required action, and both required CI checks are green at the exact head.

design_alignment_checked=true
Checked sources: G743 packet, implementation notes, review context, issue #1617, node 03, pre-fix and proposed code/tests, EN/JA documentation, PR diff, local test results, and CI.

PR #1621, opened 2026-08-28
G744: archive older supervision history

### LINKED ISSUE #1618: G744: supervision state must keep a small live file and move older records into separate archive files
# G744: supervision state must keep a small live file and move older records into separate archive files

## Goal

Keep the live supervision cycle file small by **moving** older records out into separate archive files, so its size is bounded by a recent window rather than by how long the team has been running.

## Why This Slice Exists Now

Field incident 2026-08-27. `cycles.jsonl` for one team reached **111.5 MB locally, 98.7 MB on origin**, and a host push was rejected by GitHub's 100 MB file limit. Host state is a single tracked tree shared by every domain, so this blocked pushes for **all** domains on the host, not just intent-cli.

`notify supervise shrink` (G734) ran and correctly changed almost nothing — 106,332 records before and after, 1 KB smaller, reporting `archived 0, discarded 0, rotated 0`. That is what it is for: compacting the live file in place while the supervisor runs. It cannot bound the file's size, because every record stays in it.

The emergency measure was to untrack the file by hand. Pushes work again, but that team's supervision state is now local-only, and nothing prevents the same growth in another team's directory.

## Current Observed State

- One team accumulated roughly 96,000 cycle records over about two weeks of continuous supervision.
- `shrink` has no concept of moving records anywhere.
- The affected file is untracked as an emergency measure; other teams' directories remain tracked and unbounded.

- G734's live-safety discipline: no stopping the supervisor, no losing records written during the operation.
- Shrink's in-place compaction semantics are correct and stay unchanged.
- The reader that consumes cycle records today.

## Target Repo / Path / Part

- Target paths: the supervision state writer and reader plus the `notify supervise` surface under `src/IntentSystem.Cli/Commands/`, their tests, `docs/en/12-agent-message-orchestration.md`, `docs/ja/12-agent-message-orchestration.md`
- Part: how supervision cycle history is split between a live file and archives.

## In Scope

- A recent window for the live file, stated and configurable, with a justified default.
- Moving older records into separate archive files, named so a period can be found without opening every file.
- Archived records staying readable by the same reader.
- Live-safety equal to G734's, including the live/archive boundary.
- Stating the shrink-versus-this distinction in the surfaces and both mirrors.

## Out Of Scope

- Changing shrink's in-place semantics.
- Discarding history by default.
- The host's gitignore policy — that is host practice, not CLI behaviour.

Self-contained: this body plus G734's shrink implementation as the live-safety reference.

## Acceptance Criteria

- [ ] The live file keeps only a recent window; older records are **moved** into archive files, not compacted in place and not discarded.
- [ ] The window is stated and configurable with a justified default. The incident file was 111 MB / 96,000 records over ~2 weeks — say why the chosen default is comfortable against that.
- [ ] Archive files are named so a period is findable without opening them all, and archived records remain readable by the same reader as the live file.
- [ ] Nothing is discarded by default; any discard is explicit and opt-in.
- [ ] Live-safe: no supervisor stop; a record written during the move is neither lost nor duplicated across the boundary.
- [ ] The shrink distinction is stated in the surfaces and both mirrors. `shrink` reporting `archived 0, discarded 0, rotated 0` on a 111 MB file was correct behaviour, and an operator must be pointed at the right command.
- [ ] A directory already inside the window round-trips byte-identically (fixture, not constructed objects).
- [ ] Tests: move fires past the window; records readable across the boundary; concurrent write during the move; inside-window directory untouched.

## Verification

- Test output for all four cases; full Release suite final line with the `Failed:` count; head SHA stated.
- Paste a before/after listing of a directory that crossed the window, showing the live file small and the archive present.

## Related Links

- G734 / PR #1598 (shrink — the surface this is often mistaken for).

## Knowledge Maintenance

- Node 03 closeout records the 100 MB incident and the live-window / archive split.

## Base Branch Policy

tomohisa:
Closes #1618. Implements a configurable 7-day live supervision window with UTC YYYY-MM archive files, archive-plus-live reads, lock/journal recovery, and no default discard. Concurrent append is serialized at the archive boundary and remains exactly once; shrink semantics are unchanged. Adds fixture-backed byte-identity, move, archive-reader, and deterministic concurrency tests plus EN/JA orchestration documentation. Verification: focused archive 4 passed; supervision-adjacent 80 passed; G613 terminology guard 6 passed; full Release 5267 passed, 0 failed, 1 skipped, 5268 total; git diff --check passed. Child/host claim distinction was recorded and no host repository was accessed.


tomohisa (commented):
## Workflow verdict: REQUEST-UPDATE

### Implementation finding — default-window justification is incomplete

The G744 packet and issue #1618 acceptance criterion require the configurable seven-day default to be justified against the observed incident numbers: **111.5 MB**, **96,000 records**, and **about two weeks**. The new EN and JA archive sections (and the new `SupervisionGuideText` surface) cite 96,000 records over 14 days and GitHub's 100 MB limit, but omit the incident's observed 111.5 MB file size. As written, operators cannot audit the complete measured basis for the conclusion that retaining approximately half the two-week volume is comfortably below the tracking limit.

Please add the 111.5 MB observation and its relationship to the seven-day/roughly-half-volume rationale in both EN and JA mirrors, and keep the role-facing archive guidance consistent. Add or extend a focused documentation assertion so the three-number justification cannot silently regress.

The other core ACs are evidenced:

- The move test starts with 3 records and verifies 2 moved + 1 retained, period-addressable UTC `2026-07`/`2026-08` archives, a smaller live file, and reader history count 3; therefore history is moved rather than dropped.
- The same reader consumes archives plus live. The deterministic concurrent test holds the archive before replacement, observes an append waiting on the shared lock, then proves the append is absent from the archive, present once in live, and appears once in combined history.
- The inside-window test copies a committed fixture file, snapshots every file byte-for-byte, and proves no archive directory or replacement is created.
- Archive/shrink are distinct in executable surfaces: `notify supervise archive` moves period-named history and reports no default discard, while `shrink` remains in-place with `archived 0, discarded 0, rotated 0`. Lock/journal recovery and archive-before-live publication preserve live safety without supervisor stop.

Focused archive tests passed 4/4; G613 passed 6/6. These passing tests are necessary but not sufficient; the missing 111.5 MB documentation evidence prevents approval. `git diff --check` is clean, `Closes #1618` is present, no automated review comments require action, and CI is green at `8f046eb78057294c5607b21d63864c4eed8b52af`.

design_alignment_checked=true
Checked sources: G744 packet, implementation notes, review context, issue #1618, node 03, archive/store/reader/journal code, fixture and concurrency tests, command surfaces, EN/JA docs, PR diff, local tests, and CI.

tomohisa (commented):
## Workflow verdict: APPROVED

Re-reviewed G744 / PR #1621 at exact head `09d9fbb23b71d6eacf785fc7bfeb0ffeed7198ca` against the prior REQUEST-UPDATE finding, the G744 packet/review context, issue #1618, and node 03.

The sole prior finding is resolved. Both EN and JA archive documentation, plus the role-facing `SupervisionGuideText.ArchiveRule`, now state the complete measured basis: observed **111.5 MB**, approximately **96,000 records** over approximately **14 days**, and the conclusion that a configurable **7-day** default retains roughly half that volume live while staying below GitHub's **100 MB** tracking limit. `NotifySupervisionArchiveRationaleG744RepairTests` normalizes and asserts the complete expected relationship in each mirror and the guide string, so removing any required number or relationship makes the focused assertion fail.

The repair diff is strictly limited to those two mirrors, the guide text, and that focused test. No archive/store/reader code, default window, lock/journal recovery, archive-before-live ordering, concurrent append handling, shrink behavior, or unrelated surface changed. The earlier confirmatory evidence therefore remains intact: history is moved (not discarded), archive plus live use the same reader, concurrent writes land exactly once, the inside-window fixture remains byte-identical, and archive/shrink are distinct command surfaces.

Focused archive+rationale tests passed 5/5; G613 passed 6/6. These passing tests are necessary but not sufficient; I independently confirmed the prior finding's exact scope and that no behavior changed. `git diff --check` is clean, `Closes #1618` remains present, and both required CI checks are green at the exact head.

design_alignment_checked=true
Checked sources: G744 packet, review context, issue #1618, node 03, prior finding, repair diff, EN/JA mirrors, role guidance, focused rationale/archive tests, G613, and CI.

PR #1627, opened 2026-08-28
G747: repair claim transaction defects

### LINKED ISSUE #1625: G747: claim transactions must report the real pre-commit cause and must not depend on the host's current branch
# G747: claim transactions must report the real pre-commit cause and must not depend on the host's current branch

## Goal

Finish what G738 and G743 started: a claim command that says what actually happened. Today a real pre-commit failure is reported as a cleanup failure, the claim silently lands on whatever branch the host has checked out, and `--format json` is not machine-parseable when cleanup warns.

## Why This Slice Exists Now

All three measured on 2026-08-28 against a build of merged main carrying G738 and G743 (`intent-cli 0.25.1-b8f249e-G743`).

**1. The pre-commit cause is still replaced by the cleanup message.** A `claim acquire` for a scope already held identically is a no-op: `git commit` finds nothing to commit. That pre-commit failure is real and correct — but the emitted detail is only:

```
Could not clean up claim transaction temporary directory ... before the claim state was committed
```

An operator cannot tell "already held, nothing to do" from "the transaction broke". G743's acceptance criterion 4 required exactly this and its PR body claimed to satisfy it.

**2. Claim transactions clone and push the host's current branch, not the default branch.** Proven with before/after SHAs: a `claim acquire` advanced `origin/design-thread` from `7f884113` to the claim commit while `origin/main` did not move. **Every claim written from a host whose checkout is not on the default branch is invisible to seats reading `origin/main` until someone forwards the branch by hand.** This is the mechanism behind the four "resurrected claim" incidents recorded last cycle, and behind design and the seats disagreeing about who holds a scope.

**3. The cleanup warning shares the result's stream.** `--format json` emits a prose warning line before the JSON object, so a machine consumer cannot parse it without stripping. Observed on every successful claim on this host, where cleanup always exceeds its 250 ms bound.

## Current Observed State

- The no-op acquire path surfaces only the cleanup message.
- `git push origin HEAD` from a clone of the current branch pushes to that branch's ref.
- The warning writer and the result writer are the same stream.

- G738's contract (the commit is the transaction boundary) and G743's fix (a successful push is reported as success even when teardown fails). Both are correct and their tests must keep passing unmodified.
- The 250 ms cleanup bound and its warn-and-continue behaviour. Out of scope; it is working as designed.

## Target Repo / Path / Part

- Target paths: `src/IntentSystem.Cli/Commands/ClaimCommand.cs`, its tests, `docs/en/09-developer-reference.md`, `docs/ja/09-developer-reference.md`
- Part: failure reporting, target-branch resolution, and the warning stream.

## In Scope

- Reporting a pre-commit failure's own cause, with cleanup reported separately.
- Deciding and documenting the target branch: either always the default branch, or the current branch **with the pushed ref named in the result**.
- Making `--format json` emit parseable JSON on stdout.

## Out Of Scope

- The cleanup bound and its warn-and-continue semantics.
- The claim registry format and push-retry semantics.

Self-contained: this body plus G738's and G743's tests as the contract reference.

## Acceptance Criteria

- [ ] A pre-commit failure reports its own cause. A no-op acquire for an already-held identical scope is reported as such, distinguishable from a broken transaction, with the cleanup outcome reported separately rather than in its place.
- [ ] The transaction either targets the default branch regardless of the host checkout, or names the ref it pushed to in the result. **Choose one and say why** — this decides whether a host on a feature branch can use claims at all.
- [ ] The choice is stated in both language mirrors.
- [ ] `--format json` emits parseable JSON on stdout; warnings go to stderr or into the payload.
- [ ] Tests: a no-op acquire reporting its own cause while cleanup fails; a claim from a checkout on a non-default branch; `--format json` parsing cleanly while a cleanup warning is emitted.
- [ ] G738's and G743's existing tests pass unmodified.

## Verification

- Paste the no-op acquire output showing its own cause.
- Paste before/after refs for a claim made from a non-default-branch checkout.
- Pipe `--format json` through a JSON parser and show it succeeding while a warning is emitted.

## Related Links

- G738 / PR #1609 and G743 / PR #1620 — the contracts this completes.

## Knowledge Maintenance

- Node 03 closeout records the three residual defects and how each was measured.

## Base Branch Policy

tomohisa:
Closes #1625

Implements the three G747 claim transaction repairs:

- preserve the already-held/nothing-to-commit no-op cause when bounded cleanup also fails;
- resolve and explicitly push to the remote default branch, including fail-closed symref validation;
- keep JSON stdout as one parseable document while cleanup warnings remain on stderr.

Focused G747 tests, unchanged G738/G743 claim tests, adjacent worker/claim tests, G613, and the full Release suite pass.

The child next-action/preflight registry reported G747 unheld; the supplied verified host claim authority was consumed without creating or mutating any child product-repository claim.

### DIFF
--- a/docs/en/09-developer-reference.md
+++ b/docs/en/09-developer-reference.md
@@ -2709,6 +2709,24 @@ warning: claim transaction committed successfully, but best-effort cleanup could
 This keeps the committed claim visible to the operator and to downstream
 claim-gated flows without backgrounding the claim or hand-writing its packet.
 
+### Remote default-branch targeting (G747)
+
+Every claim transaction resolves the remote default branch from
+`git ls-remote --symref origin HEAD`. The transaction clone starts from that
+branch and pushes the transaction commit explicitly to
+`refs/heads/<resolved-default-branch>`. The invoking checkout current branch
+is never used as the claim target or advanced by the refresh. If the symref is
+missing, ambiguous, unsafe, or cannot be queried, the command fails closed; it
+never falls back to the current branch. Successful transaction results include
+the resolved `target_ref`.
+
+An acquire whose active record already has the same actor and team is an
+intentional no-op. Its result says that the scope is already held and that no
+claim commit was needed (`nothing to commit`). If teardown also fails, the
+warning is separate, names the leftover path, and cannot replace that primary
+no-op cause. With `--format json`, stdout contains exactly one JSON document;
+cleanup warnings go to stderr so stdout can be piped directly to a JSON parser.
+
 ---
 
 ## Version flow
--- a/docs/ja/09-developer-reference.md
+++ b/docs/ja/09-developer-reference.md
@@ -2769,6 +2769,21 @@ warning: claim transaction committed successfully, but best-effort cleanup could
 これにより、commit 済み claim は operator と downstream の claim-gated flow から見え続け、
 claim の background 実行や packet の手書きは不要です。
 
+### remote の default branch を対象にする場合 (G747)
+
+すべての claim transaction は `git ls-remote --symref origin HEAD` から remote の
+`default branch` を解決します。transaction clone はその branch から始め、transaction
+commit を `refs/heads/<resolved-default-branch>` へ明示的に push します。呼び出し元
+checkout の current branch は claim の target にせず、refresh でも進めません。symref が
+無い、曖昧、安全に扱えない、または query できない場合は fail closed とし、current branch
+へは fallback しません。成功した transaction result には解決した `target_ref` が含まれます。
+
+active record の actor と team が同じ acquire は意図した no-op です。result は scope が
+すでに保持され、claim commit は不要 (`nothing to commit`) であることを示します。teardown
+も失敗した場合は leftover path を含む warning を別に出し、その no-op の主原因を置き換え
+ません。`--format json` の stdout は JSON 文書を 1 つだけ含み、cleanup warning は stderr
+へ出力されるため、stdout を直接 JSON parser に渡せます。
+
 ---
 
 ## バージョンフロー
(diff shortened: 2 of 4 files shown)


tomohisa (commented):
APPROVED (approve-equivalent COMMENTED verdict)

design_alignment_checked=true. I reviewed G747 packet.yaml, implementation.md, review-context.md, github-body.md, and the packet-referenced `intents/intent-cli/intent-tree/means/03-state-and-audit-strategy.md`. The default-branch decision is correct for that durable/auditable claim registry: resolve `origin` HEAD via `git ls-remote --symref`, clone/fetch that branch, explicitly push to `refs/heads/<resolved-default-branch>`, surface `target_ref`, and fail closed rather than silently writing the invoking checkout branch. Both EN/JA mirrors state the same policy and rationale.

All #1625 acceptance criteria are evidenced. I ran an isolated command-entry-point reproduction: an already-held same actor/team acquire with forced cleanup failure exited 1 and wrote one JSON document whose primary `detail` says “already held” and “nothing to commit”; its cleanup warning, including the leftover path, was on stderr. Piping stdout to `jq` succeeded. The new non-default-checkout regression verifies refs, not merely exit status: main advances to the claim commit while `design-thread` remains unchanged and the invoking checkout stays on that branch. Symref parsing rejects missing, conflicting, and unsafe values.

The focused suite passed: G747 plus unchanged G738 and G743 claim-contract tests — 8 passed, 0 failed, 0 skipped. `ClaimCommandG679Tests.cs` (G738) and `ClaimCommandG743Tests.cs` are unmodified by this PR. Existing commit-boundary behavior remains intact; the 250 ms cleanup bound, claim registry format, and push-retry policy were not expanded. `git diff --check` is clean, CI is green, `Closes #1625` is present, and the head remained `441abdf8f695be0ada192e92b887fb94aa014650`.

PR #1634, opened 2026-08-28
G750: keep supervision cycle history runtime-local

### LINKED ISSUE #1632: G750: supervision cycle history is runtime-local state and must be ignored by the CLI, not tracked in the repository
# G750: supervision cycle history is runtime-local state and must be ignored by the CLI, not tracked in the repository

## Goal

Make the CLI treat supervision cycle history as what it is — runtime-local decision state — by owning its ignore, the way it already owns the ignore for machine-local topology. An operator should never discover this by having a push rejected.

## Why This Slice Exists Now

Field incident 2026-08-28. `cycles.jsonl` for one team reached **111.5 MB**, GitHub's 100 MB file limit rejected the push, and because host state is a single tracked tree shared by every domain, **that rejection blocked pushes for every domain on the host** — not just intent-cli. The emergency measure was an operator-side gitignore plus `git rm --cached`. The file has since grown past **455 MB** locally.

**Every reader of this file is same-host:**

- `NotifySupervisionStore.ReadCycles` supplies `LastCycle` / `LastIntervalCycle` for continuity between consecutive supervisor runs.
- The install path verifies first-cycle proof (`first-cycle-proof-failed`) on the machine that installed the job.
- `SupervisionGuideText` instructs the operator to verify the live PID and the first cycle record — locally.

No cross-host reader exists. And the records are structurally machine-local: each carries a writer identity containing the **OS process id**, plus that host's observed seat state-change sequences.

## Current Observed State

- The repository already has the precedent and the mechanism: `.intent-cli/topology/.gitignore` contains a single `*`, and `git check-ignore` confirms the machine-local topology record is ignored. That ignore is CLI-owned.
- Supervision has no equivalent. `cycles.jsonl` is tracked until an operator intervenes by hand.
- **Not everything in that directory is runtime-local.** `stalls.jsonl` (16 MB, findings history), `bound.json`, `emission-policy.json`, `evidence-definitions.json`, `pre-approval-policy.json` and `shrink-audit.jsonl` are tracked today and must stay tracked.

- The topology ignore mechanism, which this should follow rather than duplicate.
- G744's archive behaviour. Archiving bounds the live file; it does not decide whether the file belongs in git. Both are needed and this is the second one.

## Target Repo / Path / Part

- Target paths: the supervision store and init scaffold under `src/IntentSystem.Cli/Commands/`, their tests, `docs/en/12-agent-message-orchestration.md`, `docs/ja/12-agent-message-orchestration.md`
- Part: the CLI-owned ignore for supervision cycle history and the guidance describing shared versus runtime-local supervision state.

## In Scope

- A CLI-maintained ignore covering cycle history, following the topology mechanism.
- A stated canonical path for a host that already tracks the file to stop tracking it.
- Documenting which supervision state is shared and which is runtime-local, with the reason.

## Out Of Scope

- The write rate that made the file large. That is a separate defect with its own unit.
- Changing what supervision reads or writes.
- Ignoring anything else in the supervision directory.

Self-contained: this body plus the existing topology ignore implementation as the mechanism reference.

## Acceptance Criteria

- [ ] The CLI writes and maintains the ignore itself, so a fresh host never hits this and no operator hand-writes a gitignore entry. Use the topology mechanism; do not invent a second one.
- [ ] The ignore is scoped to cycle history only. `stalls.jsonl` and the policy/manifest files stay tracked, proven by a test that lists what is and is not ignored in a scaffolded supervision directory.
- [ ] A host that already tracks the file gets a stated canonical path to stop — a documented command or a named repair — rather than being left to run `git rm --cached` by hand. **Asking operators to hand-edit repository state is what this unit exists to stop.**
- [ ] Both mirrors state which supervision state is shared and which is runtime-local, and why cycle history is the latter: same-host readers, records carrying a process id.
- [ ] Behaviour is otherwise unchanged — supervision reads and writes cycle history exactly as before; existing supervision tests pass unmodified.

## Verification

- Test output listing ignored versus tracked paths in a scaffolded supervision directory.
- The existing supervision suite passing unmodified.

## Related Links

- G744 / PR #1621 (archive — bounds the file's size; this decides whether it belongs in git).
- The topology ignore, as the mechanism to follow.

## Knowledge Maintenance

- Node 03 closeout records the shared-versus-runtime-local distinction and the incident that forced it.

## Base Branch Policy

tomohisa:
Closes #1632

- Add CLI-owned .intent-cli/supervision/.gitignore rules for cycles.jsonl and cycles-archive only.
- Add canonical notify supervise repair-cycle-history migration for already-tracked cycle files without deleting them.
- Keep stalls and supervision policy/manifest files trackable.
- Preserve supervision read/write and archive behavior; document the same-host/process-id ownership boundary in EN/JA.
- Add fresh-scaffold, writer-maintenance, migration, dry-run, inventory, and docs regressions.


tomohisa (commented):
APPROVED — reviewed at head `527d1f82d368bbca45aa0be59012ddeb5a13bd6f`.

This is an approve-equivalent COMMENTED review (same-account review submission). `Closes #1632` is correct.

Acceptance verification:

- The CLI owns one idempotent directory-local mechanism, `NotifySupervisionStore.EnsureCycleHistoryIgnore`, used by fresh `intent init`, normal cycle writes, archive/shrink writes, and the repair command; it is not a second ad-hoc ignore scheme. Its exact two rules are `**/cycles.jsonl` and `**/cycles-archive/`.
- I built this head and independently exercised a fresh git host. Git's ignore matcher ignored only the cycle file and archive path; it did not ignore `stalls.jsonl`, `bound.json`, `emission-policy.json`, `evidence-definitions.json`, `pre-approval-policy.json`, or `shrink-audit.jsonl`, all of which were trackable. The G750 fixture makes that inventory an assertion rather than merely asserting the ignore exists.
- I also exercised an already-tracking host: `notify supervise repair-cycle-history --dry-run --format json` reported the two tracked history paths and left the index/files unchanged; `--write` untracked precisely those two paths while both remained on disk, retaining all six shared files in the index. The command's output names preserved paths and `preserved_files=true`; no operator-side `git rm --cached` is prescribed. The regression also covers removal of obsolete root ignore rules without losing its unrelated rule.
- Existing G741 and G744 supervision tests are unmodified, pass, and the only added supervision test file is the G750 regression file. The EN and JA mirrors both state the same ownership basis: all readers are same-host consumers and cycle records carry the OS process id; shared state remains reviewable/trackable. They explicitly preserve the read/write, archive, and emission-rate boundary and name G751 as out of scope.

Validation:

- `dotnet build src/IntentSystem.Cli/IntentSystem.Cli.csproj -c Release --no-restore` — passed.
- Focused G750/G741/G744 supervision tests — 17 passed, 0 failed.
- Full `IntentSystem.Cli.Tests` suite — 5,310 passed, 0 failed, 1 skipped.
- `git diff --check 93f07f892f6514bc561493339b11e36de0e36555..527d1f82d368bbca45aa0be59012ddeb5a13bd6f` — clean; scope is the intended CLI/docs/test surfaces only.
- CI is green: source contract and npm dry-run. I reconfirmed the PR head remains `527d1f82d368bbca45aa0be59012ddeb5a13bd6f` immediately before this review.

Design alignment checked against `.intent-cli/issues/G750/packet.yaml`, `implementation.md`, `review-context.md`, `github-body.md`, and `intents/intent-cli/intent-tree/means/03-state-and-audit-strategy.md`: this creates a bounded, CLI-mediated recovery surface for the observed tracked runtime history while retaining durable shared audit/policy state, consistent with the state/audit strategy.

PR #1635, opened 2026-08-29
G751: stop persisting event-mode no-observation waits

### LINKED ISSUE #1633: G751: event-mode supervision writes a durable cycle record per wait return — about 3.6 per second, 300 MB per day
# G751: event-mode supervision writes a durable cycle record per wait return — about 3.6 per second, 300 MB per day

## Goal

Make event-mode supervision's write cadence something an operator can predict from the documentation, instead of something they discover from a 455 MB file.

## Why This Slice Exists Now

Measured 2026-08-28 on host `tomohisa/MyIntentHost`, team `intent-cli-dev`, against `cycles.jsonl` at **455 MB / 432,036 records**.

**Trigger breakdown:**

```
event-wait  430,924   (99.7%)
interval        801
event            60
absent          251
```

**Cycle durations, last 3,000 records:** minimum `0.000 s`, median `0.000 s`, maximum `0.000 s` — while those same records declare `interval_seconds: 300` and `bound_seconds: 900`. The wait returns instantly and a full ~1 KB record is persisted each time.

**Rate:** 3.6 records/second ≈ **310 MB per day for one team**. Records per day: 289,651 on 08-28, 139,339 on 08-27, against 100–1,000 per day before that. **99.6% of the file was produced in two days.**

G744 gave the live file an archive boundary and G750 removes it from git. **Neither reduces what is written** — at this rate the archives grow exactly as fast as the live file used to.

## Current Observed State

- Event-mode wait returns without elapsed time and each return persists a cycle record.
- Two writer processes were alive simultaneously (PID 59770 since 08-27, PID 43854 since 08-25) and six `notify supervise` processes were running on the machine.
- The stalls log holds 112 `duplicate-supervisor` findings, so duplicate writers are a known and separately handled condition. **This unit is about the per-return write, which a single writer alone would still produce.**

- Event mode's documented purpose: keep the blocking herdr waits inside the supervisor process while interval cycles remain the safety floor.
- G744's archive behaviour and G750's ignore decision. Both are correct and neither addresses this.
- The duplicate-supervisor finding and its G699 backoff/park handling.

## Target Repo / Path / Part

- Target paths: the event-mode wait loop in the measured supervisor under `src/IntentSystem.Cli/Commands/`, its tests, `docs/en/12-agent-message-orchestration.md`, `docs/ja/12-agent-message-orchestration.md`
- Part: what an event-mode iteration persists.

## In Scope

- Not persisting a durable cycle record for a wait that returned having observed nothing.
- Preserving the documented cadence guarantee: interval cycles as the safety floor, genuine observations still recorded.
- Stating the resulting cadence in both mirrors.

## Out Of Scope

- Duplicate supervisors and their existing finding.
- G744's archive and G750's ignore.
- Changing findings, wake routing, parking, emission policy, or the observation-only discipline.

Self-contained: this body plus the event-mode implementation and its existing tests.

## Acceptance Criteria

- [ ] A supervisor in event mode does not persist a durable cycle record for a wait that returned without observing anything. **State what now counts as a record-worthy cycle, and why.**
- [ ] The documented cadence guarantee is preserved: interval cycles remain the safety floor and a genuine observation still produces its record. Tests must fail if either is lost.
- [ ] **The rate is measured, not assumed** — the PR states records-per-hour for an event-mode supervisor before and after, on the same declared interval and bound.
- [ ] A wait that returns instantly and repeatedly is itself surfaced rather than silently absorbed — either the supervisor reports that condition, or the PR states why silence is correct. Do not simply stop writing and leave the underlying behaviour invisible.
- [ ] Findings, wake routing, parking, emission policy and the observation-only discipline pass their existing tests unmodified.
- [ ] Both mirrors describe what event mode persists and at what cadence, so growth is predictable from the docs.

## Verification

- Before/after records-per-hour for a running event-mode supervisor.
- Tests for the interval floor and for a genuine observation still being recorded.

## Related Links

- G744 / PR #1621 (archive), G750 (ignore) — both necessary, neither sufficient.
- G699 duplicate-supervisor backoff/park, unchanged by this unit.

## Knowledge Maintenance

- Node 08 closeout records the per-return write and the measurements that exposed it.

## Base Branch Policy

tomohisa:
Closes #1633

G751 makes successful event-mode wait returns without a qualifying observation operator-visible through the forced event-wait-no-observation warning without appending a durable event-wait cycle. Genuine transitions, wait failures, and the interval safety-floor cycles remain unchanged.

Verification includes deterministic no-observation, repeated instant-return, genuine-observation, interval-floor, and running-cadence tests plus G613 and the full Release suite. No release-prep or execution-unit claim state was touched.

### DIFF
--- a/docs/en/12-agent-message-orchestration.md
+++ b/docs/en/12-agent-message-orchestration.md
@@ -455,8 +455,9 @@ operation.
 
 This is ownership repair, not compaction: G734 `notify supervise shrink` keeps
 all records while compacting existing state in place, whereas G750 only changes
-which cycle-history paths are CLI-owned runtime-local files. G751 remains the
-separate later unit for write-rate design and is not part of this change.
+which cycle-history paths are CLI-owned runtime-local files. G751 is the
+separate event-mode persistence unit documented below; it does not change this
+ownership boundary.
 
 > **Preview through 1.x (G750).** Directory-local cycle-history ownership and
 > its named repair command are post-freeze preview behavior outside the 1.0
@@ -1005,6 +1006,41 @@ scheduler artifacts embed their invocation, so an installed interval-only
 artifact stays interval-only. Adoption requires running `notify supervise
 install ... --event-mode` again, inspecting the new artifact, and explicitly
 unregistering/re-registering it with the printed operator commands.
+
+### Event-mode record worthiness and measured cadence (G751 — preview-through-1.x)
+
+G751 separates a successful wait return from a record-worthy observation. A
+wait result is record-worthy when it observes a qualifying `working` to
+settled transition. A wait death, error, identity mismatch, or parse failure
+also remains record-worthy evidence and is still written as an `event-wait`
+cycle with `rearm_attempted: true`. Every periodic interval pass remains a
+recorded safety-floor cycle, including the first-cycle and continuity proof.
+
+A successful wait return with no qualifying observation—including repeated
+immediate returns while the recipient is already settled—is classified as
+`wait-returned-without-observation`. It is re-armed after the existing
+one-second delay and emits the forced `event-wait-no-observation` warning with
+the role, workspace, pane, returned status, and reason. It does not append a
+durable `event-wait` cycle. Repeated instant returns therefore remain visible
+to the operator without turning a no-observation wait into a durable-record
+hot loop; they cannot suppress the interval safety floor, genuine event
+observations, liveness/bound evidence, or the first-cycle proof. Findings,
+wake routing, parking, emission policy, and observation-only discipline are
+unchanged.
+
+The child fixture measurement used a running event-mode supervisor with
+`interval=300` seconds and `bound=900` seconds, a deterministic idle wait
+return, and a 25 ms fixture re-arm delay. Before G751, a 2.010-second sample
+produced 75 raw records (`event-wait:74`, `interval:1`), or 134341.50
+records-per-hour. After G751, a 2.009-second sample produced 1 raw record
+(`interval:1`), or 1792.32 records-per-hour. These are independently measured
+running-fixture counts, not rates inferred from source or unit tests; the
+actual build commit is reported with the PR evidence. The 300-second interval
+and 900-second bound remain the declared cadence and safety contract.
+
+G751 changes event-mode persistence only. G699 duplicate-supervisor
+backoff/park, G744 archive, G750 ownership, findings, wake routing, parking,
+emission policy, and observation-only behavior remain outside this unit.
 
 > **Preview through 1.x (G659).** Event waits, transition/wait records, and
 > event/interval de-duplication are post-freeze preview behavior. They make no
--- a/docs/en/12-agent-message-orchestration.md
+++ b/docs/en/12-agent-message-orchestration.md
@@ -1028,15 +1028,29 @@ observations, liveness/bound evidence, or the first-cycle proof. Findings,
 wake routing, parking, emission policy, and observation-only discipline are
 unchanged.
 
-The child fixture measurement used a running event-mode supervisor with
-`interval=300` seconds and `bound=900` seconds, a deterministic idle wait
-return, and a 25 ms fixture re-arm delay. Before G751, a 2.010-second sample
-produced 75 raw records (`event-wait:74`, `interval:1`), or 134341.50
-records-per-hour. After G751, a 2.009-second sample produced 1 raw record
-(`interval:1`), or 1792.32 records-per-hour. These are independently measured
-running-fixture counts, not rates inferred from source or unit tests; the
-actual build commit is reported with the PR evidence. The 300-second interval
-and 900-second bound remain the declared cadence and safety contract.
+The before/after cadence comparison keeps startup arithmetic separate from a
+warmed steady-state window. The independent review reproduction on pre-fix
+`b525191a` ran the event-mode supervisor with the same `interval=300` seconds,
+`bound=900` seconds, deterministic idle wait, and 25 ms fixture re-arm. It
+observed 73 raw records in a 2.065-second wall window (`event-wait:72`,
+`interval:1`, including one startup interval), reported 127269.78
+records-per-hour, and is explicitly a startup-plus-two-second observation—not
+a steady-state rate. That measured result is retained as the honest pre-fix
+comparison; it is not used to predict durable-file growth.
+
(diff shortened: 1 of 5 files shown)


tomohisa (commented):
REQUEST-UPDATE — reviewed at head `4774cde66944cf84264865fe8592081a3d6fe3cf`.

This is a same-account COMMENTED verdict. `Closes #1633` is correct.

## Blocking implementation finding — the required post-change rate is not measured at steady state

Packet acceptance criterion 3 and the review context require before/after **records-per-hour for a running event-mode supervisor with the same declared interval and bound**, not an extrapolation from startup. The new running-cadence test starts `RunLoop`, which immediately writes its first interval pass before beginning the 300-second sleep, then cancels after roughly two seconds. No 300-second interval boundary is observed after that initial pass. Its one-record result is therefore the inevitable startup record, not a measurement of steady-state cadence.

I reproduced this on the pre-fix base and this head with the supplied running-supervisor fixture (`interval=300`, `bound=900`, idle wait returns, 25 ms fixture re-arm):

- Pre-fix `b525191a`: 73 records in 2.065 s (`event-wait:72`, `interval:1`), 127269.78/hour; the added test correctly fails there because event-wait count is nonzero.
- Head `4774cde6`: 1 record in 2.078 s (`interval:1`), 1732.31/hour. That is the initial pass divided by two seconds, not the expected no-observation steady state of one interval record per declared 300 seconds (12/hour after startup).

The EN/JA text repeats the same invalid 2.009-second / 1-record arithmetic as 1792.32/hour. Consequently it does not let an operator predict durable-file growth, and it does not meet the packet's “rate is measured, not assumed” contract. Please replace it with a defensible before/after method that distinguishes the initial first-cycle proof from a warmed-up steady-state window, actually exercises the declared 300-second floor (or an equivalently controlled elapsed-time clock), and pin the resulting rates/method in the test and both docs/PR evidence.

Everything else I checked is consistent with the packet, but tests are necessary rather than sufficient:

- Record-worthiness is correctly narrow in the diff: a successful no-observation return is not durable; qualifying transition and failure paths retain event-wait cycle construction; interval passes remain durable. The G751 safety-floor and genuine-observation tests pass.
- The no-observation condition is visible through the forced supervisor output writer (CLI stdout JSON/markdown), never written to cycle/stall/policy state, and leaves findings, wake routing, parking, emission policy, and observation-only behavior untouched. It is intentionally not deduplicated, but the existing one-second re-arm bounds actual emission to one warning/second rather than the 25 ms fixture cadence; this is a non-blocking observation, not a persistence change.
- The only added supervision test file is G751; G659/G675/G699/G741/G744/G750 adjacent tests are unmodified and pass. The EN/JA mirrors carry the same persistence-category wording and G613 passes. G699 duplicate backoff/park, G744 archive, and G750 ignore are outside the diff.

Validation: focused G751 + unchanged adjacent supervision + G613: 57 passed, 0 failed; full `dotnet test IntentSystem.sln -c Release --no-restore`: passed (CLI 5,315 passed, 1 skipped); `git diff --check b525191a..4774cde6`: clean; CI source-contract and npm dry-run are green.

Design alignment checked against `.intent-cli/issues/G751/packet.yaml`, `implementation.md`, `review-context.md`, `github-body.md`, and intent nodes `means/08-agent-message-orchestration.md` and `means/03-state-and-audit-strategy.md`: suppressing empty durable cycles while retaining the interval/liveness proof matches the intended boundary, but the published cadence measurement must be repaired before approval.

tomohisa (commented):
APPROVED — re-reviewed at head `3fa929a58ac54fdfa906d352fdd827f08105b28f`.

This is an approve-equivalent same-account COMMENTED review. `Closes #1633` remains correct. The sole prior REQUEST-UPDATE finding is fixed.

The repaired `RunningEventModeSampleSeparatesStartupFromWarmCadenceAtTheDeclaredFloor_G751` runs the real `NotifyMeasuredSupervisor.RunLoop` event-mode path. It leaves the mandatory first `RunOnce` interval cycle separate, then its controlled production delay receives and asserts twelve exact 300-second delays, advances the same supervisor clock at each boundary, and the supervisor persists/counts the resulting cycles. My independent run produced:

`startup_first_cycle_records=1 warm_window_seconds=3600 warm_interval_records=12 event-wait_records=0 trigger_mix=interval:13 raw_records=13 records-per-hour=12.00 interval_delay_calls=12`.

That is a controlled elapsed-time observation of twelve real post-startup interval-floor boundaries, not a fabricated count or a two-second startup extrapolation. The prior pre-fix result is now explicitly and correctly retained only as a short, non-steady-state wall sample (`73` records, `72` event-wait, `1` startup interval, `2.065` seconds); it is no longer presented as method-equivalent to, or predictive of, the warmed steady-state rate. Both EN and JA mirrors pin the identical method, counts, and 12.00/hour result.

Regression boundary review:

- The repair diff from `4774cde6` contains only `docs/en/12-agent-message-orchestration.md`, `docs/ja/12-agent-message-orchestration.md`, and `NotifyEventSupervisionG751Tests.cs`; no product source moved.
- Existing interval-floor, genuine-observation, no-observation warning, record contents, duplicate-supervisor/G699, wake/parking/emission/observation-only, G744 archive, G750 ignore, and shrink behavior remain unchanged; adjacent existing tests were not edited.
- The forced `event-wait-no-observation` warning and one-second re-arm remain unchanged. Empty successful waits still have no durable event-wait cycle; genuine observations, failures, and interval proof remain durable.

Validation (tests are necessary, not sufficient; packet/intent evidence above is the sufficiency check):

- Repaired controlled-clock cadence test: passed with the exact evidence above.
- Focused G751 plus unchanged G659/G675/G699/G741/G744/G750 supervision and G613 checks: 57 passed, 0 failed.
- Isolated full `dotnet test IntentSystem.sln -c Release --no-restore`: passed; CLI 5,315 passed, 1 skipped. This uses a writable isolated NuGet/home cache; earlier unisolated cache-permission failures are environmental and were not treated as product failures.
- `git diff --check b525191a..3fa929a5`: clean; both CI checks are green; head reconfirmed immediately before submission.

Design alignment checked against `.intent-cli/issues/G751/packet.yaml`, `implementation.md`, `review-context.md`, `github-body.md`, and intent nodes `means/08-agent-message-orchestration.md` and `means/03-state-and-audit-strategy.md`: record-worthy persistence now excludes empty returns while preserving interval/liveness evidence, and the operator-facing cadence is finally measured and predictable without expanding the packet's out-of-scope boundaries.

PR #1643, opened 2026-08-29
G755: verify claims from canonical default branch

### LINKED ISSUE #1642: G755: the claim reader resolves a ref the writer never writes, so mutual exclusion silently disappears on a non-default checkout
# G755: the claim reader resolves a ref the writer never writes, so mutual exclusion silently disappears on a non-default checkout

## Goal

Make the execution-unit claim **reader** resolve the same canonical Git ref the claim **writer** already targets, so that a successful `claim acquire` is observable by every claims-gated command no matter which branch the reading checkout sits on. Resolve the canonical branch in exactly one place that both sides call, and make an unresolvable canonical ref fail closed instead of reporting that no claims store exists.

## Why This Slice Exists Now

G747 fixed the writer half of this asymmetry and its closeout recorded the reader half as an explicitly deferred separate-unit candidate (`intents/intent-cli/intent-tree/means/03-state-and-audit-strategy.md`, G747 closeout entry): push now goes to the default branch, but `claim verify` still reads the current branch's remote ref.

That residual is not an ergonomic wart. Measured today it produces a **fail-open**: a reader standing on a branch that carries no claims tree concludes that no claims store is configured at all and returns `passed: true`, so every claims-gated command skips mutual exclusion. Two seats on two branches can each believe they hold the same execution unit. The claim registry exists so that the answer to "who holds this scope" can be trusted; a reader that answers "there is no registry" because of where it is standing removes that guarantee silently.

The documented contract already says this must not happen. `docs/en/05-implementation-loop.md` states that "local absence or a stale local record is never proof of ownership **or no-store**". The shipped reader violates its own documented invariant, because absence on the reader's branch is treated as no-store.

The second-order cost is operational and was paid repeatedly: the previous design seat ran `git push -f origin origin/main:design-thread` roughly a dozen times in one cycle, purely to make the writer's ref visible to the reader.

## Current Observed State

Measured with installed `intent-cli 0.27.0-f43fbd1-G753`.

**A. Clean-room reproduction.** A purpose-built host (bare `origin`, `main` and `design-thread` created at the *identical* sha, so "the branch is simply behind" is excluded by construction). `claim acquire` was run from the `design-thread` checkout and self-reported `target_ref: refs/heads/main`; `origin/main` advanced and `origin/design-thread` did not move. Reading back the scope that was just acquired, from that same checkout, at that same moment:

| Reader checkout state | `claim verify` result | Consequence |
| --- | --- | --- |
| Branch carries **no** claims tree | `status: not-configured`, **`passed: true`** | **Fail-open.** Mutual exclusion silently disabled for all 7 gated commands |
| Branch carries a **stale** claims tree | `status: unheld`, `passed: false` | Gate refuses work that is legitimately held |
| **Detached HEAD** | `status: canonical-unavailable`, `passed: false` | Every gate hard-fails for want of a branch name |
| Checkout on the default branch | `status: owned`, `passed: true` | Correct |

The branch-alignment workaround was also measured to be *per-acquire*: after syncing `design-thread` to `main`, the already-acquired scope read back `owned`, and a scope acquired immediately afterwards read back `unheld` again.

**B. On this host, same repo, same scope, same instant.** `execution-unit:G755` was acquired for this very packet (`target_ref: refs/heads/main`, `push_succeeded: true`, commit `e43134b3`). Reading it back:

- from the `main` worktree → `status: owned`, `passed: true`
- from the `design-thread` root → `status: unheld`, `passed: false`

- The **writer** is correct and settled. G747 deliberately resolves the remote default branch via `git ls-remote --symref origin HEAD` and explicitly refuses to fall back to the current branch (`ClaimCommand.cs:119`, `:127-131`, `:141`). Do not change what the writer targets.
- G679 push-CAS remains the sole writer of claim state: pull `--ff-only` → immutable record → commit → plain push, and only push success is the acquisition fact.
- Claims are one immutable file per scope under `.intent-cli/claims/`, with history under `.intent-cli/claims/history/`.
- The four-thread collaboration model, the canonical `intent-cli` route, the child-isolation rule, and the no-AI-provider-launch rule are all unchanged and not up for relitigation.
- The `automation stalled-work` checkout-freshness false positive is a **different defect with a different fix** and is deliberately excluded here; its measured evidence is recorded in this packet's `implementation.md` for the follow-up unit.

## Target Repo / Path / Part

- Target paths: `src/IntentSystem.Cli/Commands/ClaimOwnershipVerifier.cs`, `src/IntentSystem.Cli/Commands/ClaimCommand.cs`, `tests/IntentSystem.Cli.Tests/ClaimCommandG747Tests.cs`, `docs/en/05-implementation-loop.md`, `docs/ja/05-implementation-loop.md`

Target part: the canonical-evidence read path in `ClaimOwnershipVerifier.ReadCanonicalEvidence`, the default-branch resolution it must share with `ClaimCommand.RunTransaction`, and the two documentation mirrors that currently describe the current-branch read.

## In Scope

- Extract the remote-default-branch resolution now living in `ClaimCommand.RunTransaction` into a single shared resolver, and call it from both the writer and `ClaimOwnershipVerifier.ReadCanonicalEvidence`, so the two cannot drift apart again.
- `ReadCanonicalEvidence` fetches and reads the claims tree from the **canonical** branch ref (`refs/heads/<default>` on `origin`) instead of `refs/remotes/origin/<current-branch>` (`ClaimOwnershipVerifier.cs:187`, `:200`).
- Fail closed when the canonical branch cannot be resolved or fetched: return `canonical-unavailable`, never `not-configured`.
- `not-configured` must mean "the canonical branch carries no claims store" and nothing else.
- A detached-HEAD checkout returns the canonical answer rather than refusing for want of a named branch (`ClaimOwnershipVerifier.cs:190`).
- Update both documentation mirrors so the described ref matches the shipped behavior.
- Regression tests for each state above.

## Out Of Scope

- The writer's resolution or push target — settled by G747; its existing tests must stay green **unchanged**.
- `CheckoutFreshnessProbe` / `automation stalled-work` checkout freshness. Separate defect, separate unit.
- Any configurable or overridable claims ref. The registry is cross-team and cross-domain; its contents must not depend on reader configuration any more than on reader checkout. No observed host needs this, and adding it would require changing the writer too.
- The claim record schema, push-retry semantics, take-over/release attribution, and the known bounded-cleanup residual (`temporary directory deletion exceeded 250 ms`).
- `ReadLocalEvidence`, the non-Git-root path used by fixtures and embedded callers, which stays as it is.
- Lifecycle labels, queue-state, and any host-state transition.

The child PR changes the intent-cli claim ownership reader so that it reads claim records from the repository's canonical branch — the same `refs/heads/<default-branch>` that `claim acquire` resolves and pushes to — instead of from the remote ref of whatever branch the reading checkout happens to be on. The default-branch resolution must exist in one place called by both the writer and the reader. When that canonical ref cannot be resolved or fetched, the reader must fail closed with `canonical-unavailable`; it must never report `not-configured` (which passes every gate) merely because the reader's own branch has no claims directory. A detached-HEAD checkout must get the canonical answer rather than a refusal. The PR must add tests that fail on the parent commit and pass on the change, must leave G747's writer tests untouched and green, and must correct the two documentation mirrors that currently describe the current-branch read.

## Acceptance Criteria

1. `ClaimOwnershipVerifier.ReadCanonicalEvidence` contains no `git branch --show-current` call, and does not derive the canonical ref from the checkout's branch. Verifiable by reading the file and by `grep -n "show-current" src/IntentSystem.Cli/Commands/ClaimOwnershipVerifier.cs` returning nothing.
2. The remote-default-branch resolution exists in exactly one function, and both `ClaimCommand.RunTransaction` and `ClaimOwnershipVerifier.ReadCanonicalEvidence` call it. Two independent resolutions in the two files is a failed criterion even if both happen to return `main`.
3. **Regression test — the reported defect.** Using the existing `ClaimRepositories.PrepareNonDefaultCheckout()` harness in `ClaimCommandG747Tests.cs` (feature branch `design-thread`): acquire a scope from the non-default checkout, then `ClaimOwnershipVerifier.Verify` from that same checkout returns `status: owned`, `passed: true`, naming the acquiring actor and team.
4. **Regression test — the fail-open.** A reader whose branch carries no claims tree while the canonical branch carries a populated store does **not** return `not-configured`/`passed: true`. It returns the canonical store's answer for the scope.
5. **Regression test — fail closed.** When the canonical branch cannot be resolved or fetched, the result is `status: canonical-unavailable`, `passed: false`. Not `not-configured`, not `unheld`.
6. **Regression test — `not-configured` keeps its meaning.** A host whose canonical branch genuinely carries no claims store still returns `not-configured`, `passed: true`, preserving legacy single-team behavior byte-for-byte.
7. **Regression test — detached HEAD.** A detached checkout at the canonical commit returns the canonical answer for the scope, not `canonical-unavailable` for want of a named branch.
8. Every test in criteria 3-5 and 7 is demonstrated **failing on the parent commit** and passing on the change, with the actual pre-change output pasted in the PR. A test that passes before the change does not pin this defect.
9. Assertions pin observable facts — the `status` and `passed` values, the holder and holder-team — not message sentences. A test asserting a detail string stays green when the sentence becomes false.
10. All pre-existing tests in `ClaimCommandG747Tests.cs` remain **unmodified** and green, including `NonDefaultCheckout_PushesOnlyResolvedRemoteDefaultRef_G747`.
11. `docs/en/05-implementation-loop.md` and `docs/ja/05-implementation-loop.md` describe reading the canonical branch. The current sentences ("the verifier first fetches the current branch", "verifier が current branch を最初に fetch し") are false after this change and must not survive it.
12. The full Release-configuration test suite passes, with the actual pass/fail counts pasted in the PR rather than summarized.

## Verification

- New tests in `tests/IntentSystem.Cli.Tests/ClaimCommandG747Tests.cs` (or a sibling `ClaimOwnershipVerifierG755Tests.cs`) covering acceptance criteria 3, 4, 5, 6 and 7. Each uses a real bare-origin + clone fixture, as the G747 tests already do, rather than a mock, because the defect is in the Git ref that is read.
- Pre-change demonstration: run the new tests on the parent commit and paste the failures. Criterion 8 is not satisfied by asserting that they fail; paste the output.
- `dotnet test -c Release` across the solution, counts pasted.
- `git diff --check`.
- Confirm no consumer of `ClaimOwnershipVerifier.Verify` needed a call-site change: `PacketDraftCommand`, `AutomationQueueSeedFromPacketCommand`, `WorkerIssuePreflightCommand`, `IntentNextSliceCommand`, `WorkerClaimCommand`, `IssuePublishFlowCommand`, `WorkerNextActionCommand`. If any did, say which and why in the PR body.

## Related Links

- G747 (PR #1627, issue #1625) — fixed the writer half; its closeout recorded this reader residual as a separate-unit candidate.
- G680 — introduced the shared claim-gate judgment consumed by the seven surfaces above.
- G679 (PR #1471) — the push-CAS claim primitive; the writer contract this reader must agree with.
- G743 (PR #1620, issue #1617) and G738 (PR #1609) — prior claim-registry correctness units; G743 established that a false negative from the claim registry is worse than the original defect, because operators act on the wrong answer.

## Knowledge Maintenance

- ADR candidate: none. This implements a decision G679 and G747 already made; it does not choose anew.
- Diagram candidate: none.
- Docs update: `docs/en/05-implementation-loop.md`, `docs/ja/05-implementation-loop.md` — both currently describe the current-branch read and become false with this change.
- Closeout writeback expected: yes — node 03 records that the G747 asymmetry is closed, and that the reader's failure had two opposite directions (fail-open on a branch with no store, false-unheld on a stale one).

## Guide Reachability (G645)

`no_role_facing_surface: true`. This slice adds no command, subcommand, flag, or output field. It corrects the behavior of existing surfaces (`claim verify` and the seven claims-gated commands) whose guide routes already exist and are unchanged. The two documentation mirrors are updated under Knowledge Maintenance above, not as a new route.

## Base Branch Policy

Policy: `direct-main`

- Child isolation: do not inspect or mutate host `.intent-cli/`, queue-state, runs, packets, intent tree, or host Git. The rendered GitHub issue body is the whole contract.

tomohisa:
Closes #1642

## Summary

This PR fixes the G755 claim-reader asymmetry:
- `ClaimCommand.ResolveRemoteDefaultBranch` is the single shared remote-default-branch resolver.
- `ClaimCommand.RunTransaction` and `ClaimOwnershipVerifier.ReadCanonicalEvidence` both call that resolver.
- The reader fetches and reads `refs/remotes/origin/<resolved-default-branch>`, independent of the checkout branch or detached HEAD.
- Resolution or canonical fetch failure returns `canonical-unavailable` and never `not-configured`.
- `not-configured` remains reserved for a canonical branch with no claims store, preserving the legacy output.
- G747 writer behavior and tests are unchanged.

No `branch --show-current` call remains in `ClaimOwnershipVerifier.cs`.

## Required pre-fix reproduction

The new real bare-origin tests were added before the production change and run on parent `6ea81ac85e5fc104d5cd954766c916445f751183`. The actual focused result was:

```
[xUnit.net ...] ClaimOwnershipVerifierG755Tests.NonDefaultCheckout_ReadsCanonicalClaim_G755 [FAIL]
Expected: "owned"
Actual:   "not-configured"

[xUnit.net ...] ClaimOwnershipVerifierG755Tests.DetachedHead_ReadsCanonicalClaim_G755 [FAIL]
Expected: True
Actual:   False

[xUnit.net ...] ClaimOwnershipVerifierG755Tests.NonDefaultCheckout_WithNoLocalClaimsStore_DoesNotFailOpen_G755 [FAIL]
Assert.NotEqual() Failure:
Expected: Not "not-configured"
Actual:       "not-configured"

[xUnit.net ...] ClaimOwnershipVerifierG755Tests.CanonicalBranchResolutionFailure_FailsClosed_G755 [FAIL]
Assert.False() Failure:
Expected: False
Actual:   True

Failed! - Failed: 4, Passed: 1, Skipped: 0, Total: 5
```

The canonical no-store compatibility case was the one parent test that passed.

## Resolver and consumer inventory

There is one resolver implementation, `ResolveRemoteDefaultBranch`, and two production call sites:
1. `ClaimCommand.RunTransaction` (writer).
2. `ClaimOwnershipVerifier.ReadCanonicalEvidence` (reader).

`TryParseRemoteDefaultBranch` remains only the parser used by that resolver; it is not a second branch-resolution path.

The seven named `ClaimOwnershipVerifier.Verify` consumers required no call-site changes:
- `PacketDraftCommand`
- `AutomationQueueSeedFromPacketCommand`
- `WorkerIssuePreflightCommand`
- `IntentNextSliceCommand` (two existing call sites)
- `WorkerClaimCommand`
- `IssuePublishFlowCommand`
- `WorkerNextActionCommand` (two existing call sites)

## Documentation

Both `docs/en/05-implementation-loop.md` and `docs/ja/05-implementation-loop.md` now describe the shared canonical default-branch resolver, canonical fetch, detached-HEAD behavior, fail-closed `canonical-unavailable`, and byte-compatible genuine no-store behavior. The focused documentation mirror test also asserts that the old current-branch wording is absent.

## Verification

Commands used `NUGET_PACKAGES=/private/tmp/g755-nuget NUGET_HTTP_CACHE_PATH=/private/tmp/g755-http-cache`.

- G755 focused real-bare-origin regressions: Passed 6, Failed 0, Skipped 0, Total 6.
- Unmodified `ClaimCommandG747Tests`: Passed 4, Failed 0, Skipped 0, Total 4.
- Unmodified `ClaimCommandG743Tests`: Passed 2, Failed 0, Skipped 0, Total 2.
- Adjacent claim tests (`ClaimCommandG679Tests`, `G680ClaimConsumerTests`, `G717WorkerPathTests`, G743, G747): Passed 29, Failed 0, Skipped 0, Total 29.
- G613 Japanese terminology guard: Passed 6, Failed 0, Skipped 0, Total 6.
- Full Release suite: Passed 5342, Failed 0, Skipped 1, Total 5343.
- `git diff --check`: clean.

The child recorded the supplied host authority for execution-unit:G755. Its child worker selection returned `action=wait`, `source_classification=claim-refused`, because the child registry reported holder none/unheld; issue-preflight was `actionable=true`, `classification=ready-to-implement`, `claim_status=unheld`. This is the known child/host contradiction: orchestration next-action/issue-preflight could not verify the host claim because `.git/FETCH_HEAD` could not be opened, while the supplied host origin/main claim is authoritative. No execution-unit claim was created, modified, verified, or released in the child product repo.

### DIFF
--- a/docs/en/05-implementation-loop.md
+++ b/docs/en/05-implementation-loop.md
@@ -211,12 +211,15 @@ line after broad union rules only through an explicit reviewed commit.
 ### Preview: claim-aware start surfaces (G680)
 
 Packet draft, queue seed/publish-flow, worker next-action, and release-prep use
-the same `claim verify` judgment. In a Git worktree the verifier first fetches
-the current branch and reads the claims tree from the fresh `origin` ref; local
-absence or a stale local record is never proof of ownership or no-store. If
-fresh canonical Git evidence cannot be established it fails closed. A genuine
-canonical no-store host keeps legacy single-team output byte-identical. A
-configured store requires the invoking team to hold the matching scope;
+the same `claim verify` judgment. In a Git worktree the verifier uses the same
+remote-default-branch resolver as claim acquire, fetches the canonical branch
+from origin, and reads the claims tree from its fresh `origin` ref; the reading
+checkout's current branch, local absence, or a stale local record is never
+proof of ownership or no-store. This also gives a detached checkout the same
+canonical answer. If the default branch cannot be resolved or fetched, the
+verifier fails closed with `canonical-unavailable`; a genuine canonical
+no-store host keeps legacy single-team output byte-identical. A configured
+store requires the invoking team to hold the matching scope;
 unheld and other-team refusals name scope, holder, and holder team. Next-slice
 uses that same judgment in recommendation mode: unheld and own-team units
 remain candidates, while claimed-elsewhere units are excluded with holder
--- a/docs/ja/05-implementation-loop.md
+++ b/docs/ja/05-implementation-loop.md
@@ -201,11 +201,13 @@ existing host は自動 migrate しません。最後の specific line は broad
 ### Preview: claim-aware start surface (G680)
 
 packet draft、queue seed/publish-flow、worker next-action、release-prep は同じ
-`claim verify` judgment を使います。Git worktree では verifier が current branch を最初に
-fetch し、fresh な `origin` ref の claims tree を読みます。local absence や stale local record を
-ownership / no-store の証拠にはせず、fresh canonical Git evidence を確立できなければ fail closed
-します。canonical に claims store が存在しない host だけは legacy single-team output を
-byte-identical に維持します。store が設定されている場合、invoking team が matching scope を
+`claim verify` judgment を使います。Git worktree では verifier が claim acquire と同じ
+remote-default-branch resolver を使って origin の canonical branch を解決し、そこを fetch して
+fresh な `origin` ref の claims tree を読みます。読む checkout の current branch、local absence、
+stale local record は ownership / no-store の証拠ではありません。detached checkout でも同じ
+canonical answer を返します。default branch を解決または fetch できなければ
+`canonical-unavailable` で fail closed します。canonical に claims store が存在しない host
+だけは legacy single-team output を byte-identical に維持します。store が設定されている場合、invoking team が matching scope を
 保持している必要があります。unheld / other-team の拒否は scope、holder、holder team を
 名指します。next-slice は同じ judgment を recommendation mode で使い、unheld と own-team
 unit は candidate のまま、claimed-elsewhere unit は holder evidence 付きで除外します。
--- a/src/IntentSystem.Cli/Commands/ClaimCommand.cs
+++ b/src/IntentSystem.Cli/Commands/ClaimCommand.cs
@@ -114,22 +114,9 @@ internal static class ClaimCommand
         ArgumentNullException.ThrowIfNull(warningWriter);
         deleteDirectory ??= path => Directory.Delete(path, recursive: true);
 
-        var origin = RunGit(repoRoot, ["remote", "get-url", "origin"]);
-        EnsureSuccess(origin, "resolve origin");
-        var defaultBranchResult = RunGit(repoRoot, ["ls-remote", "--symref", "origin", "HEAD"]);
-        EnsureSuccess(defaultBranchResult, "resolve origin default branch");
-        var remote = origin.StandardOutput.Trim();
-        if (remote.Length == 0)
-        {
-            throw new InvalidOperationException("claim requires an origin remote");
-        }
-
-        if (!TryParseRemoteDefaultBranch(defaultBranchResult.StandardOutput, out var defaultBranch))
-        {
-            throw new InvalidOperationException(
-                "Could not resolve origin default branch from its HEAD symref; refusing to "
-                + "fall back to the current branch.");
-        }
+        var canonicalBranch = ResolveRemoteDefaultBranch(repoRoot);
+        var remote = canonicalBranch.Remote;
+        var defaultBranch = canonicalBranch.Name;
         if (!request.Write)
         {
             return new ClaimTransactionResult(
@@ -668,6 +655,33 @@ internal static class ClaimCommand
         return true;
     }
 
+    /// <summary>
+    /// Resolve the one canonical remote branch used by both claim writes and
+    /// claim reads. The caller must fail closed when this metadata is absent
+    /// or unsafe; no caller may substitute its current checkout branch.
+    /// </summary>
+    internal static ClaimRemoteDefaultBranch ResolveRemoteDefaultBranch(string repoRoot)
+    {
+        var origin = RunGit(repoRoot, ["remote", "get-url", "origin"]);
+        EnsureSuccess(origin, "resolve origin");
+        var defaultBranchResult = RunGit(repoRoot, ["ls-remote", "--symref", "origin", "HEAD"]);
+        EnsureSuccess(defaultBranchResult, "resolve origin default branch");
+        var remote = origin.StandardOutput.Trim();
+        if (remote.Length == 0)
+        {
+            throw new InvalidOperationException("claim requires an origin remote");
+        }
+
+        if (!TryParseRemoteDefaultBranch(defaultBranchResult.StandardOutput, out var defaultBranch))
+        {
+            throw new InvalidOperationException(
+                "Could not resolve origin default branch from its HEAD symref; refusing to "
+                + "fall back to the current branch.");
+        }
+
+        return new ClaimRemoteDefaultBranch(remote, defaultBranch);
+    }
+
     private static bool IsSafeRemoteBranch(string value)
     {
         if (value.Length == 0
@@ -968,6 +982,8 @@ internal static class ClaimCommand
         writer.WriteLine($"- detail: {result.Detail}");
     }
 }
+
+internal sealed record ClaimRemoteDefaultBranch(string Remote, string Name);
 
 internal enum ClaimOperation { Acquire, Release, Takeover }
 
--- a/src/IntentSystem.Cli/Commands/ClaimOwnershipVerifier.cs
+++ b/src/IntentSystem.Cli/Commands/ClaimOwnershipVerifier.cs
@@ -184,27 +184,27 @@ internal static class ClaimOwnershipVerifier
             return ReadLocalEvidence(repoRoot, scope);
         }
 
-        var branch = RunGit(repoRoot, ["branch", "--show-current"]);
-        if (branch.ExitCode != 0 || string.IsNullOrWhiteSpace(branch.StandardOutput))
+        ClaimRemoteDefaultBranch canonicalBranch;
+        try
         {
-            return CanonicalClaimEvidence.Unavailable("the current Git checkout has no named branch");
+            canonicalBranch = ClaimCommand.ResolveRemoteDefaultBranch(repoRoot);
+        }
+        catch (Exception exception) when (exception is IOException or InvalidOperationException)
+        {
+            return CanonicalClaimEvidence.Unavailable(
+                string.IsNullOrWhiteSpace(exception.Message)
+                    ? "resolving origin default branch failed"
+                    : exception.Message);
         }
 
-        var origin = RunGit(repoRoot, ["remote", "get-url", "origin"]);
-        if (origin.ExitCode != 0 || string.IsNullOrWhiteSpace(origin.StandardOutput))
-        {
-            return CanonicalClaimEvidence.Unavailable("the current Git checkout has no origin remote");
-        }
-
-        var branchName = branch.StandardOutput.Trim();
-        var remoteRef = $"refs/remotes/origin/{branchName}";
+        var remoteRef = $"refs/remotes/origin/{canonicalBranch.Name}";
         var fetch = RunGit(repoRoot,
-            ["fetch", "--quiet", "origin", $"+refs/heads/{branchName}:{remoteRef}"]);
+            ["fetch", "--quiet", "origin", $"+refs/heads/{canonicalBranch.Name}:{remoteRef}"]);
         if (fetch.ExitCode != 0)
         {
             return CanonicalClaimEvidence.Unavailable(
                 string.IsNullOrWhiteSpace(fetch.StandardError)
-                    ? "fetching the current branch from origin failed"
+                    ? "fetching the canonical default branch from origin failed"
                     : fetch.StandardError.Trim());
         }
 
--- a/src/IntentSystem.Cli/Commands/WorkerClaimCommand.cs
+++ b/src/IntentSystem.Cli/Commands/WorkerClaimCommand.cs
@@ -193,15 +193,15 @@ internal static class WorkerClaimCommand
         }
 
         var store = ClaimOwnershipVerifier.ProbeStore(context.RepoRoot);
-        if (!store.StoreConfigured)
-        {
-            return null;
-        }
         if (!store.Available)
         {
             return ClaimOwnershipVerifier.Unavailable(
                 $"execution-unit:issue-{number}",
                 $"claim verification refused issue #{number}: fresh canonical Git evidence is unavailable ({store.Detail}).");
+        }
+        if (!store.StoreConfigured)
+        {
+            return null;
         }
 
         try
--- a/tests/IntentSystem.Cli.Tests/ClaimOwnershipVerifierG755Tests.cs
+++ b/tests/IntentSystem.Cli.Tests/ClaimOwnershipVerifierG755Tests.cs
@@ -1,0 +1,220 @@
+using System.Diagnostics;
+using System.Text.Json;
+using IntentSystem.Cli;
+using IntentSystem.Cli.Commands;
+using IntentSystem.Cli.Models;
+
+namespace IntentSystem.Cli.Tests;
+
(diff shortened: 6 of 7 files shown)


tomohisa (commented):
REQUEST-UPDATE

Implementation-finding — G755 AC 5 requires canonical-branch resolution or fetch failure to fail closed, and the packet explicitly makes the shared `ClaimOwnershipVerifier.Verify` judgment the gate for all seven consumers. `WorkerClaimCommand` remains a fail-open consumer at exact head `e50759fdaa97784bba9ceaf55da22502912277a9`.

`ClaimOwnershipVerifier.ReadCanonicalEvidence` now correctly returns `CanonicalClaimEvidence.Unavailable`, whose values are `Available=false` and `StoreConfigured=false`, when resolving or fetching the canonical branch fails (`ClaimOwnershipVerifier.cs:190-208`, `:324-325`). But `WorkerClaimCommand.ResolveIssueClaimVerification` checks `!store.StoreConfigured` first and returns `null` (`WorkerClaimCommand.cs:195-199`), before its `!store.Available` refusal branch (`:200-205`). `WorkerClaimAnalyzer` receives no verification, so an otherwise eligible issue can proceed and add `intent-issue-in-progress` instead of stopping on `canonical-unavailable`. This is precisely the unsafe distinction the new reader contract is meant to eliminate; `WorkerNextActionCommand`, the other `ProbeStore` consumer, checks `Available` first and correctly waits.

Please make unavailable canonical evidence win over the genuine-no-store bypass (normally by testing `!store.Available` before `!store.StoreConfigured`) and add a regression that exercises `WorkerClaimCommand` through canonical-resolution and/or fetch failure on an otherwise eligible issue. It must assert refusal/no label mutation. Retain and test the genuine canonical no-store case (`Available=true`, `StoreConfigured=false`) so legacy no-store behavior still proceeds. This is an implementation-finding against AC 5 and the seven-consumer contract, not a request to alter claim schema, writer target, or labels outside the command's existing decision.

The rest of the G755 reader repair is verified: there is exactly one `ClaimCommand.ResolveRemoteDefaultBranch` implementation and exactly two production calls (writer `RunTransaction` and reader `ReadCanonicalEvidence`); no reader `branch --show-current` remains; reader fetches `origin/<resolved-default>`; detached HEAD reads canonical state; and G747 writer tests are byte-unmodified. I independently reproduced the pre-head real bare-origin behavior by injecting the new tests into parent `6ea81ac8`: the owned read became `not-configured`, the empty-local-store case fail-opened, detached HEAD failed, and canonical-resolution failure passed; genuine canonical no-store passed. At head, `ClaimOwnershipVerifierG755Tests` pass 6/6, G747 4/4, and G743 2/2; CI is green and `git diff --check` is clean. Those tests are necessary but insufficient because none drives the `WorkerClaimCommand` `ProbeStore` decision.

`design_alignment_checked=true`: checked G755 packet/review-context and `intents/intent-cli/intent-tree/means/03-state-and-audit-strategy.md`. The node's G747 residual calls for the reader/writer canonical-ref symmetry and fail-closed ownership evidence; this remaining worker-claim bypass contradicts that contract. EN/JA canonical-read docs are updated, scope is otherwise limited to the reader/resolver/docs/tests, and `Closes #1642` is correct.

tomohisa (commented):
APPROVED

Re-review APPROVED-equivalent at exact head `67966c18a9467cee4ee7ad0c269a1a3f88105e8d`.

The sole prior finding is fixed. `WorkerClaimCommand.ResolveIssueClaimVerification` now checks `!store.Available` before `!store.StoreConfigured`. The new `Execute_IssueWithCanonicalUnavailable_RefusesBeforeNoStoreBypassWithoutLabelMutation_G755Repair` creates a real Git root with a broken origin and verifies the command exits `2`, reports a claim-registry refusal/canonical-evidence-unavailable error, has `Proceed=false` and `Applied=false`, produces no label edits, and calls no mutator transition. This closes the fail-open path I identified in the prior review.

The legacy condition remains distinct and unchanged: `ClaimOwnershipVerifierG755Tests.CanonicalNoStore_PreservesLegacyVerificationOutput_G755` uses a real bare-origin canonical host whose default branch has no store, compares its command JSON byte-for-byte with the local legacy result, and asserts `passed=true`, `status=not-configured`, and `store_configured=false`. With the repaired decision order, available/no-store reaches the existing `!store.StoreConfigured` bypass only after the unavailable stop has been ruled out. Thus G755 AC 4 (no fail-open from a branch-local absence) and AC 6 (genuine canonical no-store compatibility) are both satisfied.

The repair scope is exact: compared with prior reviewed head `e50759fd`, only `WorkerClaimCommand.cs` and `WorkerClaimCommandTests.cs` changed. The shared `ClaimCommand.ResolveRemoteDefaultBranch`, `ClaimOwnershipVerifier`, `WorkerNextAction`, G747/G743 test files, and EN/JA implementation-loop mirrors are byte-unchanged. The earlier reader/writer contract remains: one resolver implementation with exactly the writer and reader production call sites; no reader `branch --show-current`; canonical fetch/read; detached HEAD support; and fail-closed verifier behavior. `Closes #1642` is present.

Focused Release tests passed: `WorkerClaimCommandTests` 25/25, and the G755/G747/G743 set 12/12 (including G755 6/6, G747 4/4 unchanged, G743 2/2). `git diff --check` is clean; both source-contract and npm dry-run CI checks are green at this head. Tests are necessary, not sufficient; this approval also rests on the packet/review-context contract and direct two-state evidence above.

`design_alignment_checked=true`: rechecked G755 `packet.yaml` and `review-context.md`, `docs/en/05-implementation-loop.md`, `docs/ja/05-implementation-loop.md`, and `intents/intent-cli/intent-tree/means/03-state-and-audit-strategy.md`. Node 03's G747 residual requires canonical reader/writer symmetry and fail-closed claim evidence; the limited repair now preserves both that safety boundary and the canonical no-store compatibility contract. No node 02/04 or other out-of-scope surface moved.

PR #1648, opened 2026-08-29
G757: add role-scoped notify collect

### LINKED ISSUE #1645: G757: an external role can be sent to but cannot ask what arrived, so it has no canonical way to be woken
# G757: an external role can be sent to but cannot ask what arrived, so it has no canonical way to be woken

## Goal

Give a role with a resolvable reader a canonical way to read what has arrived **for it**, from an explicit cursor, and optionally to block until something new arrives. Today `notify collect` and `notify status` both require `--task-id`, so a role can only ask about a task it already knows exists — which is the one thing an unwoken role cannot know.

## Why This Slice Exists Now

The four-thread model supports a role that is not resident in a herdr pane (`resident: "external"` with a recorded `reader`). Delivery to such a role works: the event is appended to its reader file. What does not exist is the other half — a way for that role to **learn that something arrived**.

A pane-resident role does not need this: herdr delivers the prompt, so being woken and learning the task id are the same event. An external role gets neither. It has a file that grew, and no canonical surface that will tell it so.

The practical consequence is that an external role must either poll a path it derived itself, or be woken through some channel outside the canonical route. Both are worse than the model intends: the first duplicates path resolution the CLI already owns, and the second moves the wake signal off the canonical record, so what a role was told and what the durable log says can drift apart.

This is the missing half of the external-resident model, not a new mode. It is additive: nothing about pane-resident delivery, the session-layer transports, or the existing `--task-id` surfaces changes.

## Current Observed State

Measured with installed `intent-cli 0.27.0-f43fbd1-G753`.

```
$ intent-cli notify collect
Usage: intent-cli notify collect --domain <d> --team <t> --task-id <id> [...]

$ intent-cli notify status
Usage: intent-cli notify status --task-id <id> [--domain <d> --team <t>] [...]
```

- Both surfaces **require** `--task-id`. There is no role-scoped read.
- Neither accepts `--wait`, `--follow`, or `--tail`. There is no blocking receive anywhere in `notify`.
- There is no cursor or since-parameter, so there is no way to ask "what is new".
- `notify` subcommands are `adjudicate, collect, delegate, dispose, escalate, reconcile, report, status`. None of them answers "what has arrived for me".

The reader path itself is already resolvable inside the CLI: `NotifyEventWriter.TryResolveReadPath` (`NotifyCommand.cs:3316-3340`) returns the effective read path for a domain/team given the recorded reader value.

- The external-resident + recorded-reader model is correct and settled.
- Delivery is correct and unchanged. This slice only reads.
- `NotifyEventWriter.TryResolveReadPath` is the resolver for the effective read path and must be reused, not reimplemented. G756 surfaces the same resolver on `topology show` / `validate`; these are siblings and neither blocks the other.
- The reader file is an append-only JSONL event log.
- The four-thread model, `herdr-only` and `agmsg` transports, and every existing team shape are unchanged. Pane-resident roles keep being woken by herdr exactly as today.

## Target Repo / Path / Part

- Target paths: `src/IntentSystem.Cli/Commands/NotifyCommand.cs`, `tests/IntentSystem.Cli.Tests/NotifyCollectCommandTests.cs`, `docs/en/05-implementation-loop.md`, `docs/ja/05-implementation-loop.md`

Target part: the `notify collect` argument surface and its read path, plus the doc paragraphs describing how a non-pane role receives.

## In Scope

- A role-scoped read mode on `notify collect`: `--role <role>` instead of `--task-id`, returning the events addressed to that role from its resolved reader.
- `--since <cursor>`: return only events after the given cursor. The result carries the next cursor.
- `--wait` with a required bounded `--timeout-ms`: block until at least one new event arrives after the cursor, or the timeout expires.
- The effective reader path comes from `NotifyEventWriter.TryResolveReadPath`.
- Both documentation mirrors describe the role-scoped read and the cursor.

## Out Of Scope

- Any change to delivery, to what is written, or to where it is written.
- Any change to the existing `--task-id` behavior of `collect` or `status`. It must stay byte-identical.
- Any change to pane-resident wake behavior, herdr integration, or `notify supervise`.
- A daemon, a background watcher, a scheduler, or any process that outlives the command.
- Server-side or file-side acknowledgement state. The cursor is supplied by the caller and returned in the result; the CLI stores no per-reader read position.
- Any new transport, and any change to session-layer mode selection.

The child PR adds a role-scoped read mode to `intent-cli notify collect` so that a role with a resolvable reader can ask what has arrived for it without already knowing a task id. The new mode takes `--role` in place of `--task-id`, accepts an optional opaque `--since` cursor and returns the next cursor in its result, and accepts `--wait` together with a required bounded `--timeout-ms` that blocks until a new event arrives or the timeout expires. The effective reader path must come from the existing `NotifyEventWriter.TryResolveReadPath` rather than a new resolution. Invoking `collect` with `--task-id` must behave byte-identically to today, and the command must remain a single short-lived process that leaves nothing running. The PR must add tests covering cursor resumption without loss or duplication, a wait that returns when an event is appended, a wait that times out cleanly with no events, a missing reader file, and an unhonourable cursor being reported rather than silently reset.

## Acceptance Criteria

1. `notify collect --domain <d> --team <t> --role <role>` returns the events for that role from its resolved reader, without `--task-id`.
2. The effective reader path is obtained by calling the existing `NotifyEventWriter.TryResolveReadPath`. A second independent path computation is a failed criterion even if it returns the same string.
3. `--since <cursor>` returns only events after that cursor, and every result carries the next cursor. A test drives two successive calls across an append and asserts **no event is lost and none is returned twice**.
4. A cursor that can no longer be honoured is reported as an explicit, named result. It must **not** be silently reset to the start (which would re-deliver everything) or to the end (which would skip events). A test asserts the explicit result.
5. `--wait` requires `--timeout-ms`; a `--wait` without it is refused with a usage error rather than blocking indefinitely.
6. `--wait` returns promptly once a matching event is appended. A test appends from another writer while the command is blocked and asserts it returns with that event.
7. `--wait` that reaches its timeout exits with an explicit "no new events" result. A test asserts the result and the exit code, and that this is distinguishable from an error.
8. Without `--wait` the command performs a single non-blocking read and returns immediately, even when there are no new events.
9. A reader file that does not exist yet is handled as "no events" — not an error, and not a crash. A test covers it.
10. `notify collect --task-id <id>` behaves byte-identically to the parent commit. A test pins the existing shape, and the existing collect tests remain unmodified and green.
11. The command leaves no process, watcher, or timer running after it exits.
12. Each new test is demonstrated failing or absent on the parent commit, with the actual output pasted in the PR.
13. Both doc mirrors describe the role-scoped read, the cursor, and the bounded wait.
14. Full Release suite passes with the actual counts pasted.

## Verification

- New tests in `tests/IntentSystem.Cli.Tests/NotifyCollectCommandTests.cs` covering criteria 3, 4, 6, 7, 8, 9 and 10.
- The criterion 6 test must have a genuine second writer appending while the command blocks; a test that pre-writes the event and then calls `--wait` proves only that a non-empty read returns, not that waiting works.
- `git diff --check`.

## Related Links

- G756 (issue #1644) — sibling: surfaces the same effective reader path on `topology show` / `validate`. Neither blocks the other.
- G755 (issue #1642, PR #1643) — the claim-reader unit from the same investigation.
- `intents/intent-cli/intent-tree/means/08-agent-message-orchestration.md` — the node this slice supports.

## Knowledge Maintenance

- ADR candidate: **yes** — "External roles receive by cursor-based pull, not by pushed acknowledgement state". Target `docs/adr/`. The choice to keep the cursor caller-supplied, with no CLI-side read position, is hard to reverse once callers depend on it and deserves its rationale recorded.
- Diagram candidate: none.
- Docs update: `docs/en/05-implementation-loop.md`, `docs/ja/05-implementation-loop.md`
- Closeout writeback expected: yes — node 08 records that the external-resident model has both halves.

## Guide Reachability (G645)

`no_role_facing_surface: false`. This slice adds a role-facing read mode.

- guide_surface: `guide workflow task orchestrator-thread`; role: `design`; target_surface: `notify collect --role/--since/--wait` as the canonical receive for a non-pane role.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1645

## Summary

- Adds `notify collect --role` for external-resident roles.
- Adds caller-owned opaque `--since` / `next_cursor` resumption with prefix and complete-line integrity checks.
- Adds bounded synchronous `--wait --timeout-ms`; timeout is an explicit non-error `no-new-events` result.
- Uses `NotifyEventWriter.TryResolveReadPath` for effective scoped/legacy reader resolution, including re-resolution during a bounded wait.
- Leaves the existing `--task-id` collection path and result shape unchanged.
- Adds the cursor-based pull ADR and matching EN/JA implementation-loop guidance.

## Verification

- Parent baseline: `git show HEAD:tests/IntentSystem.Cli.Tests/NotifyCollectCommandTests.cs` returned `fatal: path ... does not exist in 'HEAD'`; installed `intent-cli 0.27.0-f43fbd1-G753` rejected the new invocation with `invalid-notification: Unknown argument '--role'`.
- Focused role-collect tests: 7 passed, 0 failed.
- Adjacent notify/event-stream suites: 37 passed, 0 failed.
- G613 Japanese terminology guard: 6 passed, 0 failed.
- Full Release suite: Failed 0, Passed 5350, Skipped 1, Total 5351.
- `git diff --check`: passed.
- The wait regression uses a genuine second `NotifyEventWriter.Append` while collect is blocked; missing-reader, cursor resumption, cursor refusal, timeout, required-bound, and task-id compatibility are covered.

## Worker boundary

The supplied authoritative host evidence is consumed: origin/main carries execution-unit:G757 for actor=implementation/team=intent-cli-dev at host commit `c8099f0fd2395fc923e9604970db7b229769570c`. The child GitHub-only selector returned `action=wait` because its local execution-unit:G757 registry is unheld, while child issue-preflight returned `actionable=true`, `classification=ready-to-implement`, `claim_status=unheld`, and the same local unheld refusal. This is the known child/host registry contradiction; no child execution-unit claim was created or mutated.

### DIFF
--- a/docs/adr/0012-external-role-cursor-pull.md
+++ b/docs/adr/0012-external-role-cursor-pull.md
@@ -1,0 +1,55 @@
+# ADR 0012: External roles receive by cursor-based pull
+
+- Status: Accepted (preview-through-1.x)
+- Date: 2026-08-29
+- Deciders: Design, orchestration, implementation, and review; recorded by G757
+- Related: G300, G578, G681, G756, G757 / #1645
+
+## Context
+
+An external-resident role has a recorded reader and can already receive a
+durable notification event. It cannot, however, ask what arrived without
+already knowing the task id. Polling a path derived by the role duplicates the
+CLI's reader-resolution rules, while a pushed acknowledgement position would
+add state and a second authority to the receive path.
+
+## Decision
+
+1. `notify collect` gains a role-scoped mode using `--role <role>` instead of
+   `--task-id`. It reads the role's recorded external reader after resolving
+   the effective path through `NotifyEventWriter.TryResolveReadPath`.
+2. The caller may supply an opaque `--since <cursor>`. The result returns a
+   `next_cursor`; the CLI stores no acknowledgement, watermark, or per-reader
+   read position. Resumption therefore remains explicit at the call site and
+   is deterministic without a server-side receive state.
+3. `--wait` is an optional bounded receive operation and requires
+   `--timeout-ms`. A timeout returns the explicit non-error `no-new-events`
+   result. A missing reader is `no-events`, and a cursor that cannot identify
+   an intact position is the explicit `cursor-unhonourable` refusal. Neither
+   case silently resets or skips the caller's position.
+4. The command is a short-lived synchronous read/poll. It creates no daemon,
+   watcher, timer, process, or durable acknowledgement state. Pane-resident
+   wake delivery, transports, supervise, and the existing task-id collection
+   path remain unchanged.
+
+## Consequences
+
+- External roles have a canonical receive surface that can discover work
+  through the durable event stream rather than an out-of-band wake channel.
+- Caller-owned cursors make no-loss/no-duplicate resumption possible while
+  keeping acknowledgement policy outside the CLI.
+- A reader replacement, truncation, or malformed cursor is visible as an
+  explicit refusal instead of causing replay or silent loss.
+- The reader path remains compatible with scoped and legacy event locations
+  because path selection is delegated to the existing shared resolver.
+
+## Rejected alternatives
+
+- Add a second path computation in `notify collect`: rejected because delivery
+  and reading must share `TryResolveReadPath`.
+- Persist a per-role acknowledgement file: rejected because it would create
+  server-side receive state and make caller progress implicit.
+- Reset an invalid cursor to the start or end: rejected because either choice
+  duplicates already-consumed events or silently loses events.
+- Run a background watcher or daemon: rejected because receive is an explicit
+  bounded command and must leave no process or watcher after exit.
--- a/docs/en/05-implementation-loop.md
+++ b/docs/en/05-implementation-loop.md
@@ -53,6 +53,32 @@ copy a long loop body from this document.
 - Select work only via `intent-cli worker next-action`; process at most one action per wake
 - All label transitions go through `intent-cli worker` — never raw `gh ... --add-label`
 
+## G757: external-role cursor-based receive
+
+An external-resident role can receive a durable event without already knowing
+the task id. Use the role-scoped read mode:
+
+```text
+intent-cli notify collect --domain <domain> --team <team> --role <role> \
+  [--since <opaque-cursor>] [--wait --timeout-ms <milliseconds>] --format json
+```
+
+The role must have a recorded external reader. The command resolves that
+reader through `NotifyEventWriter.TryResolveReadPath`, so scoped-versus-legacy
+reader compatibility stays in one place. It performs one immediate read unless
+`--wait` is supplied; `--wait` always requires a bounded `--timeout-ms` and
+returns `cause: "no-new-events"` with a non-error exit when the bound expires.
+A missing reader file is `cause: "no-events"`, not an error.
+
+The cursor is opaque and caller-owned: the CLI stores no acknowledgement or
+read position. Pass the returned `next_cursor` to the next call to resume
+without loss or duplication. If the cursor no longer identifies an intact
+position in the same reader, the command returns the explicit
+`cursor-unhonourable` result and never resets to the beginning or skips to the
+end. The existing `--task-id` collection path and result shape remain
+unchanged. Role collection is a short-lived command; it leaves no watcher,
+timer, process, or other receive state behind.
+
 ## Metadata / label safety
 
 - A child agent never manually applies workflow labels to a PR. For issue-to-PR `pr-created`, canonical `worker complete --kind issue --outcome pr-created --github-only --write` may apply target-repository `intent-target` to the PR; `intent-pr-created` remains an issue-side marker.
--- a/docs/ja/05-implementation-loop.md
+++ b/docs/ja/05-implementation-loop.md
@@ -52,6 +52,32 @@
 - 作業の選択は `intent-cli worker next-action` のみ。1 wake で最大 1 アクション
 - label 遷移はすべて `intent-cli worker` 経由 — raw `gh ... --add-label` は使わない
 
+## G757: external role の cursor-based receive
+
+external-resident の role は、まだ task id を知らなくても durable event を
+受信できます。role-scoped read mode を使います:
+
+```text
+intent-cli notify collect --domain <domain> --team <team> --role <role> \
+  [--since <opaque-cursor>] [--wait --timeout-ms <milliseconds>] --format json
+```
+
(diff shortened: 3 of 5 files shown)


tomohisa (commented):
APPROVED — semantic review of G757 / #1648 at `dd1c20d869b7f2fea338c601ed05b5c98d35b6e6`.

Independent evidence:

- AC1–2: role-scoped collect takes `--role` in place of `--task-id`; its sole effective-reader resolution is the `NotifyEventWriter.TryResolveReadPath` call in `ExecuteRoleCollect` (`NotifyCommand.cs:606`). The topology helper only validates/resolves the recorded path before that shared resolver is invoked.
- AC3–4: the focused test reads two records, appends a third, resumes from `next_cursor`, receives only the third, and then receives none. Cursors bind the recorded reader digest, byte offset, complete-line count, and prefix digest; truncation yields explicit `cursor-unhonourable` / exit 1 before any reset or skip.
- AC5–9: `--wait` without `--timeout-ms` is validation/usage error; the bound is 1–300000 ms. Timeout returns exit 0 with `outcome`, `cause` `no-new-events`, and `timed_out: true`; non-wait is immediate; a missing reader is exit-0 `no-events` and can later resume when the file appears.
- AC6 specifically: `WaitReturnsAfterGenuineSecondWriterAppends` starts an independent writer task, delays its append by 100 ms, then runs the synchronous collect with a 2 s bound. It requires the appended event and 50–1500 ms elapsed, so an implementation that returns without waiting yields `no-events` and fails. The focused suite passed 7/7.
- AC10–11: the existing task-id code remains on its prior branch and existing test files are unmodified; the new compatibility assertion retains the legacy result shape. The role wait is an in-process bounded `Thread.Sleep` poll only—no watcher, timer, process, or background task is created.
- Docs/ADR: both EN and JA mirrors and ADR 0012 agree on caller-owned opaque cursors, shared resolution, bounded wait, explicit timeout/refusal, and unchanged task-id/pane behavior.

I independently confirmed the parent lacks the new focused test, and installed `intent-cli 0.27.0-f43fbd1-G753` rejects `--role` as unknown. `git diff --check` is clean; CI is green at the exact head. Local checks passed: focused role collect 7/7; adjacent notify/event-stream/terminology selection 36/36; G613 Japanese terminology guard 6/6.

design_alignment_checked=true — checked issue #1645 acceptance criteria and scope, G757 packet/review context, ADR 0012, intent node 08, EN/JA docs, parent evidence, focused/adjacent tests, diff, and green CI.

PR #1652, opened 2026-08-29
G758: use containment for checkout freshness

### LINKED ISSUE #1647: G758: the checkout-freshness probe calls any branch that is not the default branch tip stale, and tells the operator to sync something that is already current
# G758: the checkout-freshness probe calls any branch that is not the default branch tip stale, and tells the operator to sync something that is already current

## Goal

Make `automation stalled-work`'s checkout-freshness probe report `stale` only when the checkout is genuinely missing canonical commits, instead of whenever the local HEAD sha differs from the remote default-branch tip sha. A checkout that already contains the default branch must report `current` and stay silent — without the probe fetching anything, because it is contractually read-only.

## Why This Slice Exists Now

The probe exists so a host-state answer carries a warning when it was computed from a checkout that is behind canonical. That is a real and useful guarantee, and it must keep working.

But its test is sha equality against the remote default-branch tip. Any checkout that is *ahead* of the default branch — a long-lived seat branch that has merged everything and added its own commits, which is the healthy steady state for such a branch — fails that test and is reported `stale`. The attached instruction is "Sync the checkout and re-run the report", which in that state is not actionable: there is nothing to sync, and following it changes nothing.

A warning that fires on a healthy state and asks for an impossible action is worse than no warning. It trains the reader to skip the banner, which is the same banner that carries the true positives. That is the cost being paid: the true-positive case is correct and valuable, and the false positive is eroding it.

This is the deliberately-deferred half of G755. That unit fixed the claim reader and explicitly excluded this probe, because the mechanism and the fix are different: the claim reader was resolving the wrong ref, whereas this probe resolves the right ref and applies the wrong predicate to it.

## Current Observed State

Measured with installed `intent-cli 0.27.0-f43fbd1-G753` in a purpose-built clean-room host. `checkout_freshness` is omitted from the JSON when the status is `current`, so an absent field means current and silent.

| Case | Checkout | Reported | Correct |
| --- | --- | --- | --- |
| A | tip identical to `origin/<default>` tip | `current` (silent) | yes |
| B | **ahead**: `origin/<default>` IS an ancestor of HEAD, plus one local commit | **`stale`** | **no — nothing canonical is missing** |
| C | genuinely behind `origin/<default>` | `stale` | yes — true positive, must be preserved |
| D | detached AT the `origin/<default>` tip | `current` (silent) | yes |

Case B is the defect. Cases A, C and D must not change.

The probe reads `git rev-parse HEAD`, then `git ls-remote --symref origin HEAD`, and compares the two shas for equality (`CheckoutFreshnessProbe.cs`, the `string.Equals(localHead, remoteHead, ...)` branch). `git merge-base --is-ancestor` is never used. The warning text is `"checkout freshness: stale; this answer was computed from local HEAD <x> but origin/<default> is <y>. Sync the checkout and re-run the report; this command is read-only and does not fetch, pull, reset, or sync."`

- The probe is and must remain **read-only**: it must not fetch, pull, reset, or sync. That is stated in its own class comment and in the warning text it emits, and callers rely on it. Adding a fetch to obtain ancestry information is not an acceptable fix.
- `git ls-remote --symref origin HEAD` remains the remote query. Resolving the default branch this way is correct and agrees with the claim writer; it is not what this slice changes.
- The three statuses `current`, `stale`, `unknown` stay as they are. `current` remains silent — the `checkout_freshness` field stays omitted and no warning is emitted.
- `unknown` semantics are unchanged: a probe that cannot reach the remote or read local HEAD still reports `unknown` with its reason, and `unknown` is never treated as current.
- The only production call site is `AutomationStalledWorkCommand`; `stalled-work` stays read-only and keeps its existing output shape (`items`, `excluded`, `warnings`, `stalled`, `checkout_freshness`).
- G755 has landed. This slice does not revisit the claim reader.

## Target Repo / Path / Part

- Target paths: `src/IntentSystem.Cli/Commands/CheckoutFreshnessProbe.cs`, `tests/IntentSystem.Cli.Tests/AutomationStalledWorkCommandTests.cs`

Target part: the staleness predicate in `CheckoutFreshnessProbe.Capture` and the warning text it produces.

## In Scope

- Replace sha-equality with a containment test: the checkout is `current` when the remote default-branch tip is contained in the local HEAD's history, and `stale` only when it is not.
- Do it **without fetching**, using the constraint below.
- Keep the `stale` warning text actionable, since after this change it only appears when syncing genuinely is the right advice.
- Tests for all four cases above.

## Out Of Scope

- Any fetch, pull, reset, sync, or other mutation inside the probe.
- Any change to how the default branch is resolved.
- Any new status value, and any change to `current` staying silent.
- Any change to `unknown` handling, the timeout, or the `IGitRemoteCommandRunner` seam.
- Any change to `stalled-work`'s items, exclusions, thresholds, or the rest of its output.
- The claim reader (G755, landed) and the topology reader surfaces (G756, G757).

The child PR changes `CheckoutFreshnessProbe` so that a checkout is reported `stale` only when the remote default-branch tip is genuinely absent from the local HEAD's history, rather than whenever the two shas differ. A checkout that is ahead of the default branch — one that already contains it — must report `current` and stay silent. The probe must remain strictly read-only: it must not fetch, and the containment test must therefore rely only on objects already present locally. It may use the fact that a commit reachable from HEAD is necessarily present in the local object store, so a remote tip whose object is absent locally is provably not contained and is genuinely stale. The `current`/`stale`/`unknown` vocabulary, the silence of `current`, `unknown` handling, default-branch resolution, and all other `stalled-work` output must be unchanged. Tests must cover: identical tips, ahead-and-containing, genuinely behind, and detached at the tip.

## Acceptance Criteria

1. A checkout whose HEAD **contains** the remote default-branch tip reports `current`: the `checkout_freshness` field is omitted and no freshness warning is emitted. This is case B and is the defect being fixed.
2. A checkout that does **not** contain the remote default-branch tip reports `stale`. This is case C and must be preserved — a test asserts it explicitly.
3. Cases B and C are both asserted in the same test class. They are the discriminating pair: before this change they produce the same `stale` result, and after it they must **diverge**. A PR that reports the same status for both, in either direction, has not fixed the defect.
4. A checkout whose tip is identical to the remote tip still reports `current` and stays silent (case A).
5. A detached HEAD at the remote tip still reports `current` and stays silent (case D).
6. **The probe issues no fetch, pull, reset, or sync.** A test asserts the exact git argument sequence the probe runs through its command-runner seam, and that sequence contains no such command. This is a contract, not an implementation preference.
7. The containment test uses only locally available objects. When the remote tip's object is not present locally the result is `stale`, justified by the fact that a commit reachable from HEAD is necessarily in the local object store — so an absent object proves non-containment. A test covers the absent-object path.
8. `unknown` behaviour is unchanged: remote unreachable, timeout, and unreadable local HEAD all still produce `unknown` with a reason, and `unknown` is still never treated as current. Existing `unknown` tests remain unmodified and green.
9. The `stale` warning text remains actionable and still states that the command is read-only and does not fetch, pull, reset or sync.
10. Each new test is demonstrated failing on the parent commit, with the actual output pasted in the PR.
11. Full Release suite passes with the actual counts pasted.

## Verification

- New tests in `tests/IntentSystem.Cli.Tests/AutomationStalledWorkCommandTests.cs` covering criteria 1, 2, 4, 5, 6 and 7, driven through the existing `GitCommandRunnerFactory` seam.
- Criterion 3 requires cases B and C in the same class so the divergence is visible in one place.
- `git diff --check`.

## Related Links

- G755 (issue #1642, PR #1643, merged `9a30e95a`) — the sibling defect, deliberately scoped apart from this one; its packet recorded this probe's measured evidence for exactly this follow-up.

## Knowledge Maintenance

- ADR candidate: none. This corrects a predicate; it does not choose a new direction.
- Diagram candidate: none.
- Docs update: none — the probe's behaviour is not described in the doc mirrors, only in its own warning text, which criterion 9 covers.
- Closeout writeback expected: yes — node 03 records that the freshness predicate now measures containment rather than identity.

## Guide Reachability (G645)

`no_role_facing_surface: true`. This slice adds no command, flag, or output field. It corrects when an existing warning fires; the surface and its guide routes are unchanged.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1647

## Summary

- Change checkout freshness from SHA equality to local containment.
- Treat an absent local remote-tip object as stale without fetching.
- Keep `git ls-remote --symref origin HEAD`, current/stale/unknown statuses, current silence, warning wording, timeout, runner seam, and all stalled-work behavior unchanged.

## Verification

The new same-class G758 tests cover identical tips, ahead-containing tips, genuinely behind tips, detached-at-tip, an absent remote-tip object, and the exact read-only command sequence. The successful sequence is exactly:

1. `git rev-parse HEAD`
2. `git ls-remote --symref origin HEAD`
3. `git cat-file -e <remote-tip>^{commit}`
4. `git merge-base --is-ancestor <remote-tip> HEAD`

The absent-object path stops after step 3 and reports stale. No fetch, pull, reset, sync, or other mutation is issued.

Parent failure evidence at `ec261ec4c16454d122a3baec0d48393a4245f513` with the six new tests: `Failed: 6, Passed: 0, Skipped: 0, Total: 6`. The parent reported the ahead-containing case as stale and issued only the first two Git calls; all six new tests therefore failed their expected current/sequence assertions.

- G758 focused: `Failed 0, Passed 6, Skipped 0, Total 6`
- Existing G727 freshness/unknown tests: `Failed 0, Passed 3, Skipped 0, Total 3`
- Full `AutomationStalledWorkCommandTests`: `Failed 0, Passed 125, Skipped 0, Total 125`
- Full Release suite: `Failed 0, Passed 5363, Skipped 1, Total 5364`
- `git diff --check`: clean

## Coordination boundary

The supplied host authority is execution-unit:G758 held by implementation/team intent-cli-dev, claimed at `2026-08-29T18:28:41.844237Z`, base `f632f9da13496061a31e05addeb593049edefbbb`, verified on host origin/main. Child next-action selected #1647 but returned `wait`; child issue-preflight returned `claim-unavailable` because fresh canonical Git evidence could not open `.git/FETCH_HEAD` (`Operation not permitted`). This known child/host contradiction is recorded; no child execution-unit claim was created or mutated.

### DIFF
--- a/src/IntentSystem.Cli/Commands/CheckoutFreshnessProbe.cs
+++ b/src/IntentSystem.Cli/Commands/CheckoutFreshnessProbe.cs
@@ -101,7 +101,85 @@ internal static class CheckoutFreshnessProbe
                 "the origin response did not identify a default branch and HEAD");
         }
 
-        if (string.Equals(localHead, remoteHead, StringComparison.OrdinalIgnoreCase))
+        return DetermineContainment(repoRoot, runner, localHead, defaultBranch, remoteHead);
+    }
+
+    private static CheckoutFreshnessObservation DetermineContainment(
+        string repoRoot,
+        IGitRemoteCommandRunner runner,
+        string localHead,
+        string defaultBranch,
+        string remoteHead)
+    {
+        GitRemoteCommandResult remoteTipObject;
+        try
+        {
+            // A remote tip that is not already in the local object store cannot
+            // be reachable from HEAD. Treat that absence as a real stale
+            // result without fetching or otherwise mutating the checkout.
+            remoteTipObject = runner.Run(
+                repoRoot,
+                ["cat-file", "-e", $"{remoteHead}^{{commit}}"]);
+        }
+        catch (Exception exception) when (IsProbeFailure(exception))
+        {
+            return UnknownObservation(
+                localHead,
+                defaultBranch,
+                remoteHead,
+                $"the origin default branch tip could not be checked locally ({exception.Message})");
+        }
+
+        if (remoteTipObject.TimedOut)
+        {
+            return UnknownObservation(
+                localHead,
+                defaultBranch,
+                remoteHead,
+                FirstNonEmpty(
+                    remoteTipObject.StdErr,
+                    $"git cat-file -e {remoteHead}^{{commit}} timed out after {DefaultTimeout.TotalSeconds:0.###} seconds"));
+        }
+
+        if (remoteTipObject.ExitCode != 0)
+        {
+            return new CheckoutFreshnessObservation
+            {
+                Status = Stale,
+                LocalHead = localHead,
+                DefaultBranch = defaultBranch,
+                RemoteHead = remoteHead,
+            };
+        }
+
+        GitRemoteCommandResult ancestry;
+        try
+        {
+            ancestry = runner.Run(
+                repoRoot,
+                ["merge-base", "--is-ancestor", remoteHead, "HEAD"]);
+        }
+        catch (Exception exception) when (IsProbeFailure(exception))
+        {
+            return UnknownObservation(
+                localHead,
+                defaultBranch,
+                remoteHead,
+                $"the origin default branch tip could not be compared locally ({exception.Message})");
+        }
+
+        if (ancestry.TimedOut)
+        {
+            return UnknownObservation(
+                localHead,
+                defaultBranch,
+                remoteHead,
+                FirstNonEmpty(
+                    ancestry.StdErr,
+                    $"git merge-base --is-ancestor {remoteHead} HEAD timed out after {DefaultTimeout.TotalSeconds:0.###} seconds"));
+        }
+
+        if (ancestry.ExitCode == 0)
         {
             return new CheckoutFreshnessObservation
             {
@@ -112,6 +190,17 @@ internal static class CheckoutFreshnessProbe
             };
         }
 
+        if (ancestry.ExitCode != 1)
+        {
+            return UnknownObservation(
+                localHead,
+                defaultBranch,
+                remoteHead,
+                FirstNonEmpty(
+                    ancestry.StdErr,
+                    "the origin default branch tip could not be compared locally"));
+        }
+
         return new CheckoutFreshnessObservation
         {
             Status = Stale,
@@ -120,7 +209,6 @@ internal static class CheckoutFreshnessProbe
             RemoteHead = remoteHead,
         };
     }
-
     private static CheckoutFreshnessObservation UnknownObservation(
         string? localHead,
         string? defaultBranch,
--- a/tests/IntentSystem.Cli.Tests/AutomationStalledWorkCommandTests.cs
+++ b/tests/IntentSystem.Cli.Tests/AutomationStalledWorkCommandTests.cs
@@ -337,6 +337,119 @@ public sealed class AutomationStalledWorkCommandTests : IDisposable
     }
 
     [Fact]
+    public void Capture_IdenticalRemoteTipRemainsCurrentUsingLocalContainment_G758()
+    {
+        using var fixture = new FreshnessCheckoutFixture();
+        const string localHead = "identical-tip";
+        const string remoteHead = "identical-tip";
+        var runner = CreateContainmentRunner(localHead, remoteHead);
+
+        var observation = CheckoutFreshnessProbe.Capture(fixture.CheckoutPath, runner);
+
+        Assert.NotNull(observation);
+        Assert.Equal(CheckoutFreshnessProbe.Current, observation!.Status);
+        Assert.Equal(string.Empty, observation.Warning);
+        AssertContainmentGitSequence(runner, localHead, remoteHead);
+    }
+
+    [Fact]
+    public void Analyze_AheadContainingRemoteTipRemainsCurrentAndSilent_G758()
+    {
+        using var fixture = new FreshnessCheckoutFixture();
+        const string localHead = "ahead-head";
+        const string remoteHead = "base-head";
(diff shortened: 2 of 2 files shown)


tomohisa (commented):
APPROVED — G758 semantic review at `272d64ebb43706d348d5523f30a6c7e26c2c198d`.

The change is confined to the intended production probe and its stalled-work tests.  It implements the required containment semantics with the exact read-only runner sequence: `rev-parse HEAD`, `ls-remote --symref origin HEAD`, `cat-file -e <remote>^{commit}`, then `merge-base --is-ancestor <remote> HEAD`.  The probe performs no fetch, pull, reset, or sync.

I verified the discriminating cases in the added tests:

* equal tip is current and silent;
* an ahead checkout that contains the remote tip is current and silent;
* a genuinely behind checkout is stale and retains the actionable, explicitly read-only warning;
* a detached checkout at the remote tip is current and silent; and
* a remote object absent locally stops after `cat-file`, is stale, and does not run the ancestry check.

The B/C pair uses the same containment mechanism, so it distinguishes “ahead but contains remote” from truly behind rather than treating all non-equal revisions alike. Existing unknown-freshness tests are unmodified, preserving unknown rather than inferring currency. The test also asserts the four-command sequence and prohibits mutating Git commands.

The parent failure evidence and the packet’s node 03 audit/read-only constraints are consistent with this repair. `Closes #1647` is present. CI is green and the head remained unchanged during review.

Validation run locally from this exact head:

* `dotnet test ... --filter FullyQualifiedName~AutomationStalledWorkCommandTests` — 125 passed, 0 failed.
* Full Release `IntentSystem.Cli.Tests` — 5,363 passed, 0 failed, 1 skipped.

PR #1653, opened 2026-08-29
fix: resolve task-envelope report root from recipient cwd

### LINKED ISSUE #1651: G760: the envelope's report command pins the host root absolutely and the recipient's own root as ".", so a report is lost whenever the recipient is standing somewhere else
# G760: the envelope's report command pins the host root absolutely and the recipient's own root as ".", so a report is lost whenever the recipient is standing somewhere else

## Goal

Make the `canonical-report-command` embedded in a task envelope express the recipient's report root the way it already expresses the host routing root: as a resolved absolute path, not as a working-directory-relative `.`. The intent behind `--report-root` — a sender-local outbox in the role's own writable root, with the host root never needing a write grant — is correct and unchanged. Only its encoding changes.

## Why This Slice Exists Now

The generated command is the recipient's **required final step**, and the envelope says so: "Run canonical-report-command after all other work; never hand-write a transport invocation." A recipient that follows that instruction exactly can still lose its report, because the command's correctness silently depends on where the recipient happens to be standing when it runs it.

That is not hypothetical. Three review reports were lost this cycle and are absent from **both** the seat's outbox and the host outbox. Nothing surfaced it: the GitHub verdict and the transition were applied, so the loop looked healthy, and only a direct inspection of the outbox files showed the reports were never recorded. The durable record is missing three review verdicts that were actually produced.

## Current Observed State

Measured with installed `intent-cli 0.27.0-f43fbd1-G753`. The review seat supplied the exact commands, working directories and full error text from its own history; all four commands were **identical in form**, and only the working directory differed:

| Attempt | `--report-root` | Working directory | Outcome |
| --- | --- | --- | --- |
| G755-review-1643-001 | `.` | `…/MyIntentHost` | **Failed** |
| G755-rereview-1643-001 | `.` | `…/MyIntentHost` | **Failed** |
| G756-review-1646-001 | `.` | `…/MyIntentHost` | **Failed** |
| G757-review-1648-001 | `.` | `…/IntentSystemReview` | Succeeded, sender-local |

Full failure text:

```
cause: report-outbox-write-failed
Could not persist report task '<task-id>' before transport: Access to
'/Users/tomohisa/dev/GitHub/MyIntentHost/.intent-cli/notify/intent-cli/intent-cli-dev/report-outbox.jsonl'
is denied. No transport was attempted.
```

`.` resolved to the host root, so the outbox retargeted to the host, the sandbox denied the write, and — correctly — no transport was attempted. The report was lost before it could be sent.

The command is built at `NotifyCommand.cs:2423-2427`:

```csharp
+ $"--artifact <artifact> --summary <one-line-summary> --routing-root {ShellQuote(options.RoutingRoot!)} --report-root . "
```

The routing root is resolved and shell-quoted; the report root is a bare literal.

**The correct pattern already exists in the same file.** `BuildReconciliationCommand` at `:2429-2437` emits `--report-root {ShellQuote(options.ReportRoot)}` — absolute and quoted.

**The recipient's work root is already recorded.** The team topology carries an absolute `cwd` per herdr-resident role, e.g. `review: /Users/tomohisa/dev/GitHub/IntentSystemReview`. Nothing needs to be discovered or invented.

A hypothesis was tested and **disproved** before this packet was written: that some envelope generation omitted `--report-root`. All four envelopes were inspected and carry an identical `--report-root .`, so envelope-to-envelope variation is not the cause.

- The **intent** of `--report-root` is correct and stays: the role's own writable root owns the sender-local outbox, the host routing root is read/transport authority, and an isolated child checkout is never given a host-root write grant. This slice does not change that contract; it makes the generated command express it unambiguously.
- `--routing-root` continues to be emitted as a resolved, shell-quoted absolute path.
- The team topology records an absolute `cwd` for each herdr-resident role and is the source for the recipient's work root.
- The report-outbox storage model, `notify reconcile`, and the `report-outbox-write-failed` failure behaviour (persist before transport; on failure attempt no transport) are unchanged and correct.
- An external resident may have no recorded `cwd`; that case is handled explicitly below rather than by falling back to `.`.

## Target Repo / Path / Part

- Target paths: `src/IntentSystem.Cli/Commands/NotifyCommand.cs`, `tests/IntentSystem.Cli.Tests/NotifyCommandG578Tests.cs`, `docs/en/12-agent-message-orchestration.md`, `docs/ja/12-agent-message-orchestration.md`

Target part: `BuildReportCommand` and the envelope `reporting-contract` block it feeds, plus the documentation describing the embedded report command.

## In Scope

- `BuildReportCommand` emits the recipient's **resolved absolute** work root as `--report-root`, shell-quoted, matching how `--routing-root` is already emitted and how `BuildReconciliationCommand` already emits its own report root.
- The work root is taken from the recipient role's recorded topology `cwd`.
- When the recipient has **no** recorded work root, the command must not emit a silently cwd-dependent value. It emits an explicit placeholder the recipient must substitute — the same shape the usage text already uses for unknown values — so the dependency is visible rather than hidden.
- Both documentation mirrors describe the emitted command with an absolute report root.

## Out Of Scope

- The meaning or contract of `--report-root`, the outbox storage model, and which root owns the sender-local outbox.
- `notify reconcile`, `notify collect`, and the `report-outbox-write-failed` behaviour.
- Any change to `--routing-root`.
- Any change to the envelope's other fields, the result nonce, or the completion marker.
- Granting any seat write access to the host root. The fix is to stop retargeting the outbox, not to permit the retargeted write.
- Recovering the three already-lost reports. They are gone from the durable record; this slice stops the next one.
- The `Read task envelope` pointer wording. Separate unit, already raised.

The child PR changes the `canonical-report-command` that intent-cli embeds in a task envelope so that `--report-root` carries the recipient role's resolved absolute work root, shell-quoted, instead of the literal `.`. The recipient's work root comes from its recorded topology `cwd`; when no work root is recorded, the command emits an explicit placeholder rather than a working-directory-relative value, so the dependency is never hidden. `--routing-root` and every other envelope field are unchanged, as is the meaning of `--report-root` and the report-outbox storage model. The PR must add tests proving the emitted command contains an absolute report root for a recipient with a recorded work root, that it contains no bare `--report-root .`, that the placeholder form appears when no work root is recorded, and that the emitted command is identical to today's apart from the report-root argument.

## Acceptance Criteria

1. For a recipient with a recorded work root, the emitted `canonical-report-command` contains `--report-root` followed by that **absolute** path, shell-quoted. A test asserts the exact emitted string.
2. The emitted command contains no bare `--report-root .`. A test asserts its absence explicitly, so a partial fix cannot pass.
3. For a recipient with **no** recorded work root, the command emits an explicit placeholder rather than `.` or an inferred path. A test asserts it.
4. Apart from the `--report-root` argument, the emitted command is byte-identical to the parent commit's for the same inputs. A test pins this, so the fix cannot drift into the rest of the command.
5. `--routing-root` remains a resolved, shell-quoted absolute path. A test asserts it is unchanged.
6. A work root containing a single quote is shell-quoted correctly, matching the existing `ShellQuote` behaviour used for `--routing-root`.
7. The rest of the envelope — inputs, expected artifacts, result prefix, nonce, completion marker, required-final-step line — is byte-unchanged. A test asserts the full envelope for a fixed input differs from the parent only in the report-root argument.
8. Each new test is demonstrated failing on the parent commit, with the actual output pasted in the PR.
9. Both doc mirrors describe the emitted command with an absolute report root.
10. Full Release suite passes with the actual counts pasted.

## Verification

- New and updated tests in `tests/IntentSystem.Cli.Tests/NotifyCommandG578Tests.cs`.
- Criterion 4 and 7 are best satisfied by generating the whole envelope for a fixed input and diffing against the parent output; state in the PR that the only difference is the report-root argument.
- `git diff --check`.

## Related Links

- G759 (issue #1649) — the sibling defect in the same envelope: the delivery pointer says read rather than execute. Different line, same envelope, separate unit.
- G755 (issue #1642, merged) and G756 (issue #1644, merged) — the same failure family: a path whose meaning depends on where the reader is standing.
- `intents/intent-cli/intent-tree/means/08-agent-message-orchestration.md` — node this slice supports; its design-time record carries the measured table above.

## Knowledge Maintenance

- ADR candidate: none. The contract is already decided; this makes the generated command express it.
- Diagram candidate: none.
- Docs update: `docs/en/12-agent-message-orchestration.md`, `docs/ja/12-agent-message-orchestration.md`
- Closeout writeback expected: yes — node 08 records that three review reports were lost to this and that the durable log is permanently missing them.

## Guide Reachability (G645)

`no_role_facing_surface: false`. The emitted command is a role-facing instruction every file-backed recipient runs as its required final step.

- guide_surface: `guide workflow task orchestrator-thread`; role: `review`; target_surface: the `canonical-report-command` embedded in the envelope's reporting-contract.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1651

## Summary
Task-envelope canonical-report-command now resolves --report-root from the recipient role topology cwd, normalizes it to an absolute path, and shell-quotes it. When the recipient has no recorded cwd, the envelope carries the explicit <role-work-root> placeholder. The remaining command and full envelope stay byte-compatible with the parent; routing-root, nonce, completion marker, notify reconcile/collect/outbox behavior, and host permissions are unchanged.

## Evidence
- Parent baseline 145c5a43c031353a5e5ad4d7ea9eb3fb7365304c with the three G760 regressions applied: 26 existing tests passed and 3 new tests failed. The failures were the missing explicit placeholder, the absolute shell-quoted recipient cwd, and the full-envelope report-root-only substitution; parent output still emitted --report-root ..
- Change: NotifyCommandG578Tests 29 passed, 0 failed, 0 skipped.
- Existing G719 compatibility: 4 passed, 0 failed, 0 skipped.
- Orchestration docs: 24 passed, 0 failed, 0 skipped.
- G613: 6 passed, 0 failed, 0 skipped.
- Packaged invocation smoke: 11 passed, 0 failed, 0 skipped.
- Full Release on current main c6e6922e8ca89520465adfa8f69375eefd5d4fa6: 5367 passed, 0 failed, 1 skipped, 5368 total.
- git diff --check: clean.
- GitHub CI run 33275004022: Build and test source contract passed; npm package dry-run passed.

The focused tests cover a topology cwd containing a single quote, no cwd placeholder, and exact UTF-8 byte equality against the checked-in parent envelope after replacing only the report-root segment. The G719 fixture adaptation records its existing seat root as the topology cwd so its shell-executed generated report continues to exercise the same sender-local behavior; no product outbox or transport semantics changed.

## Worker boundary
Child worker next-action returned wait for issue 1651 because local execution-unit G760 was reported holder none/unheld, and issue-preflight returned actionable=true ready-to-implement with the same local claim refusal. Supplied authoritative host evidence is consumed instead: origin/main claim d1ceba84afe2ca6acda45a2ea08a9ef29579bd699089b07e41768b1e7071c325.json records execution-unit:G760 for implementation/team intent-cli-dev at base 34a2540357b4dc0cddc594c7e63027f7d12ba357. No child execution-unit claim or host metadata was accessed or mutated.

### DIFF
--- a/docs/en/12-agent-message-orchestration.md
+++ b/docs/en/12-agent-message-orchestration.md
@@ -484,7 +484,7 @@ intent-cli notify report --domain <domain> --team <team> --from <receiver-role> 
   --to <orchestrator-role> --task-id <task-id> \
   --status completed|blocked|question --artifact <artifact> \
   --summary <one-line-summary> --routing-root <host-routing-root> \
-  --report-root . --write --format json
+  --report-root <role-work-root> --write --format json
 
 # Route a design decision to the existing events.jsonl boundary.
 intent-cli notify escalate --domain <domain> --team <team> --from <sender-role> \
@@ -637,7 +637,7 @@ unrecognised identifier would silence unsolicited reports and answers to
 escalations—the messages whose information the recipient did not know to
 request.
 
-**Sender-local report routing and executability/actionability diagnostics (G719/G731 — preview-through-1.x).** An implementation seat receives exactly one bounded `--add-dir <role-work-root>` and is not granted `--add-dir <host-routing-root>` or host-root write access. Its canonical report carries `--routing-root <host-routing-root> --report-root .`: the host root remains read/transport authority and the role root is the writable sender-local outbox. A report on a seat that genuinely lacks the host routing root must persist its local outbox, identify `report_storage_mode: sender-local-role-work-root`, and identify `host_state_sync: deferred-to-orchestration`; orchestration reconciles host pending/continuation state with `intent-cli notify reconcile --domain <d> --team <t> --task-id <id> --routing-root <host> --report-root <role-work-root> --write --format json`. It must not retry a host write, hand-write transport, or widen the seat.
+**Sender-local report routing and executability/actionability diagnostics (G719/G731/G760 — preview-through-1.x).** An implementation seat receives exactly one bounded `--add-dir <role-work-root>` and is not granted `--add-dir <host-routing-root>` or host-root write access. Its task-envelope canonical report carries `--routing-root <host-routing-root> --report-root <role-work-root>`: at generation time, `<role-work-root>` is replaced with the recipient's recorded absolute topology `cwd` and shell-quoted. If no cwd is recorded, the envelope carries the explicit `<role-work-root>` substitution placeholder; it never emits a bare `.` that depends on the receiver's current directory. The host root remains read/transport authority and the role root is the writable sender-local outbox. A report on a seat that genuinely lacks the host routing root must persist its local outbox, identify `report_storage_mode: sender-local-role-work-root`, and identify `host_state_sync: deferred-to-orchestration`; orchestration reconciles host pending/continuation state with `intent-cli notify reconcile --domain <d> --team <t> --task-id <id> --routing-root <host> --report-root <role-work-root> --write --format json`. It must not retry a host write, hand-write transport, or widen the seat.
 
 G731 makes the delivery decision capability/outcome-based across both
 `notify report` and `notify collect`: differing report and routing roots are
@@ -1396,7 +1396,7 @@ herdr agent start <logical-role> --kind copilot --pane <pane-id> -- --mode autop
 - **Role-derived roots.** Give every role one bounded `--add-dir <role-work-root>`
   for its checkout or worktree. A reviewer may additionally need `--add-dir
   <host-routing-root>` for an external-reader surface; an implementation seat
-  uses `--routing-root <host-routing-root> --report-root .` and must not receive
+  uses `--routing-root <host-routing-root> --report-root <role-work-root>` resolved from the recipient's absolute topology `cwd` (or the explicit `<role-work-root>` placeholder when absent) and must not receive
   host-root write access. Do not add unrelated developer-machine roots. Before
   delegation, the orchestrator compares workspace prerequisites with this
   recorded write envelope and prepares anything outside it under orchestrator
@@ -1445,7 +1445,7 @@ herdr agent start <logical-role> --kind codex --pane <pane-id> -- --sandbox work
 - **Role-derived roots.** Use one bounded `--add-dir <role-work-root>` for the
   role checkout/worktree. The implementation seat does not receive
   `--add-dir <host-routing-root>` or MyIntentHost write access: its canonical
-  report uses `--routing-root <host-routing-root> --report-root .`, with the
+  report uses `--routing-root <host-routing-root> --report-root <role-work-root>` resolved from the recipient's absolute topology `cwd` (or the explicit `<role-work-root>` placeholder when absent), with the
   host root as read/transport authority and the role root as the writable
   sender-local handoff. Add a host root only when another role's external
   reader explicitly requires it. The role-work-root is an ordinary-file
@@ -1496,7 +1496,7 @@ herdr agent start <logical-role> --kind codex --pane <pane-id> -- --sandbox work
 three additional facts: an expected action inside the recorded roots succeeds;
 the role reaches its canonical reporting surface; and a deliberately
 out-of-scope action is denied. For implementation, prove the report with
-`--routing-root <host-routing-root> --report-root .` and record the local outbox
+`--routing-root <host-routing-root> --report-root <role-work-root>` from the recipient's absolute topology `cwd` (or the explicit `<role-work-root>` placeholder when absent) and record the local outbox
 plus deferred host-state reconciliation without granting host-root write access.
 Capture that denial for review. A live pane or a successful allowed action alone
 is **not** READY. If a denial probe unexpectedly succeeds, first
--- a/docs/ja/12-agent-message-orchestration.md
+++ b/docs/ja/12-agent-message-orchestration.md
@@ -434,7 +434,7 @@ intent-cli notify report --domain <domain> --team <team> --from <receiver-role> 
   --to <orchestrator-role> --task-id <task-id> \
   --status completed|blocked|question --artifact <artifact> \
   --summary <one-line-summary> --routing-root <host-routing-root> \
-  --report-root . --write --format json
+  --report-root <role-work-root> --write --format json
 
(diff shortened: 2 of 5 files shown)


tomohisa (commented):
APPROVED — G760 semantic review at `f924cce8d358fe189d0a2ba3a469f75e9b3a15c6`.

`BuildReportCommand` now resolves the recipient role through the recorded topology and emits `--report-root` as `ShellQuote(Path.GetFullPath(cwd))`. The established `--routing-root` expression is unchanged. A missing recipient cwd produces the deliberate `<role-work-root>` placeholder, never an inferred/current-directory `.`.

I verified AC1–AC7 in the added G760 regression surface:

* a topology cwd containing a single quote is absolute and uses the same shell-quote form as routing root;
* the generated command explicitly rejects bare `--report-root .` and the no-cwd branch emits the documented placeholder;
* the exact command is the parent command with only its report-root argument substituted; and
* the complete fixed-input envelope is UTF-8 byte-identical to the parent envelope after precisely that one substitution, retaining inputs, expected artifacts, nonce, completion marker, required-final-step, and routing root.

I independently ran the three new regressions on prepared parent `145c5a43`; all failed as required (0 passed / 3 failed): missing placeholder, bare-dot report root instead of absolute shell-quoted cwd, and full-envelope mismatch at the report-root segment. The parent output was not accepted on assertion alone.

The five-file diff is confined to `BuildReportCommand`, its focused tests/G719 compatibility fixture adaptation, and the EN/JA mirrors. It does not alter outbox ownership/storage, persist-before-transport failure behavior, reconcile/collect, nonce/completion marker, G759 file pointer, topology semantics, or host permissions. `Closes #1651` is present.

Validation is necessary but not sufficient: focused surface including packaged smoke, G719 compatibility, docs, and G613 passed 74/74; full Release passed 5,367 with 0 failures and 1 skipped. Both CI checks are green and the head remained unchanged through review.

`design_alignment_checked=true` — checked G760 packet, implementation notes, review context, GitHub body, intent node `08-agent-message-orchestration.md`, and both documentation mirrors. This is consistent with node 08's sender-local outbox / host read-transport boundary: correct the generated path encoding without granting host-root write access.

PR #1660, opened 2026-08-30
G761: add guarded topology residence transition

### LINKED ISSUE #1655: G761: a recorded role's residence can never be changed, so a team that moves its design seat has no canonical move to make
# G761: a recorded role's residence can never be changed, so a team that moves its design seat has no canonical move to make

## Goal

Give `session-layer topology` a canonical, explicit way to change a recorded role's `resident` — `external` ⇄ `herdr` — using the compare-and-swap-plus-confirmation shape the same command family already uses for `kind` and `delivery_method`. The existing fail-closed refusal on `record` stays exactly as it is; this adds the deliberate transition it currently has no way to express.

## Why This Slice Exists Now

A team's design seat legitimately changes residence. `guide bootstrap` makes it a first-class, human-answered decision: "Ask the human which agent kind this application conversation uses and whether that kind has an inbound application monitor. Never infer or default either answer. With a monitor, design may be the recorded external reader; without one, design must be a recorded herdr seat at the routing-root cwd."

So the answer to that question determines `resident`, and the answer can change — a team adopts a front-end with a monitor, or drops one, or the deployment rule's premise changes. When it does, there is no command that can record the new answer. `record` refuses, correctly, and nothing else can do it.

The refusal message names the real boundary: *"Refusing to **silently** repair or replace it."* What is forbidden is a silent replacement. An explicit, current-value-checked, confirmed replacement is a different act, and the command family already knows how to express it — `update-kind` and `update-field` both take `--current-<x> --new-<x> --confirm-<x>`. Residence is the one recorded property with no such surface.

Observed consequence: a design seat in another domain hit this while migrating, read the fail-closed refusal as a missing feature, and considered hand-editing the topology JSON to get past it. It stopped and asked instead — the right call — but the pressure to bypass a guard is exactly what an absent transition creates. A guard with no sanctioned way through teaches people to go around it.

## Current Observed State

Verified directly against installed `intent-cli 0.27.1-5d553b7-G756`, from the command usage text:

| Surface | What it can change |
| --- | --- |
| `topology record` | New records only. Two forms, `--resident herdr` and `--resident external`. **No `--force`, `--replace`, or `--confirm-*` flag.** |
| `topology move` | `--workspace-id` and `--pane-map` only. |
| `topology update-kind` | `kind` only, via `--current-kind --new-kind --confirm-update-kind`. |
| `topology update-field` | **`--field delivery_method` only** — the flag takes exactly that one value. |
| `topology retire-legacy`, `record-profile`, `record-host-state` | Different purposes; none changes residence. |

Recording a role that already exists with a different shape returns `conflict: true` with `"Role '<role>' already has a conflicting recorded shape. Refusing to silently repair or replace it"`.

So there is no path from a recorded `external` role to a recorded `herdr` role, or back.

**Honest scope note:** the migration that surfaced this did not actually need the transition — that team's design was already `external`, and moving to another external front-end required no change at all. The refusal was protecting a correct record. So this slice fixes a real missing surface, not an active outage, and it should be sized accordingly.

- The fail-closed refusal on `record` is correct and stays. This slice does not weaken it, add a bypass flag to it, or make it repair anything silently.
- `guide bootstrap`'s rule stands: residence follows a human answer about the inbound application monitor, and is never inferred or defaulted. This slice records an answer; it does not decide one.
- `update-kind` and `update-field` define the house pattern for a guarded mutation: assert the current value, state the new one, and require an explicit confirmation flag. Follow it.
- Topology records are machine-local and directory-local-ignored; this is not shared durable state.
- Delivery resolution itself is unchanged — a `herdr` role resolves to a pane, an `external` role to its reader, exactly as now.

## Target Repo / Path / Part

- Target paths: `src/IntentSystem.Cli/Commands/SessionLayerTopologyCommand.cs`, `tests/IntentSystem.Cli.Tests/SessionLayerTopologyCommandTests.cs`, `docs/en/12-agent-message-orchestration.md`, `docs/ja/12-agent-message-orchestration.md`

Target part: the topology mutation surfaces — specifically the addition of a residence transition alongside `update-kind` and `update-field`.

## In Scope

- A canonical residence transition on `session-layer topology`, following the existing guarded-mutation shape: assert the current residence, state the new one, require an explicit confirmation flag, support `--dry-run`.
- Both directions: `external` → `herdr` and `herdr` → `external`.
- The transition takes the fields the destination residence requires and rejects the ones it forbids, matching `record`'s own two forms: `herdr` needs workspace/pane/cwd and refuses `--reader`/`--frontend`; `external` needs a reader and refuses pane fields.
- Fields that do not belong to the new residence are removed rather than left stranded on the record.
- Both documentation mirrors describe the transition.

## Out Of Scope

- Any weakening of `record`'s fail-closed refusal, and any `--force`/`--replace` on `record`.
- Any change to delivery resolution, to `move`, `update-kind`, `update-field`, `retire-legacy`, `record-profile`, or `record-host-state`.
- Deciding residence, inferring it, or defaulting it — that answer is the human's, per `guide bootstrap`.
- Migrating or repairing existing records automatically.
- Any change to the `guide design-thread` deployment rule itself. Whether a canonical blocking receive should count as an inbound application monitor is a separate question and a separate unit.

The child PR adds a canonical way to change a recorded role's residence between `external` and `herdr` in `intent-cli session-layer topology`, following the same guarded-mutation shape the command family already uses for `kind` and `delivery_method`: the caller asserts the current residence, states the new one, and passes an explicit confirmation flag, with `--dry-run` supported. The transition must require exactly the fields the destination residence requires and reject those it forbids, mirroring `record`'s two forms, and must not leave fields from the old residence stranded on the record. The existing fail-closed refusal on `record` must be unchanged, with no bypass flag added to it, and delivery resolution must be untouched. Tests must cover both directions, a wrong asserted current value, a missing confirmation flag, forbidden fields for each destination, and that the old residence's fields are gone afterwards.

## Acceptance Criteria

1. A residence transition exists on `session-layer topology` and changes a recorded role from `external` to `herdr`, and from `herdr` to `external`. Tests cover both directions.
2. It requires the caller to assert the current residence, and refuses when that assertion does not match the record. A test asserts the refusal and that the record is unchanged.
3. It requires an explicit confirmation flag, in the same shape as `--confirm-update-kind` / `--confirm-update-field`. A test asserts that omitting it refuses and changes nothing.
4. `--dry-run` reports the intended change and writes nothing. A test asserts the record is byte-unchanged after a dry run.
5. Destination-required fields are enforced: `herdr` requires workspace, pane and cwd; `external` requires a reader. A test covers a missing required field in each direction.
6. Destination-forbidden fields are rejected, matching `record`'s own rules: `herdr` refuses `--reader`/`--frontend`; `external` refuses pane fields. A test covers each.
7. After a transition, fields belonging to the old residence are **absent** from the record — no stranded `reader` on a herdr role, no stranded `pane_id` on an external one. A test asserts absence, not merely that the new fields are present.
8. `record`'s fail-closed refusal is unchanged and gains no bypass. Its existing tests remain unmodified and green, and a test asserts `record` still refuses a conflicting shape after this slice.
9. Delivery resolution is unchanged: a `herdr` role still resolves to its pane and an `external` role to its reader. Existing delivery tests remain unmodified and green.
10. Each new test is demonstrated failing or absent on the parent commit, with the actual output pasted in the PR.
11. Both doc mirrors describe the transition and state that residence follows the human answer from `guide bootstrap`, never an inference.
12. Full Release suite passes with the actual counts pasted.

## Verification

- New tests in `tests/IntentSystem.Cli.Tests/SessionLayerTopologyCommandTests.cs` covering criteria 1-9.
- Criterion 7 is the one most easily faked by asserting only the new shape; require an explicit absence assertion for the old residence's fields.
- `git diff --check`.

## Related Links

- `intents/intent-cli/intent-tree/means/08-agent-message-orchestration.md` — node this slice supports; its 2026-08-29 追記 records the verified surface inventory above and the incident that surfaced it.
- G756 (issue #1644, merged) — the sibling defect in the same file: a recorded reader that was never checked against the effective one. Same theme, that a topology record can be wrong and unmovable.

## Knowledge Maintenance

- ADR candidate: none. The guarded-mutation pattern is already established by `update-kind` and `update-field`; this follows it rather than choosing anew.
- Diagram candidate: none.
- Docs update: `docs/en/12-agent-message-orchestration.md`, `docs/ja/12-agent-message-orchestration.md`
- Closeout writeback expected: yes — node 08 records that residence became transitionable and that the refusal stayed.

## Guide Reachability (G645)

`no_role_facing_surface: false`. This slice adds a role-facing command surface.

- guide_surface: `guide bootstrap`; role: `design`; target_surface: the residence transition, reachable from the step that asks the human about the inbound application monitor.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
## Summary

Closes #1655

- Add `session-layer topology update-residence` for explicit `external` <-> `herdr` transitions.
- Require the asserted current resident, a different new resident, explicit confirmation, and `--write` or `--dry-run --format json`.
- Validate destination fields like `topology record`, remove residence-exclusive fields from the old shape, and use a topology CAS lock plus a second digest comparison before writing.
- Keep `record` conflict refusal, delivery resolution, and all existing topology mutation surfaces unchanged.
- Document that `guide bootstrap`'s human answer controls residence in both EN/JA orchestration mirrors.

## Verification

Exact checkout head: `39ade693e191d87f1d8def113ca80e7aad0404ec`

Changed paths:

- `src/IntentSystem.Cli/Commands/SessionLayerTopologyCommand.cs`
- `tests/IntentSystem.Cli.Tests/SessionLayerTopologyCommandTests.cs`
- `docs/en/12-agent-message-orchestration.md`
- `docs/ja/12-agent-message-orchestration.md`

Parent evidence was collected in an isolated worktree at parent `ff11a355377fe2b1698cce1e14f39d8c79c20bd5` with an ephemeral `G761ParentAbsenceTests` probe. The parent had no `update-residence` route:

```text
Failed!  - Failed:     1, Passed:     0, Skipped:     0, Total:     1
Error Message:
Assert.Equal() Failure: Values differ
Expected: 0
Actual:   1
```

Post-change evidence (SDK 10.0.100, Release):

- New/affected `SessionLayerTopologyCommandTests`: `Failed: 0, Passed: 17, Skipped: 0, Total: 17`.
- Topology plus orchestration documentation tests: `Failed: 0, Passed: 41, Skipped: 0, Total: 41`.
- Adjacent topology, delivery-resolution, notify, and G613 tests: `Failed: 0, Passed: 87, Skipped: 0, Total: 87`.
- Full Release suite: `Failed: 0, Passed: 5381, Skipped: 1, Total: 5382`.
- `git diff --check`: passed.

The focused tests cover both directions, wrong current value, missing confirmation, required and forbidden destination fields, dry-run byte identity, absence of old-residence fields, and unchanged `record` conflict refusal. The adjacent delivery suites remain green; no delivery-resolution source was changed.

## Coordination boundary

The supplied host claim for `execution-unit:G761` is authoritative. The child worker next-action/preflight contradiction was recorded: next-action waits on issue `#1655`, and issue-preflight is `actionable=false`, `classification=claim-unavailable`, `claim_status=canonical-unavailable` because child `.git/FETCH_HEAD` cannot be opened (`Operation not permitted`). No child product-repo execution-unit claim was created, verified, released, or mutated. The canonical GitHub issue claim was performed through the worker command as required.

### DIFF
--- a/docs/en/12-agent-message-orchestration.md
+++ b/docs/en/12-agent-message-orchestration.md
@@ -1795,7 +1795,7 @@ form a recovery dead end.
 Agent kind is whatever herdr can start: Claude, Codex, Copilot, Cursor, OpenCode,
 and others are examples, not a supported-set restriction. Logical role defaults
 are `implementation`, `review`, `interview`, and `clarify`; existing explicit
-role mappings, including legacy product-named mappings, remain valid. The three
+role mappings, including legacy product-named mappings, remain valid. The four
 update/retire mutation commands emit JSON and support only `--format json`. For `update-kind`,
 an explicit `--dry-run` takes precedence over `--write` in either flag order and
 never writes.
@@ -1840,6 +1840,32 @@ The registry initially permits only `delivery_method`, so an unknown or dotted
 name is refused rather than becoming an arbitrary JSON-path editor. The command
 changes only that field. It does not relax `record`: re-recording a different
 shape still refuses its conflict and has no force flag.
+
+#### Guarded residence transition (G761)
+
+When the human answer from `guide bootstrap` changes whether a role has an
+inbound application monitor, use the explicit residence transition instead of
+editing the topology file:
+
+```text
+intent-cli session-layer topology update-residence --domain <domain> --team <team> --role <role> --current-resident <herdr|external> --new-resident <herdr|external> [destination fields] --confirm-update-residence --write --format json
+```
+
+The human answer from `guide bootstrap` controls residence. intent-cli never
+infers it or supplies a default. The command asserts the recorded current
+resident, requires the new resident and explicit confirmation, and uses the
+same guarded compare-and-swap shape as the other topology writers. Use
+`--dry-run` for the guarded preview; it does not write the record.
+
+The destination has the same shape as `topology record`: `herdr` requires
+`--workspace-id`, `--pane-id`, and `--cwd`, and rejects `--reader` and
+`--frontend`; `external` requires `--reader` and rejects workspace, pane, cwd,
+kind, and delivery-method fields. Fields owned only by the old residence are
+removed, so a transitioned role has no stranded reader or pane fields. The
+existing `record` conflict refusal remains fail-closed and has no
+`--force`/`--replace` bypass. Delivery resolution is unchanged: herdr still
+resolves to its recorded pane and external still resolves to its recorded
+reader.
 
 #### Visible, generated mode markers
 
--- a/docs/ja/12-agent-message-orchestration.md
+++ b/docs/ja/12-agent-message-orchestration.md
@@ -1542,7 +1542,7 @@ refusal は動作する whole-team move command を示すため、二つの surf
 agent kind は herdr が起動できる任意の kind です。Claude、Codex、Copilot、Cursor、OpenCode などは
 例であり、supported-set の制約ではありません。logical role の既定値は `implementation`、`review`、
 `interview`、`clarify` です。legacy の product 名を含め、既存の明示的な role mapping はそのまま
-有効です。3 つの update/retire mutation command は JSON を出力し、`--format json` だけをサポートします。
+有効です。4 つの update/retire mutation command は JSON を出力し、`--format json` だけをサポートします。
 `update-kind` では明示した `--dry-run` が flag の順序にかかわらず `--write` より優先され、決して
 書き込みません。
 
@@ -1575,6 +1575,27 @@ herdr query を行いません。mapping が存在するか herdr-only が必要
 unknown または dotted name は任意の JSON path を編集できないよう拒否されます。この command が
 変更するのはその field だけです。`record` の conflict refusal は緩和されず、異なる shape の re-record は
 引き続き拒否され、force flag もありません。
+
+#### Guarded residence transition (G761)
+
+`guide bootstrap` の human answer が role の inbound application monitor の有無を変える場合は、topology file を
+直接編集せず、明示的な residence transition を使います。
+
+```text
+intent-cli session-layer topology update-residence --domain <domain> --team <team> --role <role> --current-resident <herdr|external> --new-resident <herdr|external> [destination fields] --confirm-update-residence --write --format json
+```
+
+`guide bootstrap` の human answer が residence を決めます。intent-cli は residence を推測せず、default も
+補いません。command は記録済みの current resident を確認し、新しい resident と explicit confirmation を
+要求し、他の topology writer と同じ guarded compare-and-swap の形を使います。guarded preview には
+`--dry-run` を指定し、record は書き換えません。
+
+destination の形は `topology record` と同じです。`herdr` は `--workspace-id`、`--pane-id`、`--cwd` が必要で、
+`--reader` と `--frontend` を拒否します。`external` は `--reader` が必要で、workspace、pane、cwd、kind、
+delivery-method field を拒否します。旧 residence だけに属する field は削除するため、transition 後に reader や
+pane field が stranded になることはありません。既存の `record` conflict refusal は fail-closed のままで、
+`--force`/`--replace` の bypass はありません。delivery resolution も変わらず、herdr は記録済み pane、external は
+記録済み reader に解決されます。
 
 #### 可視な生成済み mode marker
 
--- a/src/IntentSystem.Cli/Commands/SessionLayerTopologyCommand.cs
+++ b/src/IntentSystem.Cli/Commands/SessionLayerTopologyCommand.cs
@@ -46,6 +46,13 @@ internal static class SessionLayerTopologyCommand
         "Usage: intent-cli session-layer topology update-field --domain <name> --team <name> --role <name> "
         + "--field delivery_method --current <value|absent> --new <value> --confirm-update-field "
         + "[--dry-run|--write] [--format json]";
+    private const string UpdateResidenceUsage =
+        "Usage: intent-cli session-layer topology update-residence --domain <name> --team <name> --role <name> "
+        + "--current-resident herdr|external --new-resident herdr|external "
+        + "[--workspace-id <id> --pane-id <id> --cwd <path> | --reader <routing-root-relative-path>] "
+        + "[--frontend <name>] [--kind <kind>] [--delivery-method inline|file-backed] "
+        + "[--model <text>] [--reasoning-effort <text>] --confirm-update-residence "
+        + "[--dry-run|--write] --format json";
     private const string RetireLegacyUsage =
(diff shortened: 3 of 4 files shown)


tomohisa (commented):
## APPROVED — G761 guarded topology residence transition

Reviewed at exact head `39ade693e191d87f1d8def113ca80e7aad0404ec` against `main` / issue #1655, the G761 packet, review context, and `ICL.M.AGENT_MESSAGE_ORCHESTRATION` in `intents/intent-cli/intent-tree/means/08-agent-message-orchestration.md`.

The packet contract is met: this is an explicit, current-value-checked, confirmed `external` ↔ `herdr` transition alongside `update-kind` / `update-field`, while `record` continues to refuse silent replacement.  `TryParseUpdateResidenceArguments` requires both distinct residences, `--confirm-update-residence`, JSON, and an explicit write or dry-run; it applies the same destination shapes as `record`.  `SessionLayerTopologyWriter.UpdateResidence` validates the existing record and asserted current residence, takes the topology CAS lock for writes, compares the snapshot digest immediately before atomic replacement, and removes the old shape's fields (`reader`/`frontend` for herdr; workspace/pane/cwd/kind/delivery method for external).

Criteria 1–7 are concretely exercised in the new `SessionLayerTopologyCommandTests`: both directions, asserted-current refusal with byte-preserved record, missing confirmation, byte-identical dry-run, required and forbidden destination fields, and explicit old-field absence. Criterion 8 is preserved by the added conflicting-`record` refusal test; no `--force`/`--replace` was added. Delivery-resolution source is outside the four-file diff and the adjacent delivery/topology regression set passes. Both EN and JA mirrors say that `guide bootstrap`'s human answer controls residence and that the command neither infers nor defaults it. This follows the node-08 rule that recorded residency chooses the delivery mechanism; it does not alter that mechanism or the `guide design-thread` rule.

Parent evidence is substantive: the PR pastes the parent probe's actual nonzero route result, and I independently verified that parent `ff11a355377fe2b1698cce1e14f39d8c79c20bd5` contains no `update-residence` / `UpdateResidence` route while the reviewed head does. `git diff --check` is clean. Local focused tests passed 17/17 and adjacent topology/delivery/docs tests passed 89/89. The PR's full Release evidence is `5381` passed, `1` skipped; both required GitHub checks are green at this exact head. I also invoked the full Release suite from an unregistered exact-SHA clone because this session is read-only for the supplied checkout; it did not reach a completion line before this session's execution window, so I do not represent a local full-suite count as passing.

Tests are necessary but not sufficient; packet/intent conformance and the no-scope-creep diff review above are the basis for this approval. `Closes #1655` correctly targets the linked source issue.

PR #1671, opened 2026-08-31
G766: fail closed for metadata-branch claim stores

### LINKED ISSUE #1669: G766: a host with claims on another branch is read as a host with no claims at all, and every scope passes
# G766: a host with claims on another branch is read as a host with no claims at all, and every scope passes

## Goal

Let the claim reader distinguish "this host does not use claims" from "this host uses claims, and none are on the canonical branch yet". Today it cannot, and it resolves the ambiguity in the permissive direction: every scope returns `not-configured` with `passed: true`, including scopes that never existed.

## Why This Slice Exists Now

`not-configured` exists for a real case — a host that never adopted claims must keep legacy single-team behaviour byte-for-byte. G755 preserved that deliberately, and it should stay.

But the reader infers "no claims store" from the **absence of a claims tree on the canonical branch**, and absence there is not the same fact. A same-repo host whose earlier claims were written to its metadata branch has a populated claims store; it simply has nothing on the canonical branch until its first new `acquire` lands. For that window the reader answers as though claims were never adopted, and **every claims-gated command skips mutual exclusion**.

This is the same shape as the fail-open G755 closed, arriving through a different door. G755 fixed the reader looking at the wrong branch. This is the reader looking at the right branch and drawing a conclusion the evidence does not support.

## Current Observed State

Measured on a build of current `main` (`intent-cli 0.27.1-6cc2b05`) in a same-repo Topology B clean room — `same_repo_topology = true`, `metadata_*_branch = "main-metadata"`, host checkout on `main-metadata`, claim records present on the metadata branch and none yet on canonical:

```
claims on origin/main            : 0 files
claims on origin/main-metadata   : 2 files

claim verify execution-unit:EU-20                -> not-configured / passed: TRUE
claim verify execution-unit:EU-99-never-existed  -> not-configured / passed: TRUE
```

A scope that has never existed passes. So does one whose record is sitting on the metadata branch.

The state flips as soon as any new `acquire` puts a tree on the canonical branch:

```
claim verify execution-unit:EU-20 -> unheld / passed: false
```

So a single host moves between a **fail-open** and a **false-negative** depending on whether one new claim has landed. Both answers are wrong about EU-20; they are wrong in opposite directions, and only the first one silently disables the gate.

**Provenance, stated plainly:** this was found while verifying G763, in a clean room that happened not to run an `acquire` first. The G763 packet measured only the `unheld` side, because the earlier fixture had done an `acquire` before reading. That was luck rather than method. The `unheld` case is what G763 addresses; this is the case nobody measured.

- **`not-configured` with `passed: true` must survive for a host that genuinely never adopted claims.** That is G755's acceptance criterion 6 and this slice must not break it; a host with no claims anywhere keeps legacy behaviour byte-for-byte.
- The reader stays canonical-only for ownership. This slice does not make ordinary `claim verify` consult the metadata branch — G763's safety criterion, and reversing it would reintroduce branch-local presence deciding ownership.
- The writer's target and the canonical-branch resolution are unchanged (G747, G755).
- `claim stranded` and `claim stranded migrate` (G763) are the migration path and are not changed here.

## Target Repo / Path / Part

- Target paths: `src/IntentSystem.Cli/Commands/ClaimOwnershipVerifier.cs`, `tests/IntentSystem.Cli.Tests/ClaimCommandG766Tests.cs`, `docs/en/05-implementation-loop.md`, `docs/ja/05-implementation-loop.md`

Target part: the no-store determination in the canonical-evidence read path.

## In Scope

- The reader distinguishes a host with no claims store anywhere from one whose claims exist off the canonical branch, using evidence it can already reach — a configured metadata branch carrying claim records is the signal that the host has adopted claims.
- When claims exist off-canonical and none are canonical, verification does **not** return `passed: true`. It fails closed with a distinct status naming the situation, so an operator is told what is actually true rather than being waved through.
- A genuinely unadopted host is unaffected: still `not-configured`, still `passed: true`, byte-identical output.
- Both documentation mirrors describe the three states and which one fails closed.

## Out Of Scope

- Making ordinary `claim verify` read ownership from the metadata branch. It must not; a stranded record must still not decide ownership.
- Changing the writer target or the canonical-branch resolution.
- Changing `claim stranded` / `claim stranded migrate`.
- Auto-migrating anything.
- Any change for hosts with no metadata branch configured.

The child PR makes the claim reader stop treating "no claims tree on the canonical branch" as proof that a host has no claims store. When a configured metadata branch carries claim records and the canonical branch carries none, verification must fail closed with a distinct status naming that situation, instead of returning `not-configured` with `passed: true` and letting every scope through. A host with no claims anywhere must keep returning `not-configured` with `passed: true`, byte-identical to today. Ordinary verification must still resolve ownership only from the canonical branch — a stranded record must not become a holder. Tests must cover the fail-closed case, the genuinely-unadopted case, a never-existed scope in each, and that ownership is still never read from the metadata branch.

## Acceptance Criteria

1. When a configured metadata branch carries claim records and the canonical branch carries none, `claim verify` returns `passed: false` with a distinct status naming the situation. A test asserts the status and that it is not `not-configured`.
2. A scope that never existed also returns `passed: false` in that state. A test asserts it — this is the case that shows the gate is closed rather than merely answering about known scopes.
3. A host with **no** claims anywhere still returns `not-configured` with `passed: true`, byte-identical to the parent commit. A test pins the full result payload, not just the status field.
4. A host with no metadata branch configured is unaffected. A test asserts parity with the parent.
5. **Ownership is still never read from the metadata branch.** In the fail-closed state, the result names no holder and no holder team, even though a record exists off-canonical. A test asserts the absence of holder fields — this is the criterion that keeps G763's safety property intact.
6. Once a canonical claims tree exists, behaviour is unchanged from today: a stranded scope reads `unheld`, a canonical scope reads `owned`. A test covers both.
7. Each new test is demonstrated failing on the parent commit, with the actual output pasted in the PR.
8. Both doc mirrors describe the three states and which one fails closed.
9. Full Release suite passes with the actual counts pasted.

## Verification

- New tests using a real bare-origin fixture with a canonical and a metadata branch — the defect is about which branch holds records, so a mock cannot exercise it.
- Criterion 3 is the compatibility criterion and criterion 5 is the safety criterion; state explicitly in the PR how each was verified.
- `git diff --check`.

## Related Links

- Issue #1658 — the consumer report whose host is in this shape; a follow-up comment there asks which of the two states that host is actually in.
- G763 (issue #1663, merged `6cc2b0512`) — the stranded-claim report and migration; addresses the `unheld` side and must keep its safety property.
- G755 (issue #1642, merged `9a30e95a`) — closed the original fail-open and preserved `not-configured` for genuinely unadopted hosts.

## Knowledge Maintenance

- ADR candidate: none. This corrects an inference; the canonical-ownership decision is unchanged.
- Diagram candidate: none.
- Docs update: `docs/en/05-implementation-loop.md`, `docs/ja/05-implementation-loop.md`
- Closeout writeback expected: yes — node 03 records that absence of evidence was read as evidence of absence, and that the fixture which would have caught it differed from the one used only by an incidental `acquire`.

## Guide Reachability (G645)

`no_role_facing_surface: true`. This slice adds no command, flag, or field; it changes when an existing verification fails closed. The surfaces and their guide routes are unchanged.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1669

## Summary

When origin/default is empty but a configured metadata branch still carries active claim records, canonical claim verification now returns the distinct `metadata-branch-only` refusal for both existing and never-existed scopes. It never reads a metadata record as ownership. A genuinely unadopted host, including a valid config with no metadata branch, retains the parent `not-configured` JSON and exit behavior. The worker claim, next-action, and issue-preflight consumers treat the new status as an ownership stop.

## Evidence

- Parent oracle: `6cc2b05127f7dc8c9080e425eb5af8e0e099ace7`. Before the implementation, the G766 focused tests reported `Test Run Failed`, `Total tests: 5`, `Passed: 2`, `Failed: 3`: the metadata-branch fail-open, adjacent-consumer bypass, and documentation guard each failed. The two compatibility tests (no-store payload and canonical-only ownership) passed on the parent as expected.
- Child focused G766: `Total tests: 5`, `Passed: 5`, `Failed: 0`.
- Claim-focused suite, including G743/G747/G755/G763 and adjacent worker/claim consumers: `Total tests: 121`, `Passed: 121`, `Failed: 0`.
- G613: `Total tests: 6`, `Passed: 6`, `Failed: 0`.
- Full Release: `Total: 5398`, `Passed: 5397`, `Failed: 0`, `Skipped: 1`; exit 0.
- `git diff --check`: clean.

## Scope and ownership boundary

The child next-action result was `action=wait`, issue `#1669`, with local claim holder `none`/unheld; supported child issue-preflight was `actionable=true`, `classification=ready-to-implement`, `claim_status=unheld`. This is the known child/local registry contradiction. The supplied authoritative host evidence remains: execution-unit:G766 held by actor `implementation`, team `intent-cli-dev`, claim file `4f246379560b484f0ba64801a6ff9aa6e0a34f1aa9c8fce3ca260264759af3be.json`, claimed at `2026-08-31T01:22:02.894245+00:00`, base `91903554f4d31d902899c858550ed538cab0d653`; the host preflight refusal was the unreadable `.git/FETCH_HEAD` boundary. No child execution-unit claim was created or mutated, and no host repository/state was entered.

Only claim verification, the three worker-consumer status gates, focused tests, and the two EN/JA implementation-loop mirrors changed. G763/G755 writer, migration, canonical branch, and claim transaction behavior remain untouched.

### DIFF
--- a/docs/en/05-implementation-loop.md
+++ b/docs/en/05-implementation-loop.md
@@ -268,6 +268,21 @@ uses that same judgment in recommendation mode: unheld and own-team units
 remain candidates, while claimed-elsewhere units are excluded with holder
 evidence, so it never urges what start will refuse.
 
+### Preview: canonical-empty hosts with metadata-branch claims (G766)
+
+`not-configured` is reserved for a host with no claims store anywhere. A configured
+metadata branch that still carries active claim records proves that the host has
+adopted claims, even when the canonical branch currently has no claim tree. In that
+state the verifier returns `metadata-branch-only` and fails closed for every scope,
+including a scope that never existed. It does not read ownership from the metadata
+branch: the result has no holder or holder team, and a stranded record remains
+unheld until the explicit G763 migration path is used.
+
+This keeps the genuinely unadopted host byte-compatible while preventing the empty
+canonical branch from making every claims-gated command pass. A configured metadata
+branch with no active records does not by itself create a claims store; the reader
+only changes the legacy answer when it can verify active records there.
+
 Numbering is claim-then-draft. Claim `execution-unit:<N>` before scaffolding;
 after losing N, fast-forward, recompute, and retry the next number exactly once.
 The GitHub lifecycle label remains visible defence in depth but is not the
--- a/docs/ja/05-implementation-loop.md
+++ b/docs/ja/05-implementation-loop.md
@@ -254,6 +254,21 @@ canonical answer を返します。default branch を解決または fetch で�
 unit は candidate のまま、claimed-elsewhere unit は holder evidence 付きで除外します。
 これにより start が拒否する作業を recommendation が促しません。
 
+### Preview: canonical branch が空で metadata branch のみに claim がある場合 (G766)
+
+`not-configured` は、どの branch にも claims store がない host のために予約します。
+設定済み metadata branch に active claim record が残っていれば、canonical branch に
+現在 claim tree がなくても、その host が claims を導入済みである証拠です。この場合
+verifier は `metadata-branch-only` を返して、まだ存在しない scope を含む全ての scope を
+fail closed にします。metadata branch の record から ownership は読みません。result に
+holder / holder team はなく、stranded record は明示的な G763 migration path を使うまで
+unheld のままです。
+
+これにより、claims をまだ導入していない host の byte-compatible な挙動を保ちながら、
+canonical branch が空であることだけを理由に claims-gated command が全て pass することを
+防ぎます。設定済み metadata branch に active record がないだけなら claims store とは判定せず、
+そこに active record があることを確認できた場合だけ legacy answer を変更します。
+
 番号は claim-then-draft です。scaffold 前に `execution-unit:<N>` を claim し、N に負けたら
 fast-forward、次番号を再計算し、その番号を exactly once retry します。GitHub lifecycle label
 は visible な defence in depth のままで acquisition fact ではありません。review/closeout gate と
--- a/src/IntentSystem.Cli/Commands/ClaimOwnershipVerifier.cs
+++ b/src/IntentSystem.Cli/Commands/ClaimOwnershipVerifier.cs
@@ -2,6 +2,7 @@ using System.Diagnostics;
 using System.Text;
 using System.Text.Json;
 using System.Text.Json.Serialization;
+using IntentSystem.Cli.Infrastructure;
 
 namespace IntentSystem.Cli.Commands;
 
@@ -63,6 +64,17 @@ internal static class ClaimOwnershipVerifier
                 Holder: null,
                 HolderTeam: null,
                 Detail: "No claims store is configured; legacy single-team behavior applies unchanged.");
+        }
+
+        if (evidence.MetadataBranchOnly)
+        {
+            return Refused(
+                ClaimOwnershipVerification.StatusMetadataBranchOnly,
+                scope,
+                invokingTeam,
+                null,
+                null,
+                evidence.Detail);
         }
 
         if (evidence.RecordJson is null)
@@ -228,6 +240,23 @@ internal static class ClaimOwnershipVerifier
         var storeConfigured = claimPaths.Length > 0;
         if (!storeConfigured)
         {
+            // G766: an empty canonical claims tree is not enough to infer that
+            // the host never adopted claims. A configured metadata branch is
+            // an additional read-only adoption signal, but it never supplies
+            // ownership for ordinary verification.
+            var metadataConfiguration = ReadConfiguredMetadataBranch(repoRoot);
+            if (metadataConfiguration.Error is not null)
+            {
+                return CanonicalClaimEvidence.Unavailable(metadataConfiguration.Error);
+            }
+
+            var metadataBranch = metadataConfiguration.Branch;
+            if (metadataBranch is not null
+                && !string.Equals(metadataBranch, canonicalBranch.Name, StringComparison.Ordinal))
+            {
+                return ReadMetadataBranchEvidence(repoRoot, metadataBranch);
+            }
+
             return CanonicalClaimEvidence.NoStore;
         }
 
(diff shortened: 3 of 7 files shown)


tomohisa (commented):
APPROVED — G766 satisfies the packet contract at `4be65f1d658ccdc46a1f2fef45270d56ad742eed`.

design_alignment_checked: true

I independently reproduced the parent oracle by checking the new five-test fixture into an isolated checkout of `6cc2b05127f7dc8c9080e425eb5af8e0e099ace7`: exactly three load-bearing failures (metadata-only verification, adjacent consumer refusal, and documentation guard) and exactly two compatibility passes. At the reviewed head the same five tests pass.

The real bare-origin fixture deliberately creates a metadata branch with active records and no canonical claims tree. It proves both the known and never-existed scopes return `passed: false` / `metadata-branch-only`; command JSON exits 1 and omits holder fields. The no-store fixture pins the full legacy JSON payload and exit 0, including the genuine no-metadata-branch configuration. With a canonical claim tree present, the existing state remains `owned` for the canonical record and `unheld` for the metadata-only record.

I inspected the status consumers: `WorkerClaimAnalyzer`, `WorkerNextActionAnalyzer`, and `WorkerIssuePreflightAnalyzer` each classify `metadata-branch-only` as a configured-store refusal. The two `ProbeStore` consumers remain fail-closed because the new evidence reports `StoreConfigured: true`, then their ordinary verification path returns the refusal. All other verifier callers reject its non-passing result directly.

The EN and JA implementation-loop mirrors accurately describe all three states, the fail-closed middle state, and canonical-only ownership. The change is consistent with ICL.M.STATE's G755/G763 claim-registry invariants: absence on canonical is no longer treated as absence of a store, while metadata records never decide ownership. Scope boundaries are preserved: no writer-target/default-branch resolution, stranded/migration, auto-migration, or no-metadata-host behavior changed.

Evidence: `dotnet test ... --filter FullyQualifiedName~ClaimCommandG766Tests` (5/5); parent oracle (3 failed, 2 passed as required); adjacent G755/G763 and worker-command coverage (110/110); `git diff --check`; and both exact-head CI checks are green. Tests are necessary, not sufficient; the approval also rests on the packet criteria and intent conformance above. `Closes #1669` is present and points to the linked issue.

PR #1673, opened 2026-08-31
G767: degrade supervision history reads per record

### LINKED ISSUE #1672: G767: one unparseable record silences every supervision answer for a team, including the liveness surface built for outages
# G767: one unparseable record silences every supervision answer for a team, including the liveness surface built for outages

## Goal

Make `intent-cli notify supervise liveness` — and every other read of a team's supervision state — survive an unparseable record in the append-only history. Today one damaged line reports the whole store as unreadable and the surface returns nothing. After this slice the surface answers from the records that parsed, and states in the payload how many it could not read and from which file, so a partial reading is never mistaken for a clean one.

## Why This Slice Exists Now

G765 shipped a read-only liveness surface whose entire purpose is to answer "is supervision alive?" on a host where supervision is dead and therefore cannot report on itself. Verifying that surface on the host that motivated it produced this:

```
$ intent-cli notify supervise liveness --domain intent-cli --team intent-cli-dev \
    --routing-root /Users/tomohisa/dev/GitHub/MyIntentHost --format json
{"operation":"supervise-liveness","command_mode":"read-only","success":false,
 "error":"Supervision state at '.../supervision/intent-cli/intent-cli-dev' could not be read: 'k' is invalid after a property name. Expected a ':'. Path: $ | LineNumber: 0 | BytePositionInLine: 21.",
 "commands_executed":"none"}
```

The detector built for the outage cannot answer during the outage. It fails closed, which is the right direction and is why this is a defect rather than an incident — but a surface that only ever says "I cannot read this" is not yet the surface G765 was for.

## Current Observed State

Measured on this host with a Release build of the merged G766 head, `intent-cli 0.27.1-7adb2b5`.

**The cause, located in code.** `NotifySupervisionStore.ReadCycles` iterates every line of `cycles.jsonl` and calls `JsonSerializer.Deserialize` with no per-line guard; the `JsonException` escapes the loop, and `Read()`'s catch converts it into `Failure(directory, "Supervision state ... could not be read")`. One malformed **record** is reported as an unreadable **store**. `ReadStalls` and `ReadPromptAudits` have the identical shape.

**The extent.** A full scan of `.intent-cli/supervision/intent-cli/intent-cli-dev/cycles.jsonl` — 718,903,202 bytes, 663,959 lines — finds **9 malformed lines from 5 independent incidents**: 9803/9804, 57687/57688, 150639, 157887, 183811, 278778, 618297. Every one is the same shape: a record's write was cut part-way and the remainder became its own line.

```
line 9803: {"kind":"cycle","cy{"kind":"cycle","cycle":{"cycle_id":"fe3e...
line 9804: ,"process_start_time_source":"process","host":"Mac"},"trigger":"event-...
```

**The proportion is the point.** 9 bad lines in 663,959 — 0.0014% — disable 100% of the liveness answer. The store is append-only and unbounded, so a single interrupted write is permanent: nothing later in the file can undo it, and hand-editing host state is forbidden.

**The correct shape already exists in the same file.** `notify supervise shrink --dry-run` reads the *same* corrupt directory and answers:

```json
{"operation":"supervise-shrink","supervisor_state":"unreadable",
 "before_record_count":673868,
 "files":{"cycles":{"before_bytes":718903202,"before_record_count":663959}}}
```

One reader degrades and names what it could not read. The other returns nothing. This slice makes the second behave like the first. The sanctioned repair paths are **not** deadlocked — that was checked rather than assumed.

**Control.** On the same host in the same session, `liveness --domain remote-herdr --team remote-herdr-dev` succeeds and returns `absent_since_last_cycle: true`. The failure is specific to the corrupt store, not to the surface.

- G765's liveness surface exists and is read-only: it runs no OS lifecycle query and reports `commands_executed: "none"`.
- G765's evidence split is settled: `scheduler_installation_evidence` and `scheduler_live_state` are separate fields, and an unregistered artifact is never reported as loaded.
- Supervision history is CLI-owned and runtime-local per G750; `cycles.jsonl` is gitignored.
- `notify supervise shrink` already reads a damaged directory and reports `supervisor_state: "unreadable"` while still counting records. Copy that proven shape rather than inventing a new tolerance model.

## Target Repo / Path / Part

- Paths: `src/IntentSystem.Cli/Commands/NotifySupervisionStore.cs`, `src/IntentSystem.Cli/Commands/NotifySuperviseLivenessCommand.cs`, `tests/IntentSystem.Cli.Tests/NotifySuperviseCommandTests.cs`, `docs/en/12-agent-message-orchestration.md`, `docs/ja/12-agent-message-orchestration.md`
- Part: the per-line deserialize in `ReadCycles` / `ReadStalls` / `ReadPromptAudits`, and the `Read()` catch that converts one bad record into a failed store read.

## In Scope

- Per-record tolerance in the three supervision history readers, with the unreadable count carried out to the caller.
- Surfacing that count and a partial-reading statement in the liveness payload and summary.
- Keeping a wholly unreadable history distinguishable from a genuinely empty one.
- Both doc mirrors.

## Out Of Scope

- **How the fragments were written.** Recorded here as a named residual with the evidence that exists, and deliberately not fixed by this slice. The cycle append at `NotifySupervisionStore.Append` takes `AcquireDirectoryLock` before `File.AppendAllText`, so a plain concurrent append does not explain it. Line 9803's spliced record carries `ts 2026-08-27T19:51:39` while lines 9802 and 9804 carry `06:56:49` — a *later* record was written over the middle of an *earlier* one, which points at an out-of-order write at a stale offset rather than two appenders racing at end-of-file. That is a hypothesis, not a reproduction. **Do not let a reader that tolerates damage be mistaken for the damage being addressed.**
- Repairing or rewriting existing damaged files; archive, rotation, and write-rate work.
- Any change to what liveness concludes from records that *do* parse.

## Acceptance Criteria

1. A supervision directory whose `cycles.jsonl` contains one unparseable line still produces a liveness answer: `success: true`, with the reading computed from the records that did parse; a test asserts the answer is produced and names the fixture line.
2. That answer states the damage in the payload — at minimum a count of unreadable records and the file they are in — and the human-readable summary says the reading is partial; a test asserts **both** the count field and the presence of the partial-reading statement in the summary, so a silent skip cannot pass.
3. The damage is never rounded down to nothing: with zero unreadable records the count field is present and zero, and the summary contains **no** partial-reading statement; a test pins both, so the two states print differently.
4. When **every** cycle record is unparseable the surface does **not** report health — it reports that no readable cycle exists, the same way it reports a genuinely empty history, and the unreadable count distinguishes the two; a test asserts an all-corrupt fixture and an empty fixture produce different payloads.
5. `stalls.jsonl` and prompt-audit records get the same treatment, since `ReadStalls` and `ReadPromptAudits` share the defect; a test covers a malformed `stalls.jsonl` line.
6. A directory with no damage returns a payload byte-identical to the parent apart from the new zero-valued count field; a test pins the full payload.
7. **Safety:** `commands_executed` remains `"none"` and no OS lifecycle query is introduced; a test asserts it, keeping G765's property intact.
8. Each new test is demonstrated failing on the parent commit, with the actual output pasted in the PR.
9. Both doc mirrors state that supervision history is read record-by-record, that unreadable records are counted and surfaced rather than hidden or fatal, and that a wholly unreadable history is not health.
10. Full Release suite passes with actual counts pasted.

## Verification

Build the branch head in Release and run against a fixture directory, not against a live host:

- **Discriminating pair for criteria 2 and 3** — a directory with one malformed line and an otherwise identical directory with none must produce **different** payloads: the first carries a non-zero unreadable count and a partial-reading statement, the second a zero count and no such statement. Before this slice the first produces no payload at all.
- **Discriminating pair for criterion 4** — an all-corrupt history and an empty history must produce different payloads. Both say "no readable cycle"; only the first carries a non-zero unreadable count. If these print identically, the fix has made corruption indistinguishable from emptiness, which is worse than the current failure.
- Paste the parent-commit failure output for every new test.

## Related Links

- Parent unit: G765 (PR #1670, merged `db5394d75`) — the liveness surface this slice repairs.
- G750 — supervision cycle history is CLI-owned and runtime-local.
- G734 — `notify supervise shrink`, the sibling reader whose degrade-and-name behaviour is the model here.

## Base Branch Policy

tomohisa:
Closes #1672

## Summary

- Read cycles, stalls, and prompt-audit JSONL records independently; valid records remain readable while malformed lines produce deterministic component/file/line/reason evidence.
- `notify supervise liveness --format json` now reports `success`, an always-present `unreadable_record_count`, optional `unreadable_records`, and a partial-reading summary. All-corrupt cycle history remains absent/unhealthy and is distinct from empty history.
- This is reader/reporting behavior only. It does not repair files, prevent fragments, or change append ordering.

## Parent reproduction

On parent `7adb2b5cac8090865d19c864842dbed48ffab7d2`, the four behavior fixtures failed: `Failed: 4, Passed: 0, Skipped: 0, Total: 4`. The parent failures were whole-store read errors for malformed cycle/stall/prompt lines and exit code 1 for the liveness cases. The parent documentation probe also failed: `Failed: 2, Passed: 0, Skipped: 0, Total: 2` because neither mirror contained G767 corruption evidence.

## Verification

- G767 focused: `Failed: 0, Passed: 6, Skipped: 0, Total: 6`.
- Existing supervision/archival/liveness adjacency: `Failed: 0, Passed: 35, Skipped: 0, Total: 35`.
- G613: `Failed: 0, Passed: 6, Skipped: 0, Total: 6`.
- Full Release: `Failed: 0, Passed: 5411, Skipped: 1, Total: 5412`.
- `git diff --check`: clean.
- Head: `6deeef6e911a6bed8646b73d8325e8a2f7349aa0`.

The child issue preflight/next-action refused only because `.git/FETCH_HEAD` was unreadable (`claim-unavailable` / `canonical-unavailable`); the supplied authoritative host claim for `execution-unit:G767` was consumed. No child execution-unit claim or host state was created, modified, verified, or released.

### DIFF
--- a/docs/en/12-agent-message-orchestration.md
+++ b/docs/en/12-agent-message-orchestration.md
@@ -1025,6 +1025,34 @@ registration; registration remains the operator's explicit action.
 > change to interval, bound, cycle, shrink, archive, or repair semantics is
 > introduced.
 
+### Record-by-record supervision history reads (G767 — preview-through-1.x)
+
+Supervision history is read one non-blank JSONL record at a time. A malformed
+line in `cycles.jsonl`, `stalls.jsonl`, or the prompt-audit records does not
+make the whole team directory unreadable: valid records remain available and
+the read result carries deterministic corruption evidence with the component,
+relative file, one-based line number, and reason. The liveness JSON always
+includes `unreadable_record_count`; a non-zero count also includes
+`unreadable_records`, and its human-readable summary says that the reading is
+partial. A successful command reports `success: true`; a clean directory
+therefore reports zero without a partial-reading statement.
+
+If every cycle line is malformed, liveness has no readable cycle and does not
+report supervision as healthy. Its non-zero unreadable count and file/line
+evidence still distinguish that state from a genuinely empty history. This is
+the same degrade-and-report shape used by the sanctioned shrink reader: the
+reader preserves evidence and names the damaged input rather than silently
+turning corruption into absence. The measured incident was 9 malformed lines
+from 5 incidents in 663,959 lines (0.0014%); this slice changes how readers
+report that evidence, not how fragments were written or how appends are
+ordered. It does not repair existing files or claim to prevent future
+corruption.
+
+The reader remains read-only and does not run a supervisor or an OS lifecycle
+command. `stalls.jsonl` and prompt-audit records receive the same per-record
+treatment as cycles, so a damaged secondary history cannot silently erase the
+rest of the team's supervision answer.
+
 ### Event-driven supervision (G659 — preview-through-1.x)
 
 Add `--event-mode` to the one standing `notify supervise` invocation to opt in.
--- a/docs/en/12-agent-message-orchestration.md
+++ b/docs/en/12-agent-message-orchestration.md
@@ -1034,8 +1034,9 @@ the read result carries deterministic corruption evidence with the component,
 relative file, one-based line number, and reason. The liveness JSON always
 includes `unreadable_record_count`; a non-zero count also includes
 `unreadable_records`, and its human-readable summary says that the reading is
-partial. A successful command reports `success: true`; a clean directory
-therefore reports zero without a partial-reading statement.
+partial. A successful command returns exit code 0 and omits the failure-only
+`success` field; a clean directory therefore reports zero without a
+partial-reading statement.
 
 If every cycle line is malformed, liveness has no readable cycle and does not
 report supervision as healthy. Its non-zero unreadable count and file/line
--- a/docs/ja/12-agent-message-orchestration.md
+++ b/docs/ja/12-agent-message-orchestration.md
@@ -862,6 +862,31 @@ action のままです。
 > lifecycle の実行、automatic re-arming、interval / bound / cycle / shrink / archive /
 > repair semantics の変更はありません。
 
+### supervision history の record 単位 read と corruption evidence (G767 — preview-through-1.x)
+
+supervision history は、空白でない JSONL record を 1 行ずつ読み取ります。
+`cycles.jsonl`、`stalls.jsonl`、prompt-audit record のどれかに malformed な
+line があっても team directory 全体を unreadable にはしません。読める record は
+残し、読み取り結果には component、relative file、1-based line number、reason を持つ
+deterministic な corruption evidence を出します。liveness JSON には常に
+`unreadable_record_count` を含め、0 より大きい場合は `unreadable_records` も含めます。
+human-readable summary には reading が partial であることを明示します。成功した
+command は `success: true` を返します。したがって clean な directory は count=0 で、
+partial-reading statement を出しません。
+
+cycle line がすべて malformed の場合は readable cycle がないため、supervision を
+healthy とは報告しません。それでも non-zero の unreadable count と file/line evidence
+によって、本当に empty な history と区別できます。これは sanctioned な shrink reader
+と同じ degrade-and-report の形です。corruption を absence に黙って変換せず、読める
+evidence を保持して damaged input を示します。測定された incident は 663,959 行中
+5 incident、9 malformed line（0.0014%）でした。この unit が変えるのは reader の
+evidence の示し方だけであり、fragment がどう書かれたかや append order は調査・修正
+しません。既存 file を修復せず、将来の corruption を防ぐとも主張しません。
+
+reader は read-only のままで、supervisor や OS lifecycle command を実行しません。
+`stalls.jsonl` と prompt-audit record にも cycles と同じ record 単位の扱いを適用するため、
+secondary history の damage が team の supervision answer の残りを黙って消すことはありません。
+
 ### event-driven supervision (G659 — preview-through-1.x)
 
 1 つの standing `notify supervise` invocation に `--event-mode` を追加して有効化します。
--- a/docs/ja/12-agent-message-orchestration.md
+++ b/docs/ja/12-agent-message-orchestration.md
@@ -871,8 +871,8 @@ line があっても team directory 全体を unreadable にはしません。�
 deterministic な corruption evidence を出します。liveness JSON には常に
 `unreadable_record_count` を含め、0 より大きい場合は `unreadable_records` も含めます。
 human-readable summary には reading が partial であることを明示します。成功した
-command は `success: true` を返します。したがって clean な directory は count=0 で、
-partial-reading statement を出しません。
+command は終了コード 0 を返し、失敗応答専用の `success` field は含めません。したがって
+clean な directory は count=0 で、partial-reading statement を出しません。
 
 cycle line がすべて malformed の場合は readable cycle がないため、supervision を
 healthy とは報告しません。それでも non-zero の unreadable count と file/line evidence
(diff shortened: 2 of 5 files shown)


tomohisa (commented):
REQUEST-UPDATE — G767 has one implementation finding at `6deeef6e911a6bed8646b73d8325e8a2f7349aa0`.

design_alignment_checked: true

1. **Implementation finding — issue #1672 acceptance criterion 6 (clean payload compatibility).** `NotifySuperviseLivenessCommand` now adds `"success": true` to every successful liveness JSON response. The parent normal-result model has no `success` field; criterion 6 permits a clean directory to differ from the parent only by the new zero-valued `unreadable_record_count`. Thus clean output currently differs by two fields, not one. `CleanAndOneMalformedDiscriminatingPair_ProducesDifferentAnswers_G767` asserts only selected fields, so it does not detect this API change.

   Please preserve the parent clean payload byte-for-byte except for `unreadable_record_count: 0`, and add a full clean-payload fixture/assertion that would fail for any other added, removed, or changed field. If a `success: true` marker is needed for the damaged-answer contract, make it conditional to the damaged response rather than changing the clean compatibility surface.

I independently reproduced the parent oracle by running the new class against `7adb2b5cac8090865d19c864842dbed48ffab7d2`: all four behavioral fixtures and both documentation probes failed (6/6). At the reviewed head the focused class passes (6/6), and unmodified G765 liveness/shrink coverage passes (8/8). The three readers are guarded, evidence is deterministic and surfaced in payload and summary, discriminating pairs are present, docs state the reader-only boundary, no writer/repair/rotation scope was added, G765 remains read-only, CI is green, and `Closes #1672` is correct. These checks are necessary but not sufficient: the clean-output compatibility clause above remains unmet.

tomohisa (commented):
APPROVED — G767 now satisfies the amended packet contract at `7086571ec3e0000d495110ab95873a5f7514b0ec`.

design_alignment_checked: true

The sole prior finding is fixed. Successful clean and damaged liveness responses now omit `success`; the unchanged failure emitter retains the parent `success: false` convention. `AssertCleanPayloadMatchesParentOracle` enumerates every clean JSON property name and raw value in serialization order, permits only `unreadable_record_count: 0` beyond the parent clean shape, and explicitly rejects `success`. The focused tests also assert the absence of `success` for the one-malformed and all-corrupt successful responses and assert `success: false` on the failure path.

The original G767 contract remains intact: all three readers (cycles, stalls, prompt audits) preserve valid records while recording deterministic component/file/line/reason evidence; the unreadable count and partial-reading summary distinguish damaged success from clean; and all-corrupt history remains non-healthy and distinguishable from empty. The EN/JA mirrors state the record-level reader behavior, count/evidence, failure-only `success` convention, and the reader-only boundary. No writer, fragment-cause, repair, rotation, archive, or parsed-record conclusion changed. G765's read-only liveness invariant remains: `commands_executed` is `none` and no OS lifecycle query was introduced.

I independently built and tested the exact head: `NotifySupervisionReaderCorruptionG767Tests` passes 7/7 and unmodified G765 liveness/shrink coverage passes 8/8. The prior parent oracle was independently reproduced as six failures (four behavioral fixtures and two documentation probes). Design's independent real-host Release verification of `intent-cli 0.27.1-7086571-G767` also agrees with its separate scan: the 718,903,202-byte / 663,959-line store yields the same nine malformed line numbers, a partial liveness answer, `commands_executed: none`, and no `success` key. Exact-head CI is green and `git diff --check` is clean.

Tests are necessary, not sufficient: approval also verifies the packet's AC6 compatibility limit, the three-reader and discriminating-pair contract, and consistency with ICL.M agent-message-orchestration / G765's evidence-honesty invariant. `Closes #1672` is present and correct.

PR #1676, opened 2026-08-31
G768: make supervision history appends atomic per record

### LINKED ISSUE #1674: G768: supervision appends are not atomic, so a second writer silently destroys records that were already accepted
# G768: supervision appends are not atomic, so a second writer silently destroys records that were already accepted

## Goal

Make a supervision record, once written, survive the presence of a second writer. Today two concurrent processes appending to one team's store overwrite each other's records: the store accepts a write, reports success, and the record is gone. After this slice, two processes each appending N records produce exactly 2N records — and that guarantee holds even when only one of them cooperates with the directory lock, which is the situation that actually occurred on this host.

## Why This Slice Exists Now

G767 makes supervision readers survive damaged records. It does not stop the damage. This is where the damage comes from, and it is not hypothetical: it happened on this host, for days, while a detector reported the condition 112 times and nobody could see the consequence.

## Current Observed State

**Reproduced today**, Release build, net10.0, three consecutive runs. Two processes each call `File.AppendAllText` 600 times with a ~1.1 KB record into one file, no directory lock:

| run | records present | expected | lost | writer A | writer B |
|---|---|---|---|---|---|
| 1 | 1198 | 1200 | **2** | 600/600 | 598/600 |
| 2 | 1193 | 1200 | **7** | 596/600 | 597/600 |
| 3 | 1193 | 1200 | **7** | 598/600 | 595/600 |

Zero malformed lines in this shape — the records are overwritten **whole**. Both writers lose records; neither "wins".

**Controls, same session.** One process, 50 records → 50/50, zero loss. Two processes using real `O_APPEND` (Python `open(path,'a')`), 800 records each → **1600/1600, zero loss**. The loss is specific to the .NET append path under concurrency, not to concurrency itself.

**Production evidence**, this host's own store — this is what led here. `cycles.jsonl`: 718,903,202 bytes, 663,959 lines, **9 unparseable lines across 5 independent incidents** (lines 9803/9804, 57687/57688, 150639, 157887, 183811, 278778, 618297).

**The writers are identified, not guessed.** G676 records writer identity on every cycle record. At each incident the records on either side of the fragment carry two different pids, milliseconds apart:

```
150638  pid 43854  started 2026-08-25T22:21:23  completed 2026-08-28T00:42:48.035
150639  <41-byte orphaned tail>
150640  pid 59770  started 2026-08-27T06:43:13  completed 2026-08-28T00:42:48.043
```

**Every incident postdates the lock.** G734 introduced the cross-process `AcquireDirectoryLock` on 2026-08-25 (`4aea6b5e`); before it, `Append` used only a process-local `lock (Sync)` plus `File.AppendAllText`. All seven incidents are dated 2026-08-27 → 2026-08-29. **The lock was in the tree and the corruption happened anyway.**

**Why the lock did not help — this is the design finding.** `AcquireDirectoryLock` protects only writers that take it. A long-lived supervisor started before the lock shipped keeps running the old binary and never participates. Two such processes were alive simultaneously for days. *A defence that every participant must opt into is not a defence — it is a convention, and a stale process is under no obligation to honour it.*

**The duplicate was detected; the cost was underestimated.** G676's `duplicate-supervisor` finding fired **112 times** for this team. The detector worked perfectly. G676's decision was detection-only, on the stated basis that duplicates cost duplicate wakes and that intent-cli never kills, stops, ranks, elects, locks or leases a supervisor. This evidence does not reopen that boundary — it shows the cost also included permanent, unrepairable destruction of an append-only store, which is a different and larger cost than the one the decision weighed.

In production the same mechanism yields fragments rather than clean loss: when the overwriting record is shorter than its victim, the victim's tail survives as its own line. Clean loss and fragments are one defect at different relative record lengths.

- G676 detects duplicate supervisors and, by design, does not act on them. That boundary stands.
- G734's directory lock exists and is correct for writers that take it.
- G750: cycle history is CLI-owned and runtime-local.
- G767 (dependency) gives readers per-record tolerance. This slice reduces the damage produced; it does not remove the need to read damaged history.

## Target Repo / Path / Part

- Paths: `src/IntentSystem.Cli/Commands/NotifySupervisionStore.cs`, `tests/IntentSystem.Cli.Tests/NotifySupervisionStoreConcurrencyTests.cs`, `docs/en/12-agent-message-orchestration.md`, `docs/ja/12-agent-message-orchestration.md`
- Part: `NotifySupervisionStore.Append` and the cross-process `AcquireDirectoryLock` that is currently the only thing preventing loss.

## In Scope

- Per-record append atomicity that does not depend on every writer cooperating.
- The same guarantee for `stalls.jsonl` and the prompt-audit log, which share the append.
- Both doc mirrors and an ADR recording the guarantee.

## Out Of Scope

- Killing, stopping, ranking, electing, or leasing supervisors. G676's boundary is not reopened.
- Repairing existing damaged files, archive/rotation/shrink behaviour, write-rate work.
- Changing the record format, ordering, or what supervision concludes.

## Acceptance Criteria

1. Two concurrent OS processes each appending N supervision records to one team's store produce exactly 2N records, all parseable, with no record from either writer missing; a test drives real concurrent writers and asserts the exact count and that both writers' full record sets are present.
2. That same test **fails on the parent commit**, and the parent failure output with its actual measured loss is pasted in the PR. A test that passes on the parent proves nothing here: the loss is intermittent and a single lucky run looks clean.
3. The guarantee holds when only **one** of the two writers cooperates with the directory lock — the production situation. A writer that does not take the lock must still not destroy a record another writer already wrote; a test covers a locked writer concurrent with an unlocked one.
4. The guarantee is asserted for `stalls.jsonl` and the prompt-audit log as well as `cycles.jsonl`.
5. Single-writer behaviour is unchanged: record bytes, ordering, and file format identical to the parent for a serial sequence of appends; a test pins the produced file byte-for-byte.
6. At least 500 appends per writer in the concurrency test rather than a handful — the defect did not appear at small counts in early probing. State the chosen N and why, in the test.
7. **Safety:** no supervisor process is killed, stopped, ranked, elected, or leased, and no lifecycle command is introduced; confirmable by inspection.
8. Both doc mirrors state that supervision appends are atomic per record against non-cooperating writers, and that the directory lock is no longer the only thing standing between a duplicate supervisor and destroyed history.
9. Full Release suite passes with actual counts pasted.

## Verification

- Run the concurrency test **repeatedly** (at least 5 runs) on the branch head and report every run. One clean run is not evidence — the parent produced 2, 7 and 7 losses on three runs, so variance is expected and only a consistent zero across runs is meaningful.
- Run the same test on the parent and paste the measured loss.
- Criterion 3 is the one that matters most: it is the only one that reproduces the production situation, where one writer was a stale binary that could not have taken the lock.

## Related Links

- Depends on: G767 (issue #1672) — per-record read tolerance; touches the same file.
- G734 — introduced the directory lock (`4aea6b5e`, 2026-08-25).
- G676 — duplicate-supervisor detection; fired 112 times here.
- G750 — cycle history ownership.

## Base Branch Policy

Branch from `main` of `J-Tech-Japan/intent-system`, after G767 has merged. No rebase onto other feature branches.

tomohisa:
Closes #1674

## Summary

- Replace the regular supervision event File.AppendAllText path with one OS append primitive for each already-serialized UTF-8 JSONL record.
- Keep the existing per-team .supervision.lock for cooperating writers and recovery; the native append contract also protects a writer that does not take that lock.
- Cover cycle, stall, and prompt-audit records without changing their JSONL bytes, read behavior, shrink behavior, or lifecycle policy.
- Add the G768 ADR and matching EN/JA orchestration guidance. No kill, stop, rank, election, or lease behavior is added.

## Required evidence

- Parent regression (before AtomicAppendWriter, parent implementation at 4dcf1916 plus the G768 harness): 991 parseable records in 1000 nonblank lines; expected 1000; loss=9; unreadable_lines=9; missing=cycle-unlocked-0023,cycle-unlocked-0043,cycle-unlocked-0081,cycle-unlocked-0127,cycle-unlocked-0266,cycle-unlocked-0339,cycle-unlocked-0377,cycle-unlocked-0403,cycle-unlocked-0450.
- Final branch-head concurrency test: 5/5 repetitions passed with zero loss and zero unreadable lines. Each repetition runs two real dotnet test OS processes, one through the product store and directory lock and one through the no-lock append path; each record class runs 500 records per writer (1000 expected), for cycles, stalls, and prompt-audit records.
- Focused G768 harness plus docs/ADR guard: 4 passed, 0 failed.
- Supervision-adjacent suite (NotifySupervision filter): 98 passed, 0 failed.
- G613 Japanese terminology guard: 6 passed, 0 failed.
- Full Release suite: Failed: 0, Passed: 5416, Skipped: 1, Total: 5417.
- Serial cycle output is byte-identical to the parent oracle (the test compares the complete raw file as base64).
- git diff --check: clean.
- Head: 7dea6cb8b51b8dba7b5ec028f0d8ab6f9b9f1fcd.

The supplied host authority holds execution-unit:G768 on origin/main. Child next-action reported the scope unheld and child issue-preflight reported missing-target-declaration/canonical local metadata; this is the known child/host registry boundary. No child execution-unit claim or host state was created or mutated. Host follow-up should verify linkage; linked_pr_synced=false is expected.

### DIFF
--- a/docs/adr/0014-supervision-atomic-per-record-appends.md
+++ b/docs/adr/0014-supervision-atomic-per-record-appends.md
@@ -1,0 +1,51 @@
+# ADR 0014: supervision history appends are atomic per record
+
+- Status: Accepted (preview-through-1.x)
+- Date: 2026-08-30
+- Deciders: Design, orchestration, implementation, and review; recorded by G768
+- Related: G734, G744, G751, G767, G768 / #1674
+
+## Context
+
+Supervision history is append-only JSONL, and one team can have a cooperating
+supervisor plus another writer that does not take the team's directory lock.
+A `File.AppendAllText` open/write sequence allowed those writers to race so
+records could be lost or a line could be malformed. The existing directory
+lock coordinates cooperating writers and recovery, but it cannot protect a
+writer that does not participate in that lock.
+
+## Decision
+
+1. The regular supervision event append path serializes the complete JSONL
+   record as UTF-8 without a BOM and writes it with one OS append primitive.
+   Unix uses `O_APPEND`; Windows opens with `FILE_APPEND_DATA`. The complete
+   record is submitted in one native write, and no seek-to-end emulation is
+   used.
+2. The existing per-team `.supervision.lock` remains in place. It continues
+   to coordinate recovery, evidence-definition maintenance, and cooperating
+   writers; the OS append primitive additionally protects the record boundary
+   for an uncooperative writer.
+3. The shared append path is used by cycle, stall, and prompt-audit records.
+   The serial bytes, JSONL field order, event semantics, and read behavior are
+   unchanged. Existing shrink/audit and repair semantics remain separate.
+4. This is an integrity guarantee, not a lifecycle policy: intent-cli does not
+   kill, stop, rank, elect, or lease supervisor processes, and it does not
+   repair pre-existing corruption or arbitrary non-append file rewrites.
+
+## Consequences
+
+- One locked writer and one writer outside the directory lock retain every
+  complete record when both use the OS append contract.
+- A serial fixture can compare raw bytes with the parent implementation, so
+  the concurrency repair does not silently change the persisted format.
+- Existing damaged files and writers that replace or truncate files remain
+  outside this append-boundary decision.
+
+## Rejected alternatives
+
+- Rely only on the directory lock: rejected because the measured losing writer
+  did not acquire it.
+- Coordinate writers through a second process, watcher, election, or lease:
+  rejected because that would cross the no-lifecycle boundary.
+- Change JSONL framing or add a repair/compaction pass: rejected because this
+  unit changes the append primitive only and preserves existing history behavior.
--- a/docs/en/12-agent-message-orchestration.md
+++ b/docs/en/12-agent-message-orchestration.md
@@ -1054,6 +1054,23 @@ command. `stalls.jsonl` and prompt-audit records receive the same per-record
 treatment as cycles, so a damaged secondary history cannot silently erase the
 rest of the team's supervision answer.
 
+### Atomic per-record supervision history appends (G768 — preview-through-1.x)
+
+Supervision history writes encode each cycle, stall, and prompt-audit JSONL
+record as one UTF-8 append operation. The existing per-team
+`.supervision.lock` still coordinates cooperating writers and recovery, while
+the OS append primitive also protects record boundaries when another writer
+does not take that directory lock.
+
+The append is one complete record with no change to JSONL field order, line
+encoding, or the serial writer output. This preserves every record in the
+locked-versus-uncooperative-writer case without changing what supervision
+reads or writes. It does not repair existing corruption or claim to prevent
+fragments created by other I/O mechanisms.
+
+> **Preview through 1.x (G768).** This is an append-integrity guarantee only;
+> it does not add supervisor killing, stopping, ranking, election, or leasing.
+
 ### Event-driven supervision (G659 — preview-through-1.x)
 
 Add `--event-mode` to the one standing `notify supervise` invocation to opt in.
--- a/docs/ja/12-agent-message-orchestration.md
+++ b/docs/ja/12-agent-message-orchestration.md
@@ -887,6 +887,21 @@ reader は read-only のままで、supervisor や OS lifecycle command を実�
 `stalls.jsonl` と prompt-audit record にも cycles と同じ record 単位の扱いを適用するため、
 secondary history の damage が team の supervision answer の残りを黙って消すことはありません。
 
+### supervision history の record 単位 append (G768 — preview-through-1.x)
+
+supervision history の cycle、stall、prompt-audit は、1 つの UTF-8 append 操作で
+完全な JSONL record として書き込みます。既存の team 単位の
+`.supervision.lock` は協調する writer と recovery の調整に使い続けますが、別の writer が
+その directory lock を取得しない場合も OS の append primitive が record の境界を守ります。
+
+1 回の append に含めるのは完全な record であり、JSONL の field 順、line encoding、単独 writer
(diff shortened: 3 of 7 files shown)


tomohisa (commented):
APPROVED — reviewed at `7dea6cb8b51b8dba7b5ec028f0d8ab6f9b9f1fcd`.

`design_alignment_checked: true` — checked the G768 packet and review-context, the ICL.M / node-08 orchestration intent, ADR 0014, and both EN/JA orchestration mirrors.

- I reproduced the parent failure at `4dcf1916a94dfb871a1249fd60a3a4569b0a032c` using the submitted two-real-OS-process harness: one product-store writer taking the directory lock and one no-lock append writer. It produced 985 parseable records from 1,000 nonblank lines (15 missing/unreadable), so the regression test is load-bearing.
- On this head I ran that exact concurrency case five times. Every run passed for cycles, stalls, and prompt-audit records: 500 records from each writer per kind, no unreadable lines, and exact expected-ID equality. The shared store serializes once then calls `AtomicAppendWriter.Append` while retaining the existing directory lock; the writer uses one native `O_APPEND` write on Unix and `FILE_APPEND_DATA`/`WriteFile` on Windows, with short writes rejected. It introduces no kill/stop/rank/election/lease or repair behavior.
- The serial-cycle raw-byte parent-compatibility fixture passes, and the G767 corruption-reader test file is unmodified; its 7 focused cases pass. I also checked Design's exact-head real-host evidence: the original 9 corrupt records/lines remain unchanged, no success is reported and no commands run, and the damaged store remains 718,903,202 bytes / 663,959 lines. That is compatibility evidence, not a claim that this append-only change repairs old damage.
- Focused results: G768 concurrency 3/3 plus five stress repetitions; G768 ADR/docs 1/1; NotifySupervision adjacency 98/98; G613 Japanese terminology 6/6; G767 corruption reader 7/7. GitHub's two required checks are green at the same head, and the diff is limited to the append implementation, its tests, ADR, and EN/JA documentation.

PR #1677, opened 2026-08-31
G769: honor explicit routing root for liveness

### LINKED ISSUE #1675: G769: supervise liveness discards --routing-root, answers about wherever it was run from, and reports the root it ignored
# G769: supervise liveness discards --routing-root, answers about wherever it was run from, and reports the root it ignored

## Goal

Make `notify supervise liveness` read the supervision state under the `--routing-root` it was given, and make the `routing_root` it prints always name the root it actually read. Today the flag is accepted, validated, thrown away, and echoed back as though it had been honoured.

## Why This Slice Exists Now

The wrong answer this produces is **"supervision is absent"**. The documented operator response to absent supervision is to install and register a supervisor. Two supervisors on one team is exactly the G768 corruption condition — so this defect can manufacture the damage G768 exists to prevent.

It also defeats G767, which shipped hours earlier: from the wrong directory the surface reports `unreadable_record_count: 0` on a store with 9 known damaged records, so the corruption accounting reads clean.

## Current Observed State

Measured on this host, Release build of the merged G767 head, `intent-cli 0.27.1-4dcf191-G767`. **Same command, same explicit `--routing-root`, different shell directory:**

| cwd | answer |
|---|---|
| host root | `unreadable_record_count: 9`, "the last completed supervision cycle was 303835s ago" |
| `<host>/submodules/intent-system` | `unreadable_record_count: 0`, "No completed supervision cycle is recorded" |
| `/tmp` | same as above |

**All three print `routing_root: /Users/tomohisa/dev/GitHub/MyIntentHost`.**

**The cause is two lines.** `NotifySuperviseLivenessCommand.cs`:

```csharp
:50   _ = Path.GetFullPath(options.RoutingRoot ?? context.RepoRoot);   // computed, DISCARDED
:51   artifactRoot = context.ResolveSupervisionArtifactRootPath();      // working-directory derived
…
:95   RoutingRoot = Path.GetFullPath(options.RoutingRoot ?? context.RepoRoot),   // echoed into the payload
```

Line 50 evaluates the resolved root only for its validation side effect and discards it. Line 95 then reports it. The two lines sit six lines apart.

**Scoped by measurement, not assumption.** Only this surface is affected. Run from the host root and from `/tmp` with identical arguments: `notify status` is byte-identical, `notify supervise shrink --dry-run` is byte-identical. The sibling `shrink` resolves the artifact root the same way but declares no `--routing-root` at all, so it never claims a root it did not use. This is a **G765 residual**, introduced with the liveness surface itself.

**Provenance, stated plainly.** Found because I nearly filed a false regression. Right after merging G767 I re-ran the surface from a different shell directory, saw the count drop from 9 to 0, and began to report that the merge had broken it. Re-running from the identical directory first is what turned a wrong report into this finding.

- G765's liveness surface is read-only: `commands_executed: "none"`, no OS lifecycle query.
- G767's per-record tolerance and its clean-payload parent oracle (`AssertCleanPayloadMatchesParentOracle`) are in place and must keep passing.
- `notify status` and `notify supervise shrink` are already cwd-independent; they are the reference behaviour.

## Target Repo / Path / Part

- Paths: `src/IntentSystem.Cli/Commands/NotifySuperviseLivenessCommand.cs`, `tests/IntentSystem.Cli.Tests/NotifySuperviseCommandTests.cs`, `docs/en/12-agent-message-orchestration.md`, `docs/ja/12-agent-message-orchestration.md`
- Part: artifact-root resolution at `:47-51`, and the `RoutingRoot` echoed into the payload at `:95`.

## In Scope

- Reading state under the supplied `--routing-root`.
- Making the reported `routing_root` always name the root actually read.
- Distinguishing "no supervision state at that root" from "empty history".
- Both doc mirrors.

## Out Of Scope

- Any change to what liveness concludes from state it does read.
- G767's tolerance behaviour, G768's writer work, archive/shrink/rotation.
- Changing the default when `--routing-root` is omitted.

## Acceptance Criteria

1. With an explicit `--routing-root`, the command reads state under **that** root and produces a byte-identical payload regardless of the process working directory; a test runs the same arguments from at least two different working directories and asserts the payloads are equal.
2. That test **fails on the parent commit**, and the parent's two differing payloads are pasted in the PR.
3. With **no** `--routing-root`, behaviour is unchanged from the parent; a test pins that the working-directory-derived root is still used, so this fix does not quietly change the default.
4. When `--routing-root` names a path with no supervision state for the requested domain and team, the answer says so as a **distinct condition** and does not say "No completed supervision cycle is recorded" — not-found and empty-history are different facts; a test asserts they produce different payloads.
5. `routing_root` in the payload always names the root that was actually read; a test asserts the reported value and the directory the state came from are the same, so the payload can never again describe a location it did not use.
6. **Safety:** `commands_executed` remains `"none"` and no OS lifecycle query is introduced; a test asserts it.
7. G767's clean-payload parent oracle still passes unchanged, proving the payload shape did not drift while fixing the root; state its result in the PR.
8. Both doc mirrors state that `--routing-root` selects the state that is read, and that omitting it reads relative to the working directory.
9. Full Release suite passes with actual counts pasted.

## Verification

- **Criterion 1 is a discriminating pair in reverse**: two runs that currently print *differently* must, after the fix, print *identically*. Paste both payloads for both commits.
- **Criterion 4 is the one that prevents a lazy fix.** Making the surface honour the root is easy; making it distinguish "nothing there" from "nothing recorded" is what stops the next operator from reading a typo'd path as a dead supervisor.
- Re-run G767's tests unchanged (criterion 7).

## Related Links

- Residual of: G765 (PR #1670) — introduced this surface.
- Endangers: G768 (#1674) — a false "absent" can drive an operator into creating the duplicate supervisor that corrupts the store.
- Undermines: G767 (PR #1673, merged `4dcf1916a`) — corruption accounting reads clean from the wrong directory.

## Base Branch Policy

tomohisa:
Closes #1675

## Summary

- Resolve an explicit `--routing-root <host-root>` as the base for the configured supervision artifact store instead of reading from the caller's current working directory.
- Report the supplied absolute root in `routing_root` and add explicit JSON/markdown `supervision_state` values: `not-found`, `empty-history`, or `recorded`.
- Keep the omitted-root path and payload byte-compatible, including the unchanged G767 clean payload oracle.
- Keep `notify status`, `notify supervise shrink`, lifecycle-query boundaries, and all G765/G767 behavior unchanged.

## Evidence

- Parent oracle run before the source fix:
  `dotnet test tests/IntentSystem.Cli.Tests/IntentSystem.Cli.Tests.csproj --no-restore --filter 'FullyQualifiedName~Liveness_ExplicitRoutingRoot' --verbosity minimal`
  failed 2/2 as required. The explicit-root test reported the expected/actual cycle timestamp difference at position 305 (expected the routing-root cycle at 11:00, actual the working-directory cycle at 10:00). The not-found/empty-history test failed with `KeyNotFoundException` for the absent `supervision_state` property.
- The issue packet's three-cwd reproduction measured `unreadable_record_count=9` at the host root and `0` from both a submodule and `/tmp`; all three parent responses printed the host root, proving the explicit root was ignored.
- Post-fix focused G769: 2 passed, 0 failed.
- Preserved liveness plus unchanged G767 corruption oracle and G613: 23 passed, 0 failed.
- Adjacent supervision/reconcile/shrink/archive/runtime-local/event-mode/installation/concurrency coverage: 64 passed, 0 failed.
- Full CLI Debug suite: 5,418 passed, 1 skipped, 0 failed, 5,419 total.
- Full Release suite (`dotnet build IntentSystem.sln --configuration Release --no-restore` followed by `dotnet test IntentSystem.sln --configuration Release --no-build --no-restore --verbosity minimal`): 5,748 passed, 1 skipped, 0 failed, 5,749 total across all 11 assemblies. The CLI assembly was 5,418/1/0/5,419; the other assemblies totaled 330 passed and 0 skipped/failed.
- `git diff --check`: clean.

## Scope and coordination

Changed paths are exactly:

- `src/IntentSystem.Cli/Commands/NotifySuperviseLivenessCommand.cs`
- `tests/IntentSystem.Cli.Tests/NotifySuperviseCommandTests.cs`
- `docs/en/12-agent-message-orchestration.md`
- `docs/ja/12-agent-message-orchestration.md`

The child `intent-cli guide start` could not find `.intent-cli/config.toml`; worker guidance was then loaded through the installed guide/model/onboarding/commands/automation surfaces. The child next-action reported G769 unheld while the supplied authoritative host claim says execution-unit:G769 is held by implementation/team intent-cli-dev; this stale child/host registry distinction was recorded and no execution-unit claim was created, modified, verified, or released in the child. Issue preflight also returned `missing-target-declaration` because its parser expected a literal `Target paths:` declaration although the complete issue uses `Target Repo / Path / Part`; the standalone issue body was used as the contract.

The canonical issue worker claim proceeded without label mutation because the existing `intent-issue-in-progress` label is stale shadow state. `linked_pr_synced=false` is expected host follow-up.

### DIFF
--- a/docs/en/12-agent-message-orchestration.md
+++ b/docs/en/12-agent-message-orchestration.md
@@ -1019,6 +1019,15 @@ over-bound supervisor even when no supervisor process is alive. The evidence
 state is not a claim that a scheduler job is loaded or that the CLI performed
 registration; registration remains the operator's explicit action.
 
+When `--routing-root <host-root>` is supplied, liveness resolves the supervision
+store beneath that absolute root rather than the caller's current working
+directory, and reports the same absolute root in `routing_root`. The explicit
+JSON `supervision_state` distinguishes `not-found` (the resolved store is not
+present) from `empty-history` (the store exists but has no history) and
+`recorded` (history or corruption evidence is present), so a mistyped root is
+not reported as a dead supervisor. Omitting `--routing-root` preserves the
+existing current-directory resolution and payload behavior.
+
 > **Preview through 1.x (G765).** Persistence is operator-declared and
 > recorded, reconciliation names the keep/remove distinction, and liveness is
 > a read-only observer. No lifecycle execution, automatic re-arming, or
--- a/docs/ja/12-agent-message-orchestration.md
+++ b/docs/ja/12-agent-message-orchestration.md
@@ -857,6 +857,14 @@ orchestration thread が missing / bound 超過の supervisor を報告できま
 CLI が registration を実行したという意味ではなく、registration は operator の明示的な
 action のままです。
 
+`--routing-root <host-root>` を指定した場合、liveness は caller の
+current working directory ではなく、その absolute root 配下の supervision store を読み、
+同じ absolute root を `routing_root` に返します。JSON の `supervision_state` は、解決した
+store が存在しない `not-found`、store は存在するが history が空の `empty-history`、history
+または corruption evidence がある `recorded` を区別します。これにより誤った root を
+dead supervisor として報告しません。`--routing-root` を省略した場合は、従来の current
+directory に基づく解決と payload behavior を維持します。
+
 > **1.x を通じた preview (G765)。** persistence は operator が宣言して記録し、
 > reconcile は keep/remove の違いを明示し、liveness は read-only observer です。
 > lifecycle の実行、automatic re-arming、interval / bound / cycle / shrink / archive /
--- a/src/IntentSystem.Cli/Commands/NotifySuperviseLivenessCommand.cs
+++ b/src/IntentSystem.Cli/Commands/NotifySuperviseLivenessCommand.cs
@@ -44,11 +44,15 @@ internal static class NotifySuperviseLivenessCommand
             return 1;
         }
 
+        string routingRoot;
         string artifactRoot;
+        var hasExplicitRoutingRoot = options.RoutingRoot is not null;
         try
         {
-            _ = Path.GetFullPath(options.RoutingRoot ?? context.RepoRoot);
-            artifactRoot = context.ResolveSupervisionArtifactRootPath();
+            routingRoot = Path.GetFullPath(options.RoutingRoot ?? context.RepoRoot);
+            artifactRoot = hasExplicitRoutingRoot
+                ? ResolveSupervisionArtifactRootPath(context, routingRoot)
+                : context.ResolveSupervisionArtifactRootPath();
         }
         catch (Exception exception) when (exception is ArgumentException or NotSupportedException)
         {
@@ -63,6 +67,9 @@ internal static class NotifySuperviseLivenessCommand
             return 1;
         }
 
+        var supervisionState = hasExplicitRoutingRoot
+            ? DetermineSupervisionState(state)
+            : null;
         var now = (NotifyCommand.UtcNowFactory?.Invoke() ?? DateTimeOffset.UtcNow).ToUniversalTime();
         var lastCycle = state.LastIntervalCycle ?? state.LastCycle;
         var elapsedSeconds = lastCycle is { } cycle
@@ -92,10 +99,11 @@ internal static class NotifySuperviseLivenessCommand
         var result = new NotifySuperviseLivenessResult
         {
             Operation = "supervise-liveness",
-            RoutingRoot = Path.GetFullPath(options.RoutingRoot ?? context.RepoRoot),
+            RoutingRoot = routingRoot,
             Domain = options.Domain,
             Team = options.Team,
             CommandMode = "read-only",
+            SupervisionState = supervisionState,
             LastCompletedCycleAt = lastCycle?.CompletedAt,
             DeclaredBoundSeconds = boundSeconds,
             ElapsedSeconds = elapsedSeconds,
@@ -114,7 +122,9 @@ internal static class NotifySuperviseLivenessCommand
                 boundSeconds,
                 elapsedSeconds,
                 schedulerInstallationEvidence,
-                state.UnreadableRecords),
+                state.UnreadableRecords,
+                supervisionState,
+                state.Directory),
         };
 
         Emit(writer, result, options.Format);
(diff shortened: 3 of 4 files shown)


tomohisa (commented):
APPROVED — reviewed at exact head `84510e30b6fb558851909836285c8900a90ba4ed`.

`design_alignment_checked: true` — checked the G769/#1675 packet and review-context, node 08, the G765/G767/G768 decisions, and both EN/JA orchestration mirrors.

The implementation satisfies the packet contract, not merely its tests. Its scope is exactly the four declared files: liveness resolves an explicit `--routing-root` as the base of the configured store and reports that same resolved root; omitted-root resolution keeps the parent path. The new `supervision_state` is intentionally emitted only on the explicit-root path, preserving G767’s clean no-flag payload oracle. It adds no lifecycle query or write behavior; `commands_executed` remains `none`. `notify status` and `notify supervise shrink` source are unchanged.

- I independently reproduced the parent failure using the G769 test harness against parent CLI assembly `af8b82c37c27ff319c7468084b8ac59590f887fb`: the two fixed-clock, decoy-store cwd payloads differ at the expected cycle timestamp (10:00 versus 11:00), and the missing-root assertion fails because `supervision_state` is absent. At this head the same two tests pass: full payloads are byte-identical across differing cwd contexts, select the configured routing-root store rather than both decoys, and distinguish `not-found` from `empty-history`; the not-found summary does not claim no completed cycle.
- The fixed clock is explicit (`2026-08-30T12:00:00Z`), so equality is not masked by elapsed-time drift. Existing liveness tests inject a throwing process runner and continue to pass, proving the observation-only / no-lifecycle boundary. G767’s unchanged clean-parent oracle and corrupt-reader cases pass; G768’s append and concurrency coverage also passes.
- Verification run: 36/36 passed across `NotifySuperviseCommandTests`, G767 corruption-reader tests, G734 shrink tests, and G768 append/concurrency tests (after building the Release CLI required by one shrink integration test). GitHub’s Build and test and npm dry-run checks are green at this exact head. `git diff --check` is clean. `Closes #1675` matches the linked issue, and archive/shrink/rotation, writer behavior, and liveness conclusions for state actually read remain out of scope.

Residual classification: Design’s supplemental evidence is correct — with no `--routing-root`, liveness remains cwd-dependent, omits `supervision_state`, and the current design-thread residual invocation also omits the flag. That is non-blocking for this PR because AC3 explicitly requires the no-flag parent behavior to remain byte-compatible, whereas AC4 scopes the new distinction to an explicit root. It is a packet-authoring / dependent follow-up boundary, not an implementation finding for #1677.

Issue #1679, opened 2026-08-31
0.27.0: `claim acquire/release --write` fails fatally on Windows — temp-cleanup made fatal & moved pre-commit (regression of #1606/#1617)


hirokogure:
## Summary

On `intent-cli 0.27.0`, `claim acquire --write` (and `claim release --write`) **fail every time** on Windows and no claim is acquired/released. The error is that the claim-transaction temporary directory could not be deleted **before the claim state was committed**.

Because of this, `packet draft` (the G680 claim gate) cannot pass, so the **entire intent pipeline (design / orchestrator / worker) is blocked on Windows**.

**0.25.0 works on the same machine**: there, the same temp-deletion failure is treated as a **non-fatal, post-commit best-effort warning**, and the claim still succeeds. In 0.27.0 the cleanup was moved to **before the commit and made fatal**. This is effectively a **regression back to the 0.24.0 behavior**, which 0.25.0 had already fixed.

## Versions

- Broken: `intent-cli 0.27.0-f43fbd1-G753`
- Working: `intent-cli 0.25.0-74a1c72-G741`
- Reference: `0.24.0-df472fe-G737` (had a similar fatal failure; fixed in 0.25)

## Environment

- OS: Windows 11 Pro (10.0.26200)
- Shell: Git Bash (also reproduces under PowerShell)
- Installed as a .NET global tool (`dotnet tool update --global jtechjapan.intentsystem.cli --version 0.27.0`)
- OS username contains non-ASCII characters (`C:\Users\小暮博幸\...`). **However, the issue also reproduces with an ASCII temp path (see below), so the path is not the root cause.**
- Claims store: same-repo topology (`metadata_write_branch: intent-metadata`)

## Reproduction

```
intent-cli claim acquire --scope execution-unit:AIC-G135 --actor design --team aic --write --format json
```

### Actual output (0.27.0 — FAIL)

```json
{
  "status": "error",
  "scope": "execution-unit:AIC-G135",
  "claim_path": ".intent-cli/claims/4c523b85....json",
  "push_succeeded": false,
  "attempts": 0,
  "detail": "Could not clean up claim transaction temporary directory 'C:\\Users\\<user>\\AppData\\Local\\Temp\\intent-cli-claim-23a2687b3027465eb033633d407ebdfe' before the claim state was committed.",
  "preview_status": "preview-through-1.x"
}
```

[...]

PR #1682, opened 2026-08-31
fix: make claim transaction teardown outcome-tolerant

### LINKED ISSUE #1681: G771: a slow temp-directory delete turns four correct claim answers into a fatal error, and leftovers accumulate forever
# G771: a slow temp-directory delete turns four correct claim answers into a fatal error, and leftovers accumulate forever

## Goal

Make claim-transaction teardown best-effort on **every** path, not only after the commit. A failed delete of the temporary clone must never change a claim command's status, detail, or exit code — and stale transaction roots left by earlier runs must be swept instead of accumulating.

## Why This Slice Exists Now

Consumer report **#1679**: on a Windows 11 host, `claim acquire --write` and `claim release --write` fail every time, no claim is acquired, `packet draft` (the G680 claim gate) cannot pass, and **the entire design / orchestrator / worker pipeline is blocked**. Reported deterministic, 3/3.

## Current Observed State

**Reproduced on macOS — this was never a platform bug.** Clean-room repo with 30,041 files; deleting one transaction clone measured at **1,806 ms against the 250 ms bound**. Three commands then produced the reporter's exact signature:

```
acquire a scope held by a different actor  -> status=error push=False attempts=0  CLEANUP-FATAL
release a scope that is not held           -> status=error push=False attempts=0  CLEANUP-FATAL
release by a non-holder                    -> status=error push=False attempts=0  CLEANUP-FATAL
detail: Could not clean up claim transaction temporary directory '...'
```

Windows merely exceeds the bound far more often — AV scanning, lingering handles, many small files under `.git`.

**Mechanism, confirmed in code.** Cleanup runs in a `finally` **inside** the retry loop. `CleanupTransactionRoot` warns when `committed` is true and **throws** an `IOException` when it is false (≈ line 1293). That exception escapes `RunTransaction` and reaches the handler at ≈ line 119, which constructs exactly `status: "error"`, `push_succeeded: false`, `attempts: 0`, with the cleanup message as `detail`.

**Four early returns exit with `committed = false` and no tolerance flag.** Each is a correct answer the operator needs:

| path | the answer a slow delete destroys |
|---|---|
| acquire, scope held by another actor | who actually holds it |
| release / takeover, no active claim | `not-held` |
| release by a non-holder | "only the attributed holder may release" |
| takeover, wrong `--displaced-holder` | who the current holder is |

**The reported fix would not have worked, and this was measured rather than assumed.** I built `0.25.0-74a1c72-G741` and ran the same three cases in the same clean room: **identical CLEANUP-FATAL results**. The `committed` flag, the throw, and the `finally` placement are the same in both versions; 0.27 *added* `tolerateCleanupFailure` for the identically-held acquire rather than removing tolerance. The reporter's observation that 0.25 worked on their host is real and remains unexplained — but it is not explained by the teardown code, and that open question does not block this unit, because this fix repairs every non-committed path.

**The repository already specifies the intended contract.** `docs/en/09-developer-reference.md` documents the teardown warning verbatim — *"claim transaction committed successfully, but best-effort cleanup could not remove temporary directory … the claim result and exit code are unchanged"* — and #1606 / G738 guaranteed teardown must not fail or hang **after** the claim state is committed. The gap is exactly that scoping: before the commit, teardown may still destroy the command.

**Leftovers accumulate with no sweep.** Measured 3–6 `intent-cli-claim-*` directories left across my runs, each a full clone. Nothing removes them later; the reporter deletes them by hand.

- G738 / #1606: teardown must not fail or hang after the commit. That half already works.
- G747 remote default-branch targeting, G755 reader/writer agreement, G766 fail-closed evidence — all untouched here.
- `deleteDirectory` is already an injectable parameter on `RunTransaction`, so a failing delete is testable without touching the filesystem.

## Target Repo / Path / Part

- Paths: `src/IntentSystem.Cli/Commands/ClaimCommand.cs`, `tests/IntentSystem.Cli.Tests/ClaimCommandTeardownG771Tests.cs`, `docs/en/09-developer-reference.md`, `docs/ja/09-developer-reference.md`
- Part: the throw on the non-committed path in `CleanupTransactionRoot`, the `finally` that hosts it inside the retry loop, and the absent sweep for leftover roots.

## In Scope

- Teardown failure never changing a result, on any path.
- Always warning, naming the leftover path.
- A bounded, best-effort sweep of stale transaction roots on a later run.
- Both doc mirrors.

## Out Of Scope

- Making the 250 ms bound configurable, async deletion, or backoff tuning. Those are the reporter's suggestions 3 and 4; they change *how often* deletion succeeds. This unit makes deletion failure harmless, which is the property that matters.
- Any change to claim semantics, branch targeting, or what the four early returns report.
- Explaining why 0.25 behaved differently on the reporter's host — tracked on #1679.

## Acceptance Criteria

1. Teardown failure **never** changes a claim command's status, detail, or exit code, on any path; a test drives each of the five outcomes — acquired, held-by-another, not-held, release-by-non-holder, takeover-mismatch — with a `deleteDirectory` that always fails, and asserts each result is byte-identical to the same run with a working delete, apart from the emitted warning.
2. That test **fails on the parent commit** for at least the four non-committed outcomes, with the parent's actual output pasted in the PR.
3. A teardown failure is **always reported as a warning** naming the leftover path — a silent leak is not the fix; a test asserts the warning on both the committed and non-committed paths.
4. A genuine pre-commit transaction failure is still reported **with its own cause** and is not replaced or masked by a teardown message; a test injects a failing clone **and** a failing delete together and asserts the reported detail names the clone failure.
5. The documented committed-path warning text is unchanged; a test pins it byte-for-byte against the string in `docs/en/09-developer-reference.md`.
6. Stale transaction roots from earlier runs are swept: a later claim command removes stale `intent-cli-claim-*` directories under the OS temp root, bounded and best-effort, and **never** removes a directory belonging to a live transaction; a test creates a stale leftover plus a live root and asserts only the stale one is removed.
7. The sweep cannot itself fail a command; a test makes the sweep throw and asserts the claim result and exit code are unchanged.
8. **Safety:** no claim is committed, pushed, or reported as owned that was not; a test asserts a failing-delete acquire that never pushed still reports `push_succeeded: false` and does not claim ownership.
9. Both doc mirrors state that teardown is best-effort on every path and that stale roots are swept on a later run.
10. Full Release suite passes with actual counts pasted.

## Verification

- **Criterion 1 is the whole unit.** The oracle is a pair: the same command with a working delete and with a failing delete must produce the same result. Today they differ on four of five outcomes. Paste both for both commits.
- **Criterion 4 is the guard against over-correction.** Swallowing every teardown exception is easy and would also swallow the case where the transaction itself failed — the operator would then see a teardown warning instead of "the clone failed". The two must stay distinguishable.
- **Criterion 6's live-root protection matters more than the sweep.** A sweep that removes a concurrent transaction's workspace is worse than leftovers; this host has run four claim-bearing seats at once all week.

## Related Links

- Consumer report: #1679 (Windows 11, pipeline blocked)
- #1605 — same class on 0.24
- #1606 / G738 — "teardown must not fail or hang **after** the claim state is committed" — the guarantee this unit extends to the other side of the boundary
- #1617 / G743 — post-push receipt verification
- Node: `intents/intent-cli/intent-tree/means/03-state-and-audit-strategy.md`

## Base Branch Policy

tomohisa:
Closes #1681

Consumer report: #1679.

## Summary

- Preserve every claim outcome, detail, ownership field, and exit code when bounded temporary-root cleanup also fails.
- Preserve the original pre-commit transaction cause and report cleanup failure separately.
- Keep the existing committed warning byte-identical, including the leftover OS-temp path.
- Add a bounded stale-root sweep with exclusive lease markers; live or unreadable leases are protected.
- Document the teardown and stale-root contract in both developer-reference mirrors.

## Verification

- Parent `b111fc644dfca24b911c26eef6bad9c784ad6cd4`: `ClaimCommandTeardownG771Tests` failed 9/12 (cleanup exceptions, lost pre-commit cause, no stale sweep, and no live-root protection); the changed tests pass 12/12.
- Focused G771: `Failed: 0, Passed: 12, Skipped: 0, Total: 12`.
- Adjacent G738/G743/G747/G763/G766/G755 claim suites: `Failed: 0, Passed: 38, Skipped: 0, Total: 38`.
- G613: `Failed: 0, Passed: 6, Skipped: 0, Total: 6`.
- Full Release solution: `Failed: 0, Passed: 5762, Skipped: 1, Total: 5763` (NuGet audit disabled for the offline package smoke subprocesses; both package smoke tests also pass individually under that setting).
- `git diff --check`: clean.

The child selector reported G771 holder none/unheld (`wait`, `source_classification=claim-refused`) and child issue preflight reported `missing-target-declaration`/not actionable; host verification was canonical-unavailable because FETCH_HEAD was sandbox-denied. The supplied authoritative host claim on origin/main (claim commit `44bb82cad7396009b1c413cfa496eb6c62c1a592`, claim file `2f363434fe019cb2def68e6e3f49ffd81559247e9e0cdf3bbfeb96a9e7f1a267.json`) was consumed. No child execution-unit claim or host state was mutated.

### DIFF
--- a/docs/en/09-developer-reference.md
+++ b/docs/en/09-developer-reference.md
@@ -2727,6 +2727,23 @@ warning is separate, names the leftover path, and cannot replace that primary
 no-op cause. With `--format json`, stdout contains exactly one JSON document;
 cleanup warnings go to stderr so stdout can be piped directly to a JSON parser.
 
+## Claim transaction teardown on every outcome (G771)
+
+Cleanup is best-effort on every claim outcome. An acquire, release, takeover,
+already-held result, not-held result, holder mismatch, or retry outcome keeps its
+original status, detail, ownership fields, and exit code even when temporary-root
+cleanup also fails. The cleanup warning is separate, names the leftover path, and
+never turns a pre-commit failure into a success or hides its original cause.
+
+Each transaction creates an exclusive lease beside its temporary root. A later
+write command ensures stale transaction roots are swept with a bounded best-effort pass
+(`intent-cli-claim-*`) after a five-minute age grace period. It examines at most
+32 candidates within a 250 ms sweep budget; an active or unreadable lease is
+protected and is never removed. A stale-root delete failure is warning evidence,
+not a claim result or exit-code change. This recovers abandoned roots while
+protecting a live concurrent transaction, and does not alter the existing
+250 ms per-attempt, three-attempt cleanup contract.
+
 ---
 
 ## Version flow
--- a/docs/ja/09-developer-reference.md
+++ b/docs/ja/09-developer-reference.md
@@ -2784,6 +2784,22 @@ active record の actor と team が同じ acquire は意図した no-op です�
 ません。`--format json` の stdout は JSON 文書を 1 つだけ含み、cleanup warning は stderr
 へ出力されるため、stdout を直接 JSON parser に渡せます。
 
+## すべての claim outcome に対する transaction teardown (G771)
+
+Cleanup はすべての claim outcome で best-effort です。acquire、release、takeover、
+already-held、not-held、holder mismatch、retry の各 result は、一時 root の cleanup
+も失敗した場合でも元の status、detail、ownership fields、exit code を保持します。
+cleanup warning は別の evidence として leftover path を示し、pre-commit failure を
+success に変えず、元の原因も隠しません。
+
+各 transaction は一時 root の横に exclusive lease を作ります。後続の write command は
+5 分の age grace period 後に、`intent-cli-claim-*` に一致する stale transaction root を
+bounded に best-effort sweep します。候補は最大 32 件、sweep budget は 250 ms です。
+active または読めない lease は保護し、削除しません。stale transaction root の delete
+failure は warning evidence であり、claim result や exit code を変更しません。これにより
+放置された root を回収しながら live な concurrent transaction を保護し、既存の 250 ms
+per-attempt、3 回の cleanup contract は変更しません。
+
 ---
 
 ## バージョンフロー
--- a/src/IntentSystem.Cli/Commands/ClaimCommand.cs
+++ b/src/IntentSystem.Cli/Commands/ClaimCommand.cs
@@ -19,6 +19,11 @@ internal static class ClaimCommand
     internal const int CleanupMaxAttempts = 3;
     internal static readonly TimeSpan CleanupAttemptTimeout = TimeSpan.FromMilliseconds(250);
     internal static readonly TimeSpan CleanupRetryDelay = TimeSpan.FromMilliseconds(50);
+    private const string TransactionRootPrefix = "intent-cli-claim-";
+    private const string TransactionLeaseSuffix = ".lease";
+    private const int StaleTransactionRootSweepMaxCandidates = 32;
+    private static readonly TimeSpan StaleTransactionRootMinimumAge = TimeSpan.FromMinutes(5);
+    private static readonly TimeSpan StaleTransactionRootSweepBudget = TimeSpan.FromMilliseconds(250);
 
     private static readonly JsonSerializerOptions JsonOptions = new()
     {
@@ -149,6 +154,8 @@ internal static class ClaimCommand
                 null, null, null,
                 "Dry-run only. Re-run with --write; ownership exists only after a successful plain push.");
         }
+        SweepStaleTransactionRoots(warningWriter, deleteDirectory);
+
 
         var targetRef = $"refs/heads/{defaultBranch}";
 
@@ -161,7 +168,8 @@ internal static class ClaimCommand
         for (var attempt = 1; attempt <= request.MaxAttempts; attempt++)
         {
             var transactionRoot = Path.Combine(
-                Path.GetTempPath(), $"intent-cli-claim-{Guid.NewGuid():N}");
+                Path.GetTempPath(), $"{TransactionRootPrefix}{Guid.NewGuid():N}");
+            using var transactionLease = ClaimTransactionLease.Create(transactionRoot);
             var committed = false;
             var tolerateCleanupFailure = false;
             Exception? transactionFailure = null;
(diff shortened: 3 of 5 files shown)


tomohisa (commented):
APPROVED — reviewed at exact head `31db94abdf4025a9c12c705d342e210fde88c997`.

`design_alignment_checked: true` — checked the G771/#1681 packet and review-context, node 03, G738/G743/G747/G755/G763/G766 decisions, both EN/JA developer-reference mirrors, and consumer report #1679.

- I independently ran the G771 paired oracle against parent CLI `b111fc644dfca24b911c26eef6bad9c784ad6cd4`. Nine cases failed: cleanup replaced each of the four non-committed outcomes; clone failure was not accompanied by the required preserved-cause cleanup warning; stale sweep did not run; and the live-root protection case failed. This reproduces the result-masking defect rather than relying on the PR summary.
- AC1–AC4: the head’s paired test covers acquired, held-by-another, not-held, release-by-non-holder, and takeover-mismatch with working versus failing deletion, preserving status, detail, ownership, target, push fact, and entry-point exit code. Warnings always name the leftover. A deliberately failing clone plus failing cleanup retains the clone cause, never the cleanup cause. The dedicated G771 plus existing G738 regression set passes 13/13.
- AC5–AC8: the committed warning is pinned byte-for-byte against the EN text and present in JA; a later write sweeps only stale `intent-cli-claim-*` roots under a 32-candidate/250-ms bound and protects locked or unreadable lease markers. Sweep failures are warnings only. Held/failing-delete acquisition remains `held`, `push_succeeded: false`, and names the actual holder — no false ownership, push success, or receipt is produced.
- The implementation retains the 250-ms × 3 cleanup and 50-ms retry-delay constants; no timeout/backoff tuning, claim semantic, default-branch targeting, push-receipt, reader/writer, or claim-schema change was introduced. G743/G747/G755/G763/G766 claim suites remain green, as do G613. Design’s 30,041-file / about-1.8-s cleanup and stale/live host observations are compatible supporting evidence, not a substitute for the local parent oracle.
- Tests run: G771 + G738 13/13; G743/G747/G755/G766/G613 23/23; G763/G679 21/21. `git diff --check` is clean and both GitHub checks are green at this exact head. `Closes #1681` correctly links the consumer-driven unit; EN/JA state every-outcome best effort and bounded stale sweeping. Tests are necessary but not sufficient; these packet, consumer, and node-03 boundary checks support the approval.

PR #1686, opened 2026-09-01
G773: repair unreadable supervision history records

### LINKED ISSUE #1685: G773: quarantine unreadable supervision records so permanent damage stops being reported forever
# G773: quarantine unreadable supervision records so permanent damage stops being reported forever

## Goal

Add `intent-cli notify supervise repair-unreadable` — a sanctioned, non-destructive exit for supervision-history records that can no longer be parsed. Today G767 makes every reader degrade per record and *name* the damage, and G768 stops *new* damage from being written; but the damage that already exists is append-only and hand-editing host state is forbidden, so it is reported on every read, forever. After this slice an operator can move those lines, verbatim, into a quarantine sidecar with an audit trail, and the store reads clean.

## Why This Slice Exists Now

Measured 2026-09-01 with the **released** tool `intent-cli 0.28.0-bb7e56d-G772` against the real host that motivated the G765–G770 chain:

```
$ intent-cli notify supervise liveness --domain intent-cli --team intent-cli-dev \
    --routing-root <host-root> --format json
# supervision_state: recorded
# unreadable_records: 9 entries, all cycles.jsonl / invalid-json,
#   lines 9803, 9804, 57687, 57688, 150639, 157887, 183811, 278778, 618297
```

The store is 718,903,202 bytes / 663,959 lines. The 9 lines are from 5 pre-G768 write interleavings; nothing can ever make them parse. The v0.28.0 release notes state plainly — and a test guards the wording — that **G768 stops new corruption and does not repair existing damage**. This unit is that missing repair, and every consumer host that corrupted a store before 0.28.0 needs it, not just this one.

## Current Observed State

Measured 2026-09-01 with the released tool `intent-cli 0.28.0-bb7e56d-G772` on the real host: liveness succeeds and reports the 9 unreadable records shown above; they have been reported on every read since G767 landed and will be reported forever without this unit. The store is append-only; hand-editing host state is forbidden; no sanctioned command removes an unreadable line today. `notify supervise repair-cycle-history` untracks history from git (G750) and does not touch damaged lines; `shrink`/`archive` (G744) manage volume, not damage.

## Design (decided; follow G763's shape)

`claim stranded` / `stranded migrate` is the precedent: read-only report, `--dry-run` that lists exactly what would move, `--write` that never loses information, post-write verification.

- For each of `cycles.jsonl` / `stalls.jsonl` / prompt-audit files: every unreadable line moves **verbatim, in order,** to a per-file quarantine sidecar (e.g. `cycles.unreadable.jsonl`) in the same directory; the live file is rewritten via temp-file-plus-rename to exactly the readable lines. **No byte of any line is discarded anywhere.**
- Each run appends one audit record (per-file counts, line numbers, timestamp, writer identity) via the G768 atomic appender to a repair audit log.
- Runs under the directory lock **and** re-verifies the target file's length against the scan immediately before rename — G768 proved the lock alone is not a defence, so the repair fails closed on any concurrent growth.
- A store with nothing to repair is a strict no-op in both modes: no rewrite, and a report that prints differently from a completed repair.
- Quarantining *every* line of a file leaves a real store with **empty history** — which must read as `empty-history`, never as `not-found` and never as health (G769/G770's discriminating pair stays intact).

**Considered and rejected:** an acknowledgment ledger that leaves damaged lines in place and teaches readers to skip them. It makes every read of a 719MB file pay the parse-failure cost forever, leaves the file lying about its own contents, and reintroduces the silent-skip failure G767's criteria forbid.

**To verify, not assume:** `shrink`/`archive` (G744) read the same directory. If archive already moves unreadable lines somewhere, present the evidence and reconcile via request-update instead of double-handling.

- G767: readers degrade per record and name damage (`unreadable_records` in liveness). Reuse its per-line tolerance for the scan; do not invent a second tolerance model.
- G768: `AtomicAppendWriter` exists (real `O_APPEND` / `FILE_APPEND_DATA`) and the directory lock exists — and the lock alone is proven insufficient against a writer that predates it.
- G769/G770: `--routing-root` is honored and `not-found` vs `empty-history` is a guarded discriminating pair in liveness.
- G763: the report → dry-run → non-destructive write → post-write verification shape is settled precedent.
- v0.28.0 release notes pin, with a test, that G768 "does not repair existing damage" — this unit is that repair and must not weaken that wording; it supersedes it with a new command.

## Target Repo / Path / Part

- Paths: `src/IntentSystem.Cli/Commands/NotifySupervisionStore.cs`, `src/IntentSystem.Cli/Commands/NotifySuperviseCommand.cs` (route registration), new `src/IntentSystem.Cli/Commands/NotifySuperviseRepairUnreadableCommand.cs`, `tests/IntentSystem.Cli.Tests/`, `docs/en/12-agent-message-orchestration.md`, `docs/ja/12-agent-message-orchestration.md`
- Part: a new `notify supervise repair-unreadable` route: verbatim quarantine of unreadable JSONL lines into per-file sidecars, temp+rename rewrite of live files, under the directory lock with pre-rename length re-verification, plus an atomic audit record.

## In Scope

- The new route, its dry-run/write modes, payloads, and fail-closed race handling.
- Quarantine sidecars and the repair audit log.
- Tests for every acceptance criterion, including the byte-conservation and no-op-on-clean pair.
- EN/JA §12 documentation of the command, guarantee, and limitation.

## Out Of Scope

- Reconstructing damaged records' contents.
- Any behaviour change for liveness/shrink/archive/readers on an untouched store.
- Repair-on-read or repair from any other command path.
- G744 archive window semantics; G750 git-tracking semantics.

## Acceptance Criteria

- [ ] New route `notify supervise repair-unreadable --domain <d> --team <t> [--routing-root <host-root>] [--dry-run|--write] [--format markdown|json]`; `--dry-run` default; dry-run lists every unreadable line (file, line number, reason, byte length) and mutates nothing — a test asserts the store directory is byte-identical after dry-run.
- [ ] `--write`: a test asserts (i) sidecar's added bytes equal the removed lines byte-for-byte in order, (ii) the rewritten live file equals the original minus exactly those lines, (iii) a subsequent liveness run reports zero unreadable records.
- [ ] Clean store is a strict no-op in both modes — no rewrite, bytes and file identity unchanged; a test pins that the nothing-to-repair report and a completed-repair report print differently.
- [ ] All-corrupt file: `--write` quarantines every line and liveness then reports `empty-history` — not `not-found`, not health; end-to-end test.
- [ ] Race safety: directory lock + length re-verification immediately before rename; on change, fail closed naming the real cause, no live file replaced; a test simulates the race.
- [ ] Scope: no route other than `repair-unreadable` gains any write to history files; existing liveness fixtures re-run unchanged.
- [ ] One audit record per run via the atomic appender; test asserts it exists and parses.
- [ ] Each new test demonstrated failing or absent on the parent commit, actual output pasted.
- [ ] Both doc mirrors (EN/JA §12) document the command, the verbatim-quarantine guarantee, and the limitation: it removes unreadable lines from the read path and preserves them as evidence; it does not reconstruct what the damaged records once said.
- [ ] Full Release suite passes with actual counts pasted.

**Request-update line:** criteria 2 and 3 together. Filtering at read time instead of rewriting, rewriting unconditionally even when clean, or losing a single byte between live file and sidecar fails this packet even if every other test is green.

## Verification

Build the branch head in Release and run against fixture directories, not a live host:

- **Criteria 2+3 pair (the request-update line)**: damaged fixture → write → (a) sidecar bytes equal removed lines, (b) live file equals original minus those lines, (c) liveness reports zero unreadable; then run write again → `nothing-to-repair` with live-file byte/identity unchanged. A clean fixture must never be rewritten in either mode.
- **All-corrupt pair**: after write, liveness must print `empty-history`, and a never-existed store must still print `not-found`.
- **Race fixture**: append to the live file after the scan; the command must fail closed naming the cause, original file untouched.
- Paste the parent-commit failure/absence output for every new test.

## Related Links

- Release stating the limitation this unit removes: https://github.com/J-Tech-Japan/intent-system/releases/tag/v0.28.0
- G767 (degrade-and-name readers): https://github.com/J-Tech-Japan/intent-system/pull/1673
- G768 (atomic appends; stops new damage only): https://github.com/J-Tech-Japan/intent-system/pull/1676
- G763 (the stranded/migrate precedent this follows): https://github.com/J-Tech-Japan/intent-system/pull/1667
- G769/G770 (the not-found vs empty-history pair to preserve): https://github.com/J-Tech-Japan/intent-system/pull/1677 / https://github.com/J-Tech-Japan/intent-system/pull/1680

## Base Branch Policy

tomohisa:
## Summary
- Adds the explicit `notify supervise repair-unreadable` dry-run/write route.
- Quarantines unreadable JSONL bytes verbatim, rewrites only readable live bytes, records an atomic audit, and fails closed on concurrent growth.
- Documents the evidence-preserving/no-reconstruction boundary in both orchestration mirrors.

## Verification
- `dotnet test tests/IntentSystem.Cli.Tests/IntentSystem.Cli.Tests.csproj --filter FullyQualifiedName~NotifySupervisionRepairUnreadableG773Tests --no-restore --nologo`
- `dotnet test IntentSystem.sln --configuration Release --no-restore --nologo` (5785 passed, 1 skipped, 0 failed)
- `git diff --check HEAD^ HEAD`

Closes #1685

### DIFF
--- a/docs/en/12-agent-message-orchestration.md
+++ b/docs/en/12-agent-message-orchestration.md
@@ -1067,6 +1067,42 @@ command. `stalls.jsonl` and prompt-audit records receive the same per-record
 treatment as cycles, so a damaged secondary history cannot silently erase the
 rest of the team's supervision answer.
 
+### Evidence-preserving unreadable-record repair (G773 — preview-through-1.x)
+
+G773 adds the one explicit write path for damage already named by the G767
+reader:
+
+```bash
+intent-cli notify supervise repair-unreadable --domain <domain> --team <team> \
+  [--routing-root <host-root>] [--dry-run|--write] [--format markdown|json]
+```
+
+`--dry-run` is the default. It lists every unreadable record with its relative
+file, one-based line number, G767 reason, and source-line byte length, while
+leaving the whole store byte-identical. A clean store is a strict no-op in
+both modes: no live file, quarantine sidecar, or audit is created or replaced,
+and the result is explicitly `nothing-to-repair` rather than a completed
+repair.
+
+With `--write`, the command takes the supervision directory lock, rescans,
+and checks each target live-file length immediately before its atomic rename.
+It copies the exact unreadable byte ranges **verbatim and in order** to a
+per-file `.unreadable.jsonl` quarantine sidecar in the same directory, then
+replaces the live file with exactly its original readable bytes. Before a
+live replacement, concurrent growth makes the operation fail closed with the
+real changed-file cause; the changed live file is not replaced. One parsable
+G768 atomic audit record is appended for each completed write run with the
+file counts, line numbers, timestamp, and writer identity.
+
+Quarantine sidecars are evidence, not history: readers do not return them to
+the live/archive read set. Thus a repaired all-corrupt existing file becomes
+`empty-history` rather than `not-found` or health, and subsequent liveness
+reports zero unreadable records. Archive history follows the same per-file
+quarantine rule, so repair neither changes G744's archive window nor asks
+shrink/archive to repair corruption. This command preserves evidence; it does
+**not reconstruct** what a damaged record once said, and it never repairs on
+read.
+
 ### Atomic per-record supervision history appends (G768 — preview-through-1.x)
 
 Supervision history writes encode each cycle, stall, and prompt-audit JSONL
--- a/docs/ja/12-agent-message-orchestration.md
+++ b/docs/ja/12-agent-message-orchestration.md
@@ -899,6 +899,36 @@ reader は read-only のままで、supervisor や OS lifecycle command を実�
 `stalls.jsonl` と prompt-audit record にも cycles と同じ record 単位の扱いを適用するため、
 secondary history の damage が team の supervision answer の残りを黙って消すことはありません。
 
+### unreadable record を evidence として隔離する repair (G773 — preview-through-1.x)
+
+G773 は、G767 reader がすでに名前付きで示した既存 damage に対する唯一の明示的 write path を
+追加します。
+
+```bash
+intent-cli notify supervise repair-unreadable --domain <domain> --team <team> \
+  [--routing-root <host-root>] [--dry-run|--write] [--format markdown|json]
+```
+
+default は `--dry-run` です。relative file、1-based line number、G767 の reason、source line の
+byte length をすべて列挙し、store 全体を byte-identical のまま残します。clean store は両 mode で
+strict no-op です。live file、quarantine sidecar、audit の作成・置換はなく、結果は completed repair
+ではなく明示的に `nothing-to-repair` になります。
+
+`--write` では supervision directory lock を取得して再走査し、各 target live file を atomic rename
+する直前に length を再確認します。読めない byte range は同じ directory の file ごとの
+`.unreadable.jsonl` quarantine sidecar へ、順序を保った **verbatim** の bytes としてコピーし、その後
+live file を元の readable bytes だけに置換します。live replacement の直前に concurrent growth を
+検出した場合は、実際に変化した file を示して安全に失敗し、その live file は置換しません。completed
+write run ごとに、file count、line number、timestamp、writer identity を含む 1 つの parsable な
+G768 atomic audit record を追記します。
+
+quarantine sidecar は history ではなく evidence です。reader は live/archive read set に戻しません。
+したがって existing all-corrupt file は repair 後に `not-found` や health ではなく `empty-history` となり、
+次の liveness は unreadable record を 0 件と報告します。archive history にも同じ file 単位の
+quarantine rule を適用するため、G744 の archive window を変更せず、shrink/archive に corruption repair を
+させません。この command は evidence を保存しますが、damaged record がかつて何を述べたかを
+**reconstruct** しません。また read 時に repair することもありません。
+
 ### supervision history の record 単位 append (G768 — preview-through-1.x)
 
 supervision history の cycle、stall、prompt-audit は、1 つの UTF-8 append 操作で
--- a/src/IntentSystem.Cli/Commands/NotifyCommand.cs
+++ b/src/IntentSystem.Cli/Commands/NotifyCommand.cs
@@ -85,6 +85,7 @@ internal static class NotifyCommand
         + NotifySuperviseShrinkCommand.Usage + "\n"
         + NotifySuperviseArchiveCommand.Usage + "\n"
         + NotifySuperviseRepairCycleHistoryCommand.Usage + "\n"
+        + NotifySuperviseRepairUnreadableCommand.Usage + "\n"
         + NotifySuperviseInstallCommand.Usage + "\n"
         + NotifySuperviseReconcileCommand.Usage;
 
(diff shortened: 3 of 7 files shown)


tomohisa (commented):
## Verdict: REQUEST-UPDATE

Reviewed exact head `d9e07e35fb9b7abd7ac87a9611bd555173121366`.

`design_alignment_checked=true`: I checked G773's supplied packet/review-context contract, issue #1685, nodes 08/03, the G744/G765/G767--G770/G768 boundaries, and both EN/JA documentation. Tests are necessary but not sufficient.

The implemented parts that I independently verified are sound: the parent at `bb7e56dd12fbf0e71421d59d8249cb0f7334a996` rejects `repair-unreadable` as an unknown argument; dry-run enumerates the G767 reason/file/line/byte length without mutation; write retains readable bytes and quarantines corrupt bytes; all-corrupt becomes `empty-history` rather than `not-found`; the existing G768 lock plus immediate length check detects the exercised concurrent append; archive reading excludes `.unreadable.jsonl` sidecars; and the liveness change only exposes the shared root resolver. The strict G768/G771 truthfulness boundary and both documentation mirrors remain intact. I performed no real-host `--write`.

One blocking implementation finding remains (issue #1685 AC2/AC7: exact evidence conservation and one honest audit record per completed repair):

`RepairUnreadable` publishes a plan's quarantine replacement at `NotifySupervisionStore.cs:1223-1225`, then performs the live-file replacement at `:1240`, and only appends the audit after all replacements at `:1271`. If the live replacement throws after the sidecar succeeds (or a later plan fails), the catch returns a failed result with no audit/transaction record, while the original corrupt live bytes remain. A retry then scans that same corrupt line and appends its bytes to the already-mutated sidecar again. The eventual quarantine therefore contains a duplicate of evidence that existed only once in the source, rather than the required byte-for-byte move, and the partial mutation has no recovery/audit record.

The present race test does not cover this: its hook appends before the first length recheck, so it returns before any quarantine is created. Add a failure injection after quarantine publication and before the live rename (and cover later-file/audit failure as applicable), make the transfer recoverable or idempotent, and prove that failure leaves no byte lost, no duplicated quarantine evidence on retry, no false completed audit, and exactly one parsable audit for the eventual completed transfer.

Local checks passed: G773 focused 9/9; adjacent G744/G767/G768/shrink/archive/command compatibility 43/43; G613 6/6. Exact-head CI run 33467445025 is green and `git diff --check` is clean, but neither covers the required post-quarantine failure path.

tomohisa (commented):
## Verdict: APPROVED

Re-reviewed exact head `8830612d8f294ec99b16de8d51f9b68ec58bd17d`.

`design_alignment_checked=true`: I checked the G773 packet/review-context requirements supplied with this re-review, issue #1685, Design ruling `G773-design-ruling-1686-001`, nodes 08/03, and the EN/JA documentation. Passing tests are necessary, not sufficient; the prior AC2/AC7 failure boundary was independently examined in the transaction implementation and exercised through its fault injections.

The prior blocker is repaired. A repair now stages hash-verified replacements and persists a `repair-unreadable` transaction before publishing files. The `AfterQuarantinePublication` injection runs after sidecar publication and before its corresponding live replacement. Its single-file test proves: the original live bytes remain after the failed attempt, quarantine contains the exact unreadable raw bytes, no completed audit exists, and the retry produces readable live bytes, exactly one copy of the quarantined bytes, and exactly one parsable audit. The later-file injection proves the same convergence over cycles plus stalls; the audit failure injection proves a completed live/quarantine transfer has no completed audit until retry appends exactly one. Transaction-id audit detection prevents a late atomic-append outcome from being duplicated.

The clean-store dry-run/write byte-and-file-identity no-op and distinct pre-recheck growth-race remain intact. All-corrupt still becomes `empty-history`; liveness, shrink, archive, G767 per-record reader behavior, G768 atomic append behavior, and G744 archive compatibility remain unchanged. The re-review delta is limited to `NotifySupervisionStore.cs` and the G773 test file; the full PR scope remains the approved route/store/tests/docs boundary, with no real-host `--write`, release, tag, or publication action. `git diff --check` is clean and exact-head CI run 33484696853 is green.

Local verification passed: G773 focused 12/12, adjacent G744/G767/G768/shrink/archive/command compatibility 43/43, and G613 6/6.

PR #1779, opened 2026-09-11
G731: probe reader writability empirically instead of inferring sandbox from path shape

tomohisa:
## Summary
- Replace the structural routing-root/report-root comparison in `notify report` with an empirical write probe against the exact external reader file (`ProbeExternalReaderAppendWritability`). The `report-routing-root-write-required` refusal now fires only when the probe is actually denied, so a writable host completes the append and marks the sender-local outbox delivered in place.
- Name the recovery action: `notify reconcile` and probe-denied summaries emit the concrete `intent-cli notify collect` command; `collect` on a writable host appends the reader event, marks the outbox delivered, and records the continuation chain while leaving pending state to reconcile.
- Add `NotifyReportOutboxStore.FindEntriesByDeliveryError` to surface probe-denied outbox entries for recovery.
- Update en/ja orchestration docs: refusals are measured against the write, never inferred from the shape of the roots.

## Verification
- `dotnet build src/IntentSystem.Cli/IntentSystem.Cli.csproj -c Release` — 0 warnings, 0 errors
- `dotnet test tests/IntentSystem.Cli.Tests/IntentSystem.Cli.Tests.csproj -c Release` — 5185 passed, 1 pre-existing skip, 0 failed
- Deny path retains coverage via `ExternalReaderRetainsSenderLocalHandoffWhenHostRootIsDenied_G719` (chmod-based host denial still refuses before append and names collect as recovery)

Refs G731


tomohisa:
Closing: this PR was produced by a local-LLM evaluation run (Qwen3.8 Flash Next on oMLX 0.7.0.dev2, thinking disabled), not intended for merge. It exists so the measurement can be inspected. Notes are in the write-up linked from https://zenn.dev/jtechjapan_pub/articles/qwen-flash-next-runtime-comparison

Issue #1771, opened 2026-09-07
Bootstrap: create and record an Orca Run mailbox for an external design resident (cross-project mail has no durable home today)


tomohisa:
## Request from the sekiban-dcb-ts design seat (operator-directed)

The operator asked me to check whether intent-cli's onboarding for an **external design resident** already sets up an Orca Run mailbox for the seat, and if not, to ask for it to be packetized.

### What happened

- The sekiban-dcb-ts design seat runs as an external resident (Claude Code) in team `sekiban-dcb-ts-orch`. Its intent-cli topology records only a reader file; it had **no Orca Run**.
- Other design threads on this machine each have one (`run_f804e068674b` SekibanCloud, `run_fac6298d9140` Sekiban, `run_ba04941cb504` SekibanWasmRuntime, `run_06a53b53e187` remote-herdr, `run_6c6a87da8fbb` intent-cli).
- Cross-project mail from other designs therefore arrived at a **terminal handle**, which leaves no durable record on Orca (SekibanCloud pointed this out on 2026-09-04). A review-seat consultation report also failed canonical delivery because that seat's sandbox cannot write the Host event path, and only reconciliation from the report root recovered it.
- On 2026-09-07 I created `run_585adfc0e774` by hand with `orca orchestration run-create --objective "..."`, posted the seat's status there, announced it to the three other design Runs, and then had to **wake each design terminal** with `orca terminal send` because a Run mailbox is passive: nobody notices mail until they run `check`.

Neither `intent-cli guide bootstrap` nor `intent-cli guide onboarding` mentions Orca, `run-create`, or a mailbox today (checked 2026-09-07).

### Ask

1. Is an Orca Run mailbox for an external design resident already planned as part of bootstrap / onboarding / topology record? If yes, point me at the unit.
2. If not, please packetize it. Suggested shape, for the architect to correct:
   - **Bootstrap step**: when the design role is an external resident and Orca is present, run `orca orchestration run-create --objective "<domain> design thread — ..."` once, and **record the run id in the topology record** (e.g. `roles.design.orca_run_id`) and in the handover document, so `session-layer topology show` and the next design seat both know the mailbox.
   - **Cross-team addressing**: a read-only way to list the Run ids of other design threads on the host (today one must read `orca orchestration run-list` and match by prose).
[...]

PR #1821, opened 2026-09-16
G837: record, show and validate an adopted Orca Run mailbox on the team-shape seat (no orca execution)

### LINKED ISSUE #1820: G837: record, show and validate an adopted Orca Run mailbox on the team-shape seat (no orca execution)
# G837: record, show and validate an adopted Orca Run mailbox on the team-shape seat (no orca execution)

## Goal

Some design and Steward seats already receive cross-team mail on an Orca Run that someone created by hand. intent-cli does not record which Run that is, so the next seat and other design threads have to find it by reading prose. Let intent-cli record one exact, already-adopted Orca Run id per team on the seat that the recorded team shape selects. intent-cli then shows it, validates it, lists it read-only across the host, and includes it in the bootstrap handover. It also records a receive-policy label and checks it against the seat's recorded frontend, but never acts on it. intent-cli never runs `orca`, never creates a Run, and never sends, checks, or schedules anything.

The same unit fixes two texts that are wrong today: the G822 refusal promises that #1774 delivers dispatch identity, and the design-thread Orca block says intent-cli adds no Orca option. It also makes the notify `event-append-failed` failure name its reconciliation command when the report root is the routing root.

## Why This Slice Exists Now

- Operator-filed #1771 (2026-09-07) reports the following:
  - The sekiban-dcb-ts external design seat had no Orca Run. The operator created `run_585adfc0e774` by hand.
  - Other design threads each use a Run, but no record in intent-cli links a team to its Run.
  - A review seat's report failed canonical delivery because its sandbox could not write the host event path. Reconciliation from the report root recovered it, but the failure text did not name that route.
- #1774 (G813) is retired as `decomposed` (PR #1781 shipped only team identity). The remainder is re-scoped here, against current main.
- Operator decisions, 2026-09-15 (`reports/G813-orca-mailbox-design-20260907.md`, final section):
  - intent-cli records and validates only.
  - The seat to bind depends on team shape: a five-seat team binds the steward, a four-seat team binds design, and `solo-conductor` binds design.
  - Dispatch identity is split out to a later unit.
- Operator answers to this packet's design questions (2026-09-15):
  - A team with no team-mode record is refused.
  - The frontend list is closed: `orca` means `orca-push`, `claude-app` or `codex-app` means `inbox-pull`, and anything else is refused.
  - The handover surface is `guide bootstrap`.
  - A missing binding is informational only.
  - Recording a binding on the live `intent-cli-dev` team is outside this unit.
- The preview.3 scope amendment (2026-09-14) says:
  - Design and Steward seats use Orca Run mailboxes.
  - App-based design seats pull on a scheduled wake, and Orca is the push route.

## Current Observed State

On child main `75d68523`:

- **Role record.** `NotifyRecordedRole` (`src/IntentSystem.Cli/Commands/NotifyRoleTopology.cs:6-24`) has `Resident`, `WorkspaceId`, `PaneId`, `Reader`, `Cwd`, `Kind`, `DeliveryMethod`, `Frontend`, `LaunchArguments`, envelope profile fields, `Model`, `ReasoningEffort`, and `WakeCommand`. No field holds an Orca Run.
  - `Resolve` builds roles at `NotifyRoleTopology.cs:484-498`.
  - It refuses a topology with no roles: `contains no roles` at `:505`, and in `Validate`, `contains no recorded roles` at `:907`.
  - `wake_command` is accepted only on an external resident (`:449-455`).
- **Frontend is external-only.**
  - `update-field --field frontend` refuses any resident other than `external` (`SessionLayerTopologyCommand.cs:2474-2481`).
  - `record` refuses `--frontend` for a herdr resident (`:1110-1114`).
  - `update-residence` to herdr removes `frontend` (`:2603`).

  A herdr seat therefore can never record a frontend.
- **Flat topology layout.** In a flat layout, every team-level property that is not in the envelope list is read as a role:
  - reader side: `IsTopologyEnvelopeProperty`, `NotifyRoleTopology.cs:1320-1322`;
  - writer side: `IsEnvelopeProperty`, `SessionLayerTopologyCommand.cs:2984-2986`, and `containsDirectRoles`, `:2828-2829`.

  A new team-level object would therefore be misread as a role.
- **Multi-team files.** `TrySelectTeam` also accepts a `teams` object keyed by team name (`NotifyRoleTopology.cs:1041-1051`). The canonical path is still one file per team (`:234-240`).
- **Topology writers and locks.**
  - `move` (`:1952-1955`) and `update-residence` (`:2530-2532`, digest re-check `:2651-2656`) take `AcquireCasLock` (`:2960-2964`, file `<topology>.json.lock`) under `--write`. They create the directory, then the lock, then call `EnsureLocalIgnore`, before any refusal check. A refusal can therefore leave the lock and ignore files behind.
  - `record` (`:1784`), `record-host-state` (`:1917`), `record-profile` (`:2338`), `update-kind` (`:2425`), and `update-field` (`:2505`) write with `WriteAtomically` and no lock.
  - `WriteAtomically` writes a `.<file>.<guid>.tmp` sibling and moves it into place (`:2988-3004`). It serializes with `FileJsonOptions` (`:1636`), which uses the default encoder.
  - `EnsureLocalIgnore` writes `.intent-cli/topology/.gitignore` containing `*` (`:3007-3017`, path `NotifyRoleTopology.cs:243-245`).
  - `record-host-state` (`:1814-1920`) is a dedicated declaration command: an idempotent no-op on an exact match, and a refusal on a silent replace.
  - `record` compares an existing role only on `KnownRoleFields` (`:1638-1642`).
- **`topology update-field`** (`SessionLayerTopologyCommand.cs:2442-2513`).
  - It has an allow-list of `delivery_method`, `frontend`, and `wake_command` (`:2445-2447`), and sets one string per call.
  - It is guarded only by the stated `--current` value (`:2482-2491`).
  - It needs a valid topology and an already recorded role (`:2460-2472`).
  - Its result has no named cause (`:3290`).
- **Team-mode gate for topology.** Before subcommand dispatch (`:116-130`), `Execute` resolves team mode for any `session-layer topology` call that carries `--domain` (`:97-114`). The resolve uses the `--team` value, or none.
  - `authoring-only` returns `not-applicable-team-mode` with exit 1 (`EmitNotApplicable`, `:1405-1439`). This happens before argument parsing.
  - A thrown `InvalidOperationException` prints `team-mode-unreadable: …` with exit 1. That includes `TeamModeResolutionException` for an omitted `--team` in a domain with several team-scoped entries and no domain-wide entry (`TeamMode.cs:227-244`).
- **`topology show` / `topology validate`.**
  - `ExecuteShow` (`:212-395`) renders `SessionLayerTopologyShownRole` (`:3326-3341`). The JSON options are snake_case with `WhenWritingNull` (`:68-73`).
  - `ExecuteValidate` (`:133-210`) adds reader-path findings in the command layer only when the store validation is valid (`:174-185`; `:170-173` is the `--live` preflight). `valid` is true only when every finding is informational.
  - The precedent for an informational finding is `host-state-role-missing` (`NotifyRoleTopology.cs:955-966`).
- **Team mode (G691/G833).**
  - `.intent-cli/team-mode.json` holds `schema_version` and `entries[]` of `domain`, `team?`, `mode`, `updated_at`, `transitions` (`TeamMode.cs:104-129`).
  - `TryRead` deserializes into these typed records and fails closed on any invalid entry (`:152-205`, `Validate` `:282-343`).
  - Absence resolves to `delivery` with `Source = Default` (`:216-218`, `:258-259`). A team request falls back to a domain-wide entry (`:252-255`).
  - The internal `Resolve(TeamModeState?, domain, team)` never throws when a team is given (`:213-267`).
  - `team-mode set` reads, rebuilds every entry from typed properties, and writes (`TeamModeCommand.cs:100-147`). It writes through `TeamModeStore.Write` (`TeamMode.cs:275-280`), a plain `File.WriteAllText` with no lock and no digest check.
  - `solo-conductor` completion is measured from the recorded mode alone (`GuideBootstrapCommand.cs:379`, `:389-409`).
- **Tracking on hosts** (measured on this operator host). `.intent-cli/team-mode.json` is git-tracked, while `.intent-cli/topology/**` is ignored by `.intent-cli/topology/.gitignore` (`*`). One existing lock sits outside an ignored directory, `QueueReprioritizeCommand.cs:508` → `.intent-cli/queue-state.reprioritize.lock`, and that file is committed on this host.
- **Roles.** `LogicalRoleNormalizer` has canonical `architect`, `orchestrator`, `builder`, `reviewer`, and `steward`, with aliases `design`, `orchestration`, `implementation`, and `review` (`LogicalRoleNormalizer.cs:11-38`).
- **`guide design-thread` Orca text.** `GuideDesignThreadCommand.cs:16-17`, `:234`, `:242-255`. `DurableRecord` says the block "adds no intent-cli option" (`:254`).
- **G822 refusal text.** `AutomationHostLoopNextActionCommand.cs:656-660`, reason `dispatch-identity-source-unavailable` (`:583`). The evidence ends with "until the Orca Run mailbox binding (#1774) provides dispatch identity." It is asserted at `G822ActionableIdentityRefusalTests.cs:74`.
- **`event-append-failed` hint.**
  - For `report` / `collect`, `ReportRoot` is always resolved (`NotifyCommand.cs:257-266`).
  - `senderLocalReport` is true only when the roots differ (`:2426-2427`).
  - An append failure gives `report-routing-root-write-required` when sender-local, and `event-append-failed` otherwise (`:2659-2662`). The outbox is marked undelivered (`:2663-2666`).
  - The `event-append-failed` summary ends with "Fix reader access and retry notify." (`:2672-2673`).
  - `BuildReconciliationCommand` returns null unless the roots differ (`:3303-3309`).
  - `notify reconcile` on an undelivered outbox returns `sender-local-report-not-delivered` with a collect recovery (`:1624-1646`).
  - `escalate` also emits `event-append-failed` (`:3153`).
- **Orca Run ids (measured 2026-09-15, read-only `orca orchestration run-list --json` on the operator Mac).**
  - All 38 ordinary Run ids match `run_` followed by 12 lowercase hex characters.
  - The only other id is the `run_legacy_local` adoption tombstone.

- G592 topology record/show/validate; G776 `wake_command`; G739 declared role fields.
- G691/G833 team mode, including `solo-conductor`.
- G811 notify reconcile semantics: pending records, the sender-local outbox, and collect recovery.
- G822 host-loop identity refusal; G795 role vocabulary.
- Installed Orca exposes `orchestration run-create`, `run-use`, `run-show`, `run-list`, `send`, and `check` (`reports/G813-packet-review-20260907.md`). This unit calls none of them.

## Target Repo / Path / Part

- Target paths: `src/IntentSystem.Cli/`, `tests/IntentSystem.Cli.Tests/`, `docs/en/`, `docs/ja/`

Target part: `recorded Orca Run binding on the team-shape seat (topology role for four/five-seat teams, local orca-runs record for solo-conductor) with lock-guarded record-orca-run write and team-mode set guard, read-only orca-runs discovery, show/validate/bootstrap handover exposure, closed-list receive-policy label; G822 and design-thread text corrections; equal-root event-append-failed reconciliation command`

## In Scope

- One guarded write command: `session-layer topology record-orca-run`.
- One read-only discovery command: `session-layer topology orca-runs`.
- A local solo binding record under `.intent-cli/orca-runs/`.
- A shared team-mode lock, with a guard and digest re-check in `team-mode set --write`.
- Binding exposure in `topology show`, `topology validate`, `team-mode validate`, and `guide bootstrap`.
- The notify `event-append-failed` reconciliation command.
- Text corrections for the G822 refusal and the design-thread Orca block.
- Tests, docs, and ledger entries.

## Out Of Scope

- Running `orca` in any form: create, adopt, show, list, send, check, or wait. Verifying that a Run exists or is live.
- Run creation and send/check adapters, which wait for measured Orca create idempotency.
- Executing a receive policy: pushing, waking, polling, or scheduling. Installing a scheduled wake.
- Dispatch identity, including dispatch generation and digest (G822 remainder, a later unit).
- More than one binding per team, bindings on seats other than the one the shape selects, and bindings on herdr-resident seats.
- Inferring shape, seat, seat kind, or Run from Orca presence, objective prose, residence, or co-location.
- Changing notify routing: a binding is never a delivery target.
- Adding locks to the existing unlocked topology writers.
- Changing `record`, `update-field`, `update-residence`, `move`, `retire-legacy`, `record-host-state`, or `record-profile`.
- Changing `escalate`'s `event-append-failed`.
- Editing published release notes.
- Migrating existing topology or team-mode files.
- Recording a binding on the live `intent-cli-dev` team.

**1. Recorded shape of a binding.** A team has at most one binding.

- **Four- or five-seat team.** The binding is an optional object on the exact role key in `.intent-cli/topology/<domain>/<team>.json` (a local-ignored file):

  ```json
  "roles": { "steward": { "resident": "external", "reader": "…", "frontend": "orca",
    "orca_run": { "run_id": "run_585adfc0e774", "receive_policy": "orca-push" } } }
  ```

  It lives on the role, never at team level, because a flat layout reads team-level objects as roles. `orca_run` is not added to `KnownRoleFields` or to either envelope-property list.
- **`solo-conductor` team.** The binding is its own record, `.intent-cli/orca-runs/<domain>/<team>.json`:

  ```json
  { "schema_version": "1", "domain": "d", "team": "t",
    "orca_run": { "role": "design", "run_id": "run_…", "receive_policy": "inbox-pull", "frontend": "claude-app" } }
  ```

  The first write creates `.intent-cli/orca-runs/.gitignore` containing `*`, the same pattern as `EnsureLocalIgnore`. `team-mode.json` never carries a binding, and `TeamModeEntry` is unchanged. Section 11 gives the reason: no older intent-cli writer can touch this file.
- **Sync across clones.** Both binding locations are local-ignored, so no binding syncs across clones. `team-mode.json` is git-tracked, so team mode does sync. A fresh clone therefore has the team mode but no binding. This matches Run ids, which belong to one Orca runtime.
- **Lenient reads.** A malformed `orca_run` never makes `TeamModeStore.TryRead`, topology `Resolve`, notify delivery, the store `Validate`, `topology show` validity, or `guide bootstrap` fail. It is reported only by the surfaces in sections 5 to 7.
- **Grammar.**
  - `run_id` must match `^run_[0-9a-f]{12}$`. `run_legacy_local` does not match.
  - `receive_policy` is `orca-push` or `inbox-pull`.
  - Solo `role` must normalize to `architect`.
  - Solo `frontend` is a string.
  - The solo record's `domain` and `team` must equal its path segments (`binding-identity-mismatch`).

**2. Team shape.** The shape is decided only from recorded fields, with an explicit team, in this order:

1. **Team-mode read.** `TeamModeStore.TryRead` then `Resolve(state, domain, team)`. A throw gives `team-shape-unreadable`, carrying the store message.
2. **Unrecorded mode.** `Source == Default` gives `team-shape-unrecorded`. Fix: `intent-cli team-mode set --domain <d> --team <t> --mode delivery|solo-conductor --write`.
3. **Authoring-only.** `mode == authoring-only` gives `team-shape-not-bindable`.
4. **Solo-conductor.** `mode == solo-conductor` gives shape `solo-conductor`, required seat `architect`, and location `orca-run-file`. The resolved entry must be team-scoped, or `team-shape-solo-entry-not-team-scoped`.
5. **Delivery.** `mode == delivery` (team-scoped or domain-wide):
   - Resolve the topology. An unresolved result gives `topology-unresolved`, carrying the store cause.
   - A role key that does not normalize gives `team-shape-unrecognized`.
   - The canonical set `{architect, orchestrator, builder, reviewer, steward}` gives `five-seat` with required seat `steward`.
   - The set `{architect, orchestrator, builder, reviewer}` gives `four-seat` with required seat `architect`.
   - Any other set gives `team-shape-unrecognized`, naming the missing canonical roles.
   - Location is `topology-role`.
   - Several keys may normalize to one canonical role. They are never merged; the operator names the exact key.

One shared pure resolver serves the write command, validate, discovery, show, bootstrap, and the `team-mode set` guard.

**3. Receive-policy label** (closed list, never executed). The seat kind comes from the recorded frontend only: the role's `frontend` for four/five-seat, or the binding's `frontend` for solo.

| Recorded seat | Accepted policy | Otherwise |
|---|---|---|
| `resident: external`, `frontend: orca` | `orca-push` | `receive-policy-seat-mismatch` |
| `resident: external`, `frontend: claude-app` or `codex-app` | `inbox-pull` | `receive-policy-seat-mismatch` |
| `resident: external`, any other or absent frontend | none | `receive-policy-seat-kind-unrecorded`; fix: `session-layer topology update-field --field frontend --current <value\|absent> --new orca\|claude-app\|codex-app …` (solo: `--frontend`) |
| `resident: herdr` | none | `receive-policy-herdr-seat` |

A herdr seat can never record a frontend (see Current Observed State), so every herdr seat is refused. The fix names `intent-cli session-layer topology update-residence --domain <d> --team <t> --role <role> --current-resident herdr --new-resident external --reader <routing-root-relative-path> --frontend orca --confirm-update-residence --dry-run --format json`. The operator records the seat as external with frontend `orca`. `frontend` stays a label, and delivery routing never reads the binding.

**4. Write: `record-orca-run`.**

```text
intent-cli session-layer topology record-orca-run --domain <d> --team <t> --role <role> --current <run-id|absent|malformed> --new <run-id|absent> [--receive-policy orca-push|inbox-pull] [--frontend <name>] --confirm-record-orca-run (--dry-run|--write) --format json
```

- **Gate exemption.** `record-orca-run` and `orca-runs` are exempt from the pre-dispatch team-mode gate (`SessionLayerTopologyCommand.cs:97-114`). `Execute` skips the gate when `args[0]` is one of them. Team mode is read only by the section 2 resolver, with an explicit team, so the order below is the real dispatch order. Every other subcommand keeps the gate unchanged.
- **Registration.** It goes in the topology switch, `Usage`, the help listing, and the `guide commands list` session-layer purpose.
- **Order.** Exit 1 on every refusal. A refusal changes no pre-existing file; the only paths it may create are listed under *Paths a refusal may create*.
  0. **Parse.** A missing required option or confirm, both or neither of `--dry-run` / `--write`, or a `--format` other than `json` prints the error and usage as text. This applies to authoring-only teams too.
  1. `orca-run-id-malformed` (`field`: `new` or `current`).
  2. `receive-policy-missing`, `receive-policy-not-accepted` (a policy with `--new absent`), or `receive-policy-invalid`.
  3. **Team-mode lock (write only).** Create `.intent-cli/locks/` and write `.intent-cli/locks/.gitignore` (`*`) if it is absent, then open `.intent-cli/locks/team-mode.lock`. A busy lock refuses `record-lock-busy` with `field: team-mode-lock`. A dry run skips this step.
  4. **Shape.** The section 2 causes, in order, from a read taken under the team-mode lock. These causes fix the binding location.
  5. **Topology lock (write only, location `topology-role` only).**
     - If `.intent-cli/topology/.gitignore` is not exactly `*`, call `EnsureLocalIgnore`. Then take `AcquireCasLock` on the resolved topology path.
     - A busy lock refuses `record-lock-busy` with `field: topology-lock`.
     - Once the lock is held, re-read the topology bytes and record their digest. Steps 6 to 12 evaluate that locked snapshot. If the locked snapshot no longer resolves to the step 4 shape and location, refuse `record-cas-lost`.
     - A solo location takes no topology lock. Its step 9 reads the topology file without a lock.
     - **Precedence.** A team-mode lock busy (step 3) comes before every shape cause. Every shape cause (step 4) comes before a topology lock busy (step 5).
  6. `role-not-recorded` (four/five-seat), or `binding-seat-not-allowed`: the canonical role of `--role` is not the required seat. The message names the shape and the required seat.
  7. `frontend-not-accepted` (`--frontend` on four/five-seat; fix: `update-field`), or `frontend-missing` (solo with an id and no `--frontend`).
  8. When `--new` is an id, the section 3 causes: `receive-policy-herdr-seat`, `receive-policy-seat-kind-unrecorded`, `receive-policy-seat-mismatch`.
  9. `binding-duplicate` (another role key in the team records `orca_run`), or `binding-location-conflict` (solo, while the team's topology file records `orca_run` on any role).
  10. `binding-identity-mismatch` (an existing solo file names another domain or team).
  11. `current-mismatch`. The recorded value is its `run_id`, `absent`, or `malformed`.
  12. `record-cas-lost` (write only). Immediately before the write, the target file's digest differs from the locked snapshot: the step 5 digest for topology, or the solo binding file digest (or its absence) read under the team-mode lock.
  13. `record-write-failed` (IO).
- **Paths a refusal may create.** Only under `--write`; a dry run creates nothing.
  - From step 3 onward (including a step 3 busy refusal): `.intent-cli/locks/`, `.intent-cli/locks/.gitignore`, `.intent-cli/locks/team-mode.lock`.
  - From step 5 onward (including a step 5 busy refusal): `.intent-cli/topology/.gitignore`, if it was not already `*`, and `.intent-cli/topology/<domain>/<team>.json.lock`.
  - Nothing else. `.intent-cli/orca-runs/`, its `.gitignore`, and its binding files are created only by a successful solo write.
  - Lock files are never deleted.
- **Ignore before lock.** Both lock steps write the ignore file before opening the lock, so a lock file never exists outside an ignored directory. This deliberately differs from the `move` / `update-residence` order (`SessionLayerTopologyCommand.cs:1952-1955`, `:2530-2532`: directory, lock, then ignore).
- **Writes.** Topology writes use `WriteAtomically` and change only the role's `orca_run` node. A solo write creates or replaces only the binding file, and `--new absent` deletes it. `team-mode.json` is never written.
- **Outcomes.** A request equal to the recorded binding returns `already_recorded: true` with exit 0 and does not rewrite the file. Under `--write` it may still have created the step 3 and step 5 paths. A dry run takes no lock and creates or writes nothing.
- **Result JSON** (snake_case, all keys): `operation` (`topology-record-orca-run`), `domain`, `team`, `role`, `canonical_role`, `team_shape` (`five-seat|four-seat|solo-conductor|null`), `team_mode`, `binding_location` (`topology-role|orca-run-file|null`), `record_path`, `current`, `new`, `receive_policy`, `frontend`, `mode` (`dry-run|write`), `applied`, `changed`, `already_recorded`, `conflict`, `cause`, `field`, `fix`, `summary`. Every success summary ends with `Recorded only; intent-cli did not run orca or verify the Run.`
- **Byte equality.** A topology write re-serializes with `FileJsonOptions`. The file bytes equal base plus `orca_run` only for a fixture that the CLI writer itself produced. For a hand-authored fixture, the guarantee is a `JsonNode.DeepEquals` of everything except `orca_run`.

**5. Validate: the binding-health table.** This table is the single source for every surface. Sections 6 and 7 and the acceptance criteria refer to it and never restate it.

- **Health.** For one existing binding, the applicable causes are evaluated in the row order below. `health` is the first applicable cause, or `recorded` when none applies. A surface that lists findings lists every applicable cause in this order.
- **Columns.** `TV` is a `topology validate` finding on the bound role (`role` = key, `field` = `orca_run`). `TMV` is a `team-mode validate` finding on the solo binding file, rendered as `<cause>: <message>`. `OR` is an `orca-runs` `health` value and `findings[]` entry. `n/a` means the cause cannot arise for that surface's binding kind. `unreachable` means the surface answers earlier, and the reason is given.

| # | Cause | Applies to | TV | TMV | OR |
|---|---|---|---|---|---|
| 1 | `orca-run-binding-malformed` (`orca_run` is absent or null, is not an object, or has a field of the wrong JSON type) | both | yes | yes | yes |
| 2 | `binding-identity-mismatch` | solo file | n/a | yes | yes |
| 2a | `orca-run-file-schema-unsupported` (the file parses, but `schema_version` is absent or is any value other than `"1"`; it is therefore never `orca-run-file-unparseable`) | solo file | n/a | yes | yes |
| 3 | `orca-run-id-malformed` | both | yes | yes | yes |
| 4 | `receive-policy-invalid` | both | yes | yes | yes |
| 5 | `team-shape-unreadable` | both | unreachable (the gate answers `team-mode-unreadable`) | unreachable (base unreadable-state output) | yes |
| 6 | `team-shape-unrecorded` | both | yes | yes | yes |
| 7 | `team-shape-not-bindable` | both | unreachable (the gate answers `not-applicable-team-mode`) | yes | yes |
| 8 | `team-shape-solo-entry-not-team-scoped` | solo file | n/a | yes | yes |
| 9 | `topology-unresolved` | topology role | never (see *Row 9* below) | n/a | yes |
| 10 | `team-shape-unrecognized` | both | yes | yes | yes |
| 11 | `binding-location-not-allowed` | both | yes (role binding on a `solo-conductor` team) | yes (solo file on a `delivery` team) | yes |
| 12 | `binding-seat-not-allowed` | both | yes | yes | yes |
| 13 | `frontend-missing` | solo file | n/a | yes | yes |
| 14 | `receive-policy-herdr-seat` | topology role | yes | n/a | yes |
| 15 | `receive-policy-seat-kind-unrecorded` | both | yes | yes | yes |
| 16 | `receive-policy-seat-mismatch` | both | yes | yes | yes |
| 17 | `binding-duplicate` | topology role | yes (every role binding after the first key in ordinal order) | n/a | yes |
| 18 | `binding-location-conflict` | solo file | n/a | yes | yes |

- **Not health.** These are `orca-runs` `unreadable_records` entries only. They never produce a `bindings[]` row, a health value, or a `topology validate` or `team-mode validate` Orca finding: `topology-file-unparseable`, `topology-identity-mismatch`, `topology-team-missing`, `orca-run-file-unparseable`, and `team-mode-unreadable` (listed once). Two of them widen to a read failure as well as a parse failure. Every read this unit guards catches all three of `IOException`, `UnauthorizedAccessException`, and `JsonException`, the set at `NotifyRoleTopology.cs:968`; the two-type filter at `TeamMode.cs:200` is not the pattern to copy, because it would miss a permission-denied file: `topology-file-unparseable` also carries a topology file that cannot be read, and `orca-run-file-unparseable` a sidecar that cannot be read. `topology-identity-mismatch` and `topology-team-missing` do not widen, since both can only arise from a file that already parsed. `orca-run-file-unparseable` is the one exception to the "no `team-mode validate` finding" rule: see the `team-mode validate` bullet below. An unreadable `team-mode.json` also gives every remaining row the health `team-shape-unreadable` (row 5).
- **`TeamModeStore.TryRead` widens to `UnauthorizedAccessException`** (`TeamMode.cs:200`, which today catches only `IOException` or `JsonException`). It is in scope here because this unit makes every `orca-runs` row and every shape resolution depend on that read: without the widening, a permission-denied `team-mode.json` escapes as an unhandled exception instead of becoming `team-mode-unreadable` and row 5. The widened catch throws the same `InvalidOperationException` with the same message shape, so `team-mode show` / `set` / `validate` and `guide bootstrap` report it through their existing paths and no other behaviour changes.
- **What counts as a binding.** A topology role is a binding only when it carries an `orca_run` property; a role without one is simply unbound and produces no row and no finding. A solo binding file is a binding record by its existence, so a file that parses but has no (or a null) `orca_run` is row 1, never a silent `recorded` row with `run_id: null`.
- **Row 9 is a health value, not an unreadable record.** `orca-runs` reads a topology file's `orca_run` properties directly, so a file that can be read and parsed, passes identity and team selection, and records a binding always produces a `bindings[]` row, even when `NotifyRoleTopologyStore.Resolve` fails. Its health is `topology-unresolved`, and the message carries the store cause.
- **`topology validate` never emits `topology-unresolved`**, whatever made `Resolve` fail. What it suppresses is exactly that cause and the informational `orca-run-binding-absent`. The shape-independent causes (rows 1, 3 and 4) still apply to a bound role, and the shape-dependent causes (rows 5 to 8 and 10 to 18) are simply undecidable without a resolved roster, so they are absent too. There are two cases:
  - **The store validation also fails.** Both functions check the same fact independently, so `Validate` is already invalid and the Orca step never runs. Each of these makes both fail: a role that is not an object (`Validate:823-827`, `Resolve:367-372`), `model` (`:832`, `Resolve:416-424`), `reasoning_effort` (`:841`, `Resolve:427-435`), an unsupported resident (`:859-865`, `Resolve:376-381`; `:375` is only the `ReadString`, and the role-object failure above ends at `:372`), an empty roster (`:904-908`, `Resolve:501-507`), a missing or ambiguous `workspace_id` (`:937-945`, `Resolve:509-518`), and an invalid `host_state` (`:915-927`, `Resolve:535-541`). `envelope_profile` is the only one `Validate` learns from `Resolve` itself, and only because that call adds a finding when the cause is `profile-invalid` (`:947-951`).
  - **Validate-only checks do not appear here.** An unsafe or unavailable `reader` is a `Validate` finding (`:881-890`) and never a `Resolve` failure, so it makes the store validation invalid on its own, without any `topology-unresolved` to report.
  - **A Resolve-only field fails.** `wake_command` grammar and its external-resident rule are read only by `Resolve` (`:438-455`, the herdr rule at `:449-455`); `Validate` never reads them, so the topology can be valid while `Resolve` fails. The Orca step then emits no `topology-unresolved` and no `orca-run-binding-absent`, and no shape-dependent cause, but it still emits rows 1, 3 and 4 for a bound role, with `valid` false and exit 1.
    - The output is byte-identical to base only when no bound role trips row 1, 3 or 4.
    - The `orca-runs` row still reports the binding, with the health its own first applicable cause: row 3 for a malformed id, and `topology-unresolved` only when no earlier row applies.
- **`topology validate`.** Gate unchanged. After the reader-path findings, and only when the store validation is valid, add one finding per applicable `TV` cause per bound role. Any such finding makes `valid` false with exit 1. With no applicable cause, the output is byte-identical to base except for the absent finding below.
- **No binding on any role.**
  - Shape `five-seat` or `four-seat`, and every other finding informational: one informational `orca-run-binding-absent`, naming the required seat and ending with its `record-orca-run` command. This is the only cause that is not in the table, and it belongs to `topology validate` alone.
  - Shape unrecorded, unrecognized, or otherwise undecided: nothing is added.
- **`team-mode validate --domain <d> --team <t>`.** When `.intent-cli/orca-runs/<d>/<t>.json` exists, evaluate the `TMV` column for that file, whatever the team mode.
  - Append one finding per applicable cause, and set `valid` false with exit 1 only when at least one applies.
  - **A file that cannot be read or parsed.** Append `orca-run-file-unparseable: <message>`, with the message naming the failure and set `valid` false with exit 1. It is the only *Not health* cause that `team-mode validate` reports, because `team-mode validate` is the surface a solo seat runs for its own team, and a corrupt sidecar would otherwise be visible only in `orca-runs`. It produces no health value and no `bindings[]` row, so the table is unchanged.
  - **A healthy binding adds no finding.** `valid` stays true, the exit code stays 0, and the output is byte-identical to base.
  - Without that file, the output is byte-identical to base.
  - An unreadable `team-mode.json` keeps the base failure output.
  - **`--team` omitted (intended).** The rule keys on `<d>/<t>.json`, so a domain-wide invocation evaluates no binding and stays byte-identical to base. This is deliberate: without `--team` the command resolves one mode through the domain entry or a unique team-scoped fallback, and fails closed as `team-mode-ambiguous` when several teams match (`TeamMode.cs:227-244`). It holds no team list, so enumerating every team's binding would widen its scope and could report findings for teams whose mode it did not resolve. The host-wide view is `session-layer topology orca-runs [--domain <d>]`, and the docs say so.

**6. Discovery (read-only): `orca-runs`.**

```text
intent-cli session-layer topology orca-runs [--domain <d>] [--format markdown|json]
```

- **Gate.** Exempt (section 4). `team-mode.json` is read once with `TeamModeStore.TryRead`, and each row resolves with its explicit team. A domain with several team-scoped entries and no domain-wide entry therefore lists normally.
- **Inputs.**
  - Files exactly at depth two under `.intent-cli/topology/<domain>/` and `.intent-cli/orca-runs/<domain>/`, whose name ends in `.json` and does not start with `.`. This skips `.gitignore`, `*.lock`, and the `.<file>.<guid>.tmp` files from `WriteAtomically`. Any other name is ignored.
  - For a topology file, the team is the file stem. It is read through the store identity and `TrySelectTeam` rules, so a `teams` object contributes only the stem's team. Other teams in that object are not listed; their canonical records live at their own paths.
  - The removed legacy `.intent-cli/role-pane-mapping.json` is not read.
  - A topology file contributes a row per role that carries `orca_run`. Every solo binding file that can be read and parsed contributes exactly one row, even when its identity does not match (row 2), its `orca_run` is absent or null (row 1), or its `schema_version` is unsupported (row 2a).
- **Output.** It writes nothing, takes no lock, and runs no process.
- **Exit.** Exit 0 when the listing completes, including unhealthy rows and unreadable records. Exit 1 only when the routing root cannot be enumerated (`orca-runs-unreadable`).
- **JSON.**
  - `routing_root`, `domain_filter`.
  - `bindings[]`, sorted by domain, team, role (ordinal). Each row has `domain`, `team`, `role`, `canonical_role`, `team_shape`, `binding_location`, `record_path`, `run_id`, `receive_policy`, `health`, `findings[]`.
  - `unreadable_records[]` with `record_path`, `cause`, `message`, carrying exactly the section 5 *Not health* causes and nothing else.
  - `summary`, always including `Recorded bindings only; no Run was contacted.`
- **Health.** `health` and `findings[]` are the section 5 table's `OR` column. Rows have no gate, so every cause in the table can appear, including `team-shape-not-bindable` for a binding stranded on an authoring-only team and `topology-unresolved` for row 9.

**7. Show and handover.**
- **`topology show`** (gate unchanged). `SessionLayerTopologyShownRole` gains `orca_run`, omitted when absent: `{ "run_id": string|null, "receive_policy": string|null, "health": "recorded"|<cause> }`. Markdown appends ` orca_run=<run_id|malformed>/<receive_policy|malformed> (<health>);` to that role's line. `valid` and the exit code are unaffected.
- **`guide bootstrap --domain <d> --team <t>`.** In the delivery and solo flows only:
  - **No binding** (no `orca_run` on any role, and no solo binding file). `orca_run_binding` is omitted and no markdown line is inserted. The JSON and markdown are byte-identical to base. This is the case for every team on a host that has never recorded a binding, in both shapes.
  - **A binding exists.** The result gains `orca_run_binding` with `role`, `binding_location`, `health` (the section 5 `OR` column), `run_id` and `receive_policy` (only when `health == recorded`), and `receive_instruction`.
  - `receive_instruction` is exactly `Receive by Orca push: this seat runs in an Orca terminal; intent-cli sends and wakes nothing.` for `orca-push`, or `Receive by scheduled pull: this app seat reads its canonical inbox and status once per scheduled wake; intent-cli schedules nothing.` for `inbox-pull`. It is null otherwise.
  - When a binding exists, markdown inserts exactly one line immediately before `FinalHandoffStatement`:
    - `health == recorded`: `Orca Run mailbox (recorded, not verified against Orca): <run_id> on seat '<role>' (<receive_policy>). <receive_instruction>`
    - `health == topology-unresolved`: `Recorded Orca Run binding for seat '<role>' is not usable (topology-unresolved); the team topology does not resolve, so the shape is undecided. Run intent-cli session-layer topology orca-runs --domain <d> --format json for the cause, then repair the topology record.` `topology validate` is deliberately not named here, because it never emits this cause (section 5).
    - any other health: `Recorded Orca Run binding for seat '<role>' is not usable (<health>); run intent-cli session-layer topology validate --domain <d> --team <t> --format json.`, or the same sentence with `intent-cli team-mode validate --domain <d> --team <t> --format json` for a solo binding file. Both surfaces do report those causes.
    - The "not usable" line is never emitted when there is no binding.
  - `FinalHandoffStatement` strings and every other field are unchanged.

**8. Team-mode lock and `team-mode set` guard.**
- **Lock.** `.intent-cli/locks/team-mode.lock`, opened with `FileShare.None` like `AcquireCasLock`. The first acquisition creates `.intent-cli/locks/.gitignore` containing `*`. The lock lives in that CLI-ignored directory, not beside the tracked `team-mode.json`, so it never adds untracked noise to a host checkout (compare `QueueReprioritizeCommand.cs:508`, whose lock is committed on this host). The docs state that the directory is local-only.
- **`team-mode set --write`.** Order:
  1. Create the locks directory and its ignore file, as in section 4 step 3. Then acquire the lock, or print `team-mode-write-refused: team-mode-lock-busy: …` with exit 1.
  2. Read `team-mode.json` and its digest.
  3. Evaluate the guard.
  4. Build the new state.
  5. Re-read the digest. A changed digest prints `team-mode-write-refused: team-mode-cas-lost: …` with exit 1.
  6. Write, then release.
- **Dry run.** `team-mode set` without `--write` takes no lock, creates nothing, and writes nothing, but it evaluates the same guard on the current state.
  - A guard refusal prints the same `team-mode-write-refused: orca-run-binding-present: …` line with exit 1.
  - A dry run that passes the guard exits 0, and its summary carries the same pinned sentence below whenever an affected topology record could not be read or parsed.
  - It is therefore byte-identical to base whenever every affected topology record could be read and parsed, and differs from base only by that sentence otherwise. It never writes, in either case.
- **Guard.** The guard compares the effective mode, not the entry-level flag.
  - For every team *T* in the target domain that has a binding (a solo binding file, or `orca_run` on any role of a topology file that can be read and parsed, found by listing `.intent-cli/topology/<domain>/` and `.intent-cli/orca-runs/<domain>/`), compute `before = Resolve(currentState, domain, T).Mode` and `after = Resolve(newState, domain, T).Mode`, where `newState` is the state the set would write.
  - If `before != after` for any such team, the set refuses: `team-mode-write-refused: orca-run-binding-present: team '<T>' in domain '<domain>' records an Orca Run binding at <record_path>; the set would change its effective mode from '<before>' to '<after>'; clear it first with <record-orca-run … --current <id> --new absent --confirm-record-orca-run --write --format json>`, with exit 1.
  - **Why the effective mode.** It is exactly the input the section 2 shape resolver reads, so a binding's shape can change only when its team's effective mode changes (the roster is untouched by a set). The entry-level `changed` flag (`TeamModeCommand.cs:103-109`: `changed = !alreadyRecorded` for the exact `(domain, team)` entry) is true when a team-scoped entry is added that repeats the domain-wide mode. Using it would refuse that harmless set.
  - The entry-level flag also says nothing about which teams a domain-wide entry affects. The effective mode covers domain-wide sets without a separate affected-team rule.
- **A topology file the guard cannot read or parse.** Both failures are one class, and both are treated the same way: the file does not block the set, and the guard cannot see whether it holds a binding. This is accepted rather than fixed here.
  - A set never writes a topology or binding file, so nothing is lost. The worst outcome is a stranded binding.
  - **While the file stays unreadable**, the only signals are its `orca-runs` `unreadable_records` entry and the summary sentence below. There is no `bindings[]` row, so no health cause appears.
  - **Once the file can be read again**, the stranding shows up as an `orca-runs` health (`team-shape-not-bindable`, `binding-location-not-allowed`, or `binding-seat-not-allowed`) and as the matching `topology validate` finding.
  - Failing the set closed instead would let one corrupt or unreadable topology block team-mode changes for every other team in the domain, and would change base behaviour for teams that have no binding at all.
  - **Summary sentence (pinned, both modes).** Whenever the guard could not read or parse at least one affected topology file, the `team-mode set` summary, in `--write` and in dry run alike, ends with exactly: `Orca Run binding guard: <n> affected topology record(s) could not be read or parsed and were not checked: <paths>.` `<n>` is the count and `<paths>` is the routing-root-relative paths, ordinal-sorted and joined by a comma and one space (`, `). The sentence belongs to the exit-0 summary only: a guard refusal, `team-mode-lock-busy`, or `team-mode-cas-lost` prints its refusal line and exits 1 with no summary object (`TeamModeCommand.cs:174-178`), so it carries no pinned sentence. With no such file, neither mode's summary changes, and both stay byte-identical to base.
- **Unchanged cases.** A set that changes no bound team's effective mode (including same-mode sets, sets on teams without bindings, and a new team-scoped entry that repeats the domain-wide mode) has base output and exit code, in both modes. Two exceptions, and only these: `--write` may create the `.intent-cli/locks/` paths, and either mode appends the pinned summary sentence when an affected topology record could not be read or parsed.
- **Shared lock.** `record-orca-run --write` holds the same lock from section 4 step 3 through its write. A `record-orca-run` and a `team-mode set` can therefore never interleave between check and write.

**9. Text corrections.**
- **G822.** Replace `AutomationHostLoopNextActionCommand.cs:659-660` with `"and no CLI argument or checkout change can supply it. Team-scoped host-loop cannot reach qualified until a later unit delivers a dispatch generation and digest source; the Orca Run mailbox binding does not provide dispatch identity."`. The reason code, classification, and other evidence stay unchanged.
- **`guide design-thread`.**
  - `DurableRecord` becomes `Canonical intent-cli notify remains durable. intent-cli records an adopted Run id with session-layer topology record-orca-run and shows it in topology show and guide bootstrap, but never runs, launches, verifies, or manages Orca.`
  - `SetupOrder` gains a fourth step: `Record the adopted <run-id> on the seat the recorded team shape selects (steward for five-seat, design for four-seat and solo-conductor; the seat must be external with frontend orca, claude-app, or codex-app) with intent-cli session-layer topology record-orca-run … --dry-run first.`
  - No other field changes.
- **Docs.** The `docs/en` and `docs/ja` `12-agent-message-orchestration.md` Orca paragraphs match.

**10. Notify reconciliation command.** For `report` and `collect` failure results whose cause is `event-append-failed` or `report-routing-root-write-required`, `reconciliation_command` is always:

`intent-cli notify reconcile --domain <d> --team <t> --task-id <id> --routing-root '<routing>' --report-root '<resolved report root>' --write --format json`

This holds even when the roots are equal. The `event-append-failed` summary for these operations replaces `Fix reader access and retry notify.` with `The report is retained at '<outbox path>' and marked undelivered; orchestration reconciles with '<reconciliation_command>', which names the notify collect recovery while the report is undelivered.` Success results, other causes, and `escalate` are byte-identical to base.

**11. Migration, back-compat, and older writers.**
- **No migration.** Files without `orca_run`, and hosts without `.intent-cli/orca-runs/`, behave exactly as on base, apart from three exceptions:
  - the informational `orca-run-binding-absent`, for recorded four/five-seat shapes;
  - the creation of the ignored `.intent-cli/locks/` directory by `team-mode set --write`;
  - the pinned `team-mode set` summary sentence, in either mode, when an affected topology record could not be read or parsed.

  A team with no binding whose affected topology records all read and parse is byte-identical to base on every surface, in both modes, apart from the informational `orca-run-binding-absent` that `topology validate` adds for a recorded four- or five-seat shape.
- **Older writers and solo bindings.** An older intent-cli's `team-mode set` re-serializes the whole `team-mode.json` through the typed `TeamModeEntry` (`TeamModeCommand.cs:122-146`; serialization `TeamMode.cs:269-273`, `TeamModeStore.Write` `TeamMode.cs:275-280`). Had solo bindings lived on the entry, any mode change by an older CLI for any domain or team would have dropped every solo `orca_run` on the host silently. Because `team-mode.json` is git-tracked, that loss would have been committed and propagated.
  - This unit therefore keeps solo bindings out of `team-mode.json`. No shipped writer knows `.intent-cli/orca-runs/`, so an older CLI cannot remove a binding.
  - The remaining older-CLI effect is stranding: an older `team-mode set` can change a bound team's mode without the guard. The binding file stays, and `team-mode validate`, `topology validate`, and `orca-runs` report `binding-location-not-allowed` or `binding-seat-not-allowed`.
- **Topology writers.** Older and current `record`, `update-field`, `update-residence`, `move`, `update-kind`, `record-host-state`, and `record-profile` edit topology JSON nodes in place, so they preserve an `orca_run` role property.
- **Pre-existing risk, unchanged.** `record`, `record-host-state`, `record-profile`, `update-kind`, and `update-field` take no topology lock. A concurrent run of one of them and `record-orca-run` can lose either write. `record-orca-run`'s digest re-check detects a change made before its own write, but not one made after it.
- **Later seat changes.** After `update-residence` or a roster change, `validate` reports the resulting `receive-policy-*` or `binding-seat-not-allowed`, and no writer refuses.

## Acceptance Criteria

- [ ] **AC1 — Four/five-seat write.**
  - Setup: CLI-written fixtures for a recorded `delivery` five-seat team (`steward` external, `frontend: orca`) and a four-seat team (`design` external, `frontend: claude-app`).
  - Dry run plans, takes no lock, and writes nothing.
  - `--write` makes the file bytes equal base plus only `orca_run`. A hand-authored variant is compared by DeepEquals, excluding `orca_run`.
  - A rerun is `already_recorded`.
  - `--new absent` clears.
  - A policy-only change applies.
  - Test: `OrcaRunBindingWriteTests`.
- [ ] **AC2 — Solo write.**
  - Setup: a team-scoped `solo-conductor` entry and no topology file.
  - `--role design --frontend claude-app --receive-policy inbox-pull --write` creates only `.intent-cli/orca-runs/<d>/<t>.json` and its `.gitignore`. `team-mode.json` is byte-unchanged.
  - `--new absent` deletes the file.
  - A malformed binding file keeps `team-mode show`, `guide bootstrap`, and topology commands working, and surfaces in `team-mode validate` and `orca-runs`.
  - **Healthy path.** With the written binding in place, `team-mode validate --domain <d> --team <t>` keeps `valid: true`, exit 0, and output byte-identical to base, in both markdown and JSON. `orca-runs` shows health `recorded`.
  - **Absent or null `orca_run`.** A solo file that can be read and parsed but records `orca_run` absent, and a second one that records it as `null`: both give health `orca-run-binding-malformed` with `run_id: null`, never `recorded`; `team-mode validate` fails with that cause; and `guide bootstrap` renders the "not usable" line, never the recorded line.
  - **Unreadable or unparseable sidecar.** One sidecar with broken JSON and one whose read throws, through the test-only read seam (see *Unreadable fixtures* below): both make `team-mode validate --domain <d> --team <t>` report `orca-run-file-unparseable` with exit 1, and `orca-runs` lists each under `unreadable_records` with no `bindings[]` row. The same pair for a topology file gives `topology-file-unparseable`.
  - **Unsupported `schema_version`.** Two solo files that each carry a well-formed `orca_run`, one with no `schema_version` and one with `"2"`: both give `orca-run-file-schema-unsupported` on `orca-runs` and `team-mode validate`, and neither appears in `unreadable_records`. The fixtures carry a well-formed `orca_run` on purpose: row 1 precedes row 2a, so a file that is both malformed and of an unsupported version reports `orca-run-binding-malformed` first. That order is deliberate. A payload that cannot be read as a binding is the more specific fault, and the row order is not changed for this fixture.
  - **`--team` omitted.** `team-mode validate --domain <d>` with that binding present is byte-identical to base and exits 0.
  - **Unreadable `team-mode.json`.** With its read made to throw through the same seam, `orca-runs` exits 0 with `team-shape-unreadable` health and one `team-mode-unreadable` record, and `team-mode validate` reports its base unreadable-state output instead of an unhandled exception.
  - Test: `OrcaRunBindingSoloTests`.
- [ ] **Unreadable fixtures (applies to AC2, AC5, AC7, and AC8).** Every "cannot be read" fixture uses a test-only file-read seam that throws `UnauthorizedAccessException` or `IOException` for one named path. It never uses a file mode, which root ignores, and never a directory at the file's path: `File.Exists` is false for a directory and the `.json` enumeration skips it, so such a fixture would produce no record at all instead of an unreadable one. Existence checks and enumeration keep their current file semantics.
- [ ] **AC3 — Refusal order matches real dispatch.**
  - Each section 4 cause has its own test, in order: parse (including an authoring-only team with a missing confirm, which gets usage and not `not-applicable-team-mode`), `orca-run-id-malformed` (both fields, `run_legacy_local`, and an authoring-only team), the receive-policy causes, `record-lock-busy` (team-mode lock held, and topology lock held), `team-shape-unreadable`, `team-shape-unrecorded`, `team-shape-not-bindable`, `team-shape-solo-entry-not-team-scoped`, `topology-unresolved`, `team-shape-unrecognized`, `role-not-recorded`, `binding-seat-not-allowed` (design on five-seat, orchestration on four-seat, solo `--role review`), `frontend-not-accepted`, `frontend-missing`, `receive-policy-herdr-seat` (with the update-residence fix literal), `receive-policy-seat-kind-unrecorded`, `receive-policy-seat-mismatch`, `binding-duplicate`, `binding-location-conflict`, `binding-identity-mismatch`, `current-mismatch` (including `malformed`), `record-cas-lost`, and `record-write-failed`.
  - Each test asserts exit 1, `cause`, `fix` where defined, and unchanged bytes in every pre-existing file. It also asserts that the set of new paths is exactly the section 4 list for that step: none for a dry run or for steps 0 to 2; the locks paths from step 3; the topology ignore and lock paths from step 5.
  - Precedence tests:
    - team-mode lock held plus an unrecorded mode gives `record-lock-busy` (`team-mode-lock`);
    - topology lock held plus an unrecorded mode gives `team-shape-unrecorded`;
    - topology lock held plus a valid five-seat shape gives `record-lock-busy` (`topology-lock`);
    - a solo binding with the topology lock held succeeds.
  - Test: `OrcaRunBindingRefusalTests`.
- [ ] **AC4 — Alias roster.**
  - Five-seat: `design` external `claude-app`, `architect` external `claude-app`, and `steward` external `orca`, plus the other seats. Binding `steward` succeeds, and `architect` refuses `binding-seat-not-allowed`.
  - Four-seat (the same roster without `steward`): `design` binds `inbox-pull`, and a second binding on `architect` (external `claude-app`) reaches and refuses `binding-duplicate`.
  - No key is merged or renamed.
  - Test: `OrcaRunBindingAliasTests`.
- [ ] **AC5 — Discovery.**
  - A host with a healthy five-seat binding, a solo binding, a malformed id, a binding whose topology does not resolve (row 9), a topology that cannot be parsed, one whose read throws, a `teams`-object file, `.lock`, `.tmp`, and `.gitignore` siblings, a legacy `role-pane-mapping.json`, and a team with no binding: `orca-runs --format json` lists exactly the section 6 rows and unreadable records, with exit 0.
  - Mixed domain: team-scoped five-seat `delivery` plus team-scoped `solo-conductor`, with no domain-wide entry. `orca-runs --domain d` exits 0 and lists both. `record-orca-run` works for each team.
  - An unreadable `team-mode.json` gives exit 0 and `team-shape-unreadable` health.
  - Nothing is written, and a fake `orca` first on PATH records zero calls.
  - Test: `OrcaRunDiscoveryTests`.
- [ ] **AC6 — Show and handover.**
  - `topology show` JSON and markdown add `orca_run` only on the bound role.
  - `guide bootstrap` carries the exact literals for `orca-push` (five-seat steward), `inbox-pull` (solo), a malformed binding (the "not usable" line, no `run_id`, and the shape's validate command), and a `topology-unresolved` binding (the `orca-runs` line, which names no validate command).
  - **No binding, byte-identical.** For a recorded four-seat team, a recorded five-seat team, and a `solo-conductor` team with no binding, `guide bootstrap --format markdown` and `--format json` equal base byte for byte, with no `orca_run_binding` key and no inserted line. The same holds for `topology show`.
  - Test: `OrcaRunShowHandoverTests`.
- [ ] **AC7 — Validate health.**
  - For every table row, one test asserts the surface mapping: a `TV` finding, a `TMV` finding, an `OR` health value, or the stated `unreachable` / `n/a` outcome. `topology validate` emits each `TV` cause with exit 1.
  - **Row 9, both cases.** A topology file that can be read and parsed, selects its team, and records `roles.steward.orca_run`, but whose `Resolve` fails: once on a missing `workspace_id` (store validation also invalid), and once on a `wake_command` on a herdr role that is **not** the bound seat (a Resolve-only field, store validation valid). The two roles cannot be the same one: the Resolve-only failure needs a herdr role carrying `wake_command` (`Resolve:449-455`), while a bound seat must be external, because section 3 and section 4 step 8 make `record-orca-run` refuse a herdr seat with `receive-policy-herdr-seat`, so no binding can be recorded on that role in the first place. The third fixture below mutates this one, so it inherits the same separation. In both, `orca-runs` gives one `bindings[]` row with health `topology-unresolved`, no `unreadable_records` entry, and exit 0. In both, `topology validate` emits no `topology-unresolved` finding and no `orca-run-binding-absent`: the first keeps its base store findings, and the second is byte-identical to base. The second fixture's binding is well formed on purpose, since the shape-independent causes (rows 1, 3 and 4) still apply when the store validation is valid; a third fixture that adds a malformed `run_id` to it asserts that `topology validate` still emits the row 3 finding with exit 1, and that its `orca-runs` health is `orca-run-id-malformed`, not `topology-unresolved`, because row 3 precedes row 9.
  - Each *Not health* cause produces an `unreadable_records` entry and no row.
  - A recorded four-seat team without a binding gets exactly one informational `orca-run-binding-absent` and exit 0.
  - An unrecorded mode is byte-identical to base.
  - A structural error suppresses the absent finding.
  - `team-mode validate` emits the `TMV` causes, including `team-shape-solo-entry-not-team-scoped` (a solo binding file whose team resolves only through a domain-wide solo entry), and `binding-location-not-allowed` after a stranding to delivery.
  - **Authoring-only stranding** (simulated older writer):
    - a topology role binding whose team becomes `authoring-only`: `orca-runs` row health `team-shape-not-bindable`, never `recorded`; `topology validate` keeps its base `not-applicable-team-mode` output; `guide bootstrap` (authoring-only flow) renders no binding;
    - a solo binding file on the same kind of team: `team-mode validate` reports `team-shape-not-bindable` with exit 1.
  - Health order: a binding with both a malformed id and a disallowed seat reports `orca-run-id-malformed` as `health`, with both in `findings[]`.
  - Test: `OrcaRunValidateTests`.
- [ ] **AC8 — Team-mode lock, guard, and interleaving.**
  - A solo binding plus `set --mode delivery --write` refuses `orca-run-binding-present`.
  - A topology role binding plus `set --mode solo-conductor` refuses.
  - After clearing, the set applies.
  - A held lock refuses `team-mode-lock-busy`.
  - The guard on a dry run: `set --mode delivery` without `--write` on a bound solo team prints the `orca-run-binding-present` line with exit 1 and creates no path, including `.intent-cli/locks/`.
  - Repeating the domain-wide mode: with a domain-wide `delivery` entry and a five-seat team `T` holding a topology binding, `team-mode set --domain d --team T --mode delivery --write` applies. The entry-level `changed` is true, but the effective mode is unchanged. The binding's health stays `recorded`.
  - A domain-wide set to `solo-conductor` that would change `T`'s effective mode refuses. The same set, when `T` has its own team-scoped `delivery` entry, applies.
  - Interleaving uses a test hook between guard evaluation and write:
    - a hook that rewrites `team-mode.json` makes the set refuse `team-mode-cas-lost`;
    - a hook that runs `record-orca-run --write` gets `record-lock-busy`, and the set then applies;
    - the reverse, a hook inside `record-orca-run` that runs `team-mode set --write`, gets `team-mode-lock-busy`.
  - An affected topology that cannot be parsed, and one whose read throws, each leave the set unblocked. Both the `--write` and the dry-run summary carry the pinned sentence naming both files.
  - Sets on teams without bindings, with every affected topology record readable, are byte-identical to base in both `--write` and dry run, and the lock directory's `.gitignore` is `*`.
  - Dry run, both halves: with an unreadable affected topology record, the dry-run summary ends with the pinned sentence and nothing is written (`team-mode.json` bytes, the locks directory, and every topology and binding file are unchanged); with no such record, the dry-run output is byte-identical to base.
  - Stranding by a simulated older writer (rewriting `team-mode.json` directly) is reported by `team-mode validate` and `orca-runs`, and the binding file is intact.
  - Test: `TeamModeOrcaRunGuardTests`.
- [ ] **AC9 — Notify reconciliation.**
  - An equal-root `report` with an unwritable reader:
    - cause `event-append-failed`;
    - the section 10 command with equal roots;
    - the summary names the undelivered outbox.
  - Following the command:
    1. `reconcile` returns `sender-local-report-not-delivered` with collect recovery.
    2. `collect --write` from a writable context appends exactly one event.
    3. A second `reconcile` converges without a second event.
  - A sender-local failure keeps its base output except the always-present command.
  - Success reports and `escalate` are byte-identical to base.
  - Test: `NotifyEventAppendReconcileHintTests`.
- [ ] **AC10 — Text and guide parity.**
  - The G822 evidence carries the section 9 literal and no `#1774`; `G822ActionableIdentityRefusalTests.cs:74` is updated with its reason.
  - The design-thread `durable_record` and the fourth `setup_order` step match section 9. Oracles that exclude `orca_operating_block` keep their hashes, and covering oracles are updated with the diff pasted.
  - `guide commands list` names both subcommands.
  - The en/ja ledger covers the binding records, both commands and every cause, the lock, the guard, the older-writer note, and the notify change. The G612 and G628 ledger tests pass.
- [ ] **AC11 — Unchanged surfaces and test safety.**
  - Existing topology subcommands keep the gate and their base output for fixtures without `orca_run`, and preserve `orca_run` when present.
  - No path starts `orca`, and a logging fake `orca` shows zero calls.
  - GitHub seams (for example `CandidateListerFactory`) default to throwing fakes, and no test reaches real GitHub.
- [ ] **AC12 — Validation.**
  - Focused tests and the full Release suite pass, with counts pasted.
  - Build-guarded mutations each fail a test: dropping the run-id grammar, dropping the required-seat check, dropping the herdr refusal, dropping the `team-mode set` lock or digest re-check, and making `orca-run-binding-absent` non-informational.
  - `git diff --check` is clean.
  - Exact-head CI is green.

## Verification

- Paste dry-run and write JSON for AC1 (five-seat steward) and AC2 (solo).
- Paste one refusal per cause.
- Paste `orca-runs` JSON for the AC5 host and the mixed domain.
- Paste `topology validate` for a malformed binding and for an absent binding.
- Paste the bootstrap handover markdown for `orca-push` and `inbox-pull`.
- Paste the AC8 interleaving outputs and the notify AC9 chain.
- Paste base-versus-head diffs for the unchanged surfaces.
- Paste focused and full-suite counts and the mutation results.
- Paste the fake `orca` call log (empty).

## Related Links

- #1771 (source), #1774 (retired G813, decomposed), PR #1781 (G813 team identity).
- G592 topology, G776 `wake_command`, G691/G833 team mode, G795 roles, G811 notify reconcile, G822 host-loop identity refusal.
- Host rulings: `reports/G813-orca-mailbox-design-20260907.md` (operator decisions 2026-09-15), `reports/PREVIEW3-SCOPE-RULING-20260914.md` (Orca amendment), `reports/SOLO-CONDUCTOR-MODEL-DESIGN-DRAFT-20260914.md`.

## Knowledge Maintenance

- ADR candidate: none
- Diagram candidate: none
- Docs update: en/ja `1.0-compatibility-ledger.md`, en/ja `12-agent-message-orchestration.md`
- Closeout writeback expected: yes. A cross-team mailbox is recorded as an exact adopted Run id on the seat that the recorded team shape selects, in a location that no older writer can erase. intent-cli records, shows, and validates it and never runs the transport. A receive policy is a closed-list label checked against the recorded frontend.

## Guide Reachability (G645)

- `guide bootstrap` → design/steward → recorded Orca Run handover line and `orca_run_binding`
- `guide design-thread` → architect → `session-layer topology record-orca-run`
- `guide commands list` → orchestration → `session-layer topology record-orca-run` and `session-layer topology orca-runs`
- `session-layer topology validate` → architect → `orca-run-binding-absent` fix command

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1820.

intent-cli records one already-adopted Orca Run id on the seat the recorded team shape selects — steward on a five-seat team, design on a four-seat or `solo-conductor` team — then shows it, validates it, lists it read-only across the host, and puts it in the `guide bootstrap` handover. A receive-policy label is recorded and checked against the seat's recorded frontend.

**intent-cli never runs `orca`.** Run creation, send/check adapters and receive-policy execution stay out of scope until Orca's create idempotency has been measured. Dispatch identity is a separate later unit; this one only corrects the G822 refusal text that promised it.

New surfaces:
- `session-layer topology record-orca-run` — the guarded write, with a shared team-mode lock and a digest re-check.
- `session-layer topology orca-runs` — read-only discovery across the host.
- `orca_run` on `topology show`, `topology validate`, `team-mode validate` and `guide bootstrap`.
- A `team-mode set` guard that refuses a mode change which would strand a binding, and names any topology record it could not read.

Solo-conductor bindings live in `.intent-cli/orca-runs/<domain>/<team>.json` rather than on the team-mode entry, so an older CLI's `team-mode set` cannot silently drop them.

Also fixes the report-routing reconciliation message from #1771: an `event-append-failed` failure now names its reconcile command even when the report root equals the routing root.

A host with no binding is byte-identical to base on every surface, apart from one informational `orca-run-binding-absent` finding on `topology validate` for a recorded four/five-seat team.

Design review: Claude subagent and Cursor Grok 4.6 xHigh, 10 rounds. Implementation review: the same two runtimes, 3 rounds, plus three conductor verification passes that re-ran the suite, the mutations and base-vs-head independently.


tomohisa (commented):
Implementation review — Cursor Grok 4.6 xHigh (`cursor-grok-4.6-xhigh`, read-only `--mode ask`), 3 rounds against head `eaf3f3c8`'s predecessors.

Round 1 (`33d68c40`): request-changes, 7 blocking — guard evaluated before the lock; `already_recorded` ignored `receive_policy`; a solo write rewrote an unchanged file; a malformed current value was unreachable; an empty or wrongly-typed `orca_run` read as healthy; the guard skipped bindings it should refuse on; a topology write dropped the trailing newline.

Round 2 (`8c302bf2`): approve. Verified in code that the lock order is now lock → read+digest → guard → build → re-check → write with the lock released on every refusal path, that `record-orca-run` holds the same lock so the old window is closed, and that the fake-`orca` harness now fails when the fake is absent.

Round 3 (`ae5714e9`): approve. Confirmed the `orca_run: null` shape is malformed on every surface and the guard's own clear command applies, that the shared `EnsureLocalIgnore` revert restores base repair behaviour, and that `DefaultIgnoreCondition.Never` and `[JsonIgnore]` are scoped to the two result types.

tomohisa (commented):
Implementation review — independent Claude subagent (read-only), 3 rounds.

Round 1 (`33d68c40`): request-changes, 5 blocking — the pre-lock guard race; `notify delegate` lost its base summary because the replacement was put on the shared delivery-failure path; a receive-policy-only change never applied; a malformed solo binding could not be cleared; and the fake-`orca` assertion was vacuous because it used this session's absolute paths.

Round 2 (`8c302bf2`): request-changes, 2 blocking — a role recorded as `orca_run: null` became permanently un-clearable once the guard counted it as a binding, and the write-ignore-only-when-absent change had been applied to the shared topology writer, silently stopping stale-`.gitignore` repair for `topology record`, `move`, `update-field` and `update-residence`.

Round 3 (`ae5714e9`): approve. Walked the null-shape clear path end to end, confirmed the shared-writer revert is character-identical to base, and checked the blast radius of the two serialization changes. Non-blocking notes on that round were applied in `eaf3f3c8`.

tomohisa (commented):
Conductor verification — three independent passes that re-ran everything rather than reading the builder's evidence.

Pass 1 (`33d68c40`) found the builder's evidence unreliable: mutation 3 was recorded as a clean build when its patch does not compile (CS0162), mutation 5's anchor matched two sites and hit the right one by ordering luck, the saved suite log recorded a failure the report called green, the focused count was unreproducible, and base-vs-head had only covered a four-seat team. It also confirmed the guard/lock ordering violation independently, and found that rewording the pinned `team-mode set` summary sentence broke no test.

Pass 2 (`8c302bf2`) reproduced 6823 passed / 1 skipped across 11 assemblies and 95 for the `_G837` filter exactly, drove all ten blocking fixes through the CLI, and killed 13 mutations including two of its own against the pinned sentence's separator and ordering.

Pass 3 (`aae8804a`) reproduced 6832 / 1 / 11 and 104 focused, killed all 14 mutations, and compared base against head on five-seat, four-seat and solo-conductor shapes — 27 comparisons, all identical except the one difference the contract permits. Its own mutation of the solo malformed-binding message survived the whole suite; that literal was corrected and pinned in `eaf3f3c8`.

Every pass verified the base binary's identity first (no `OrcaRunBinding` symbols) and probed both fakes for liveness. `orca` was never invoked.

